PR #26, opened 2022-09-28
📄 Update the scan github description

tas50:
We can scan users too

### DIFF
--- a/apps/cnspec/cmd/scan.go
+++ b/apps/cnspec/cmd/scan.go
@@ -215,7 +215,7 @@ configure your Azure credentials and have SSH access to your instances.`,
 				Short: "Scan a VMware vSphere VM",
 			},
 			"github": {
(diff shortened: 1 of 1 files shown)


apps/cnspec/cmd/scan.go
```diff
@@ -209,7 +209,7 @@ configure your Azure credentials and have SSH access to your instances.`,
 				Short: "Scan a VMware vSphere VM",
 			},
 			"github": {
-				Short: "Scan a GitHub organization or repository",
+				Short: "Scan a GitHub organization, repository, or user",
```
scottford-io: @tas50 Great catch. Only thing is this is still a bit misleading. We don't "Scan a user". What we do is scan organizations, a single repository, or a repository under a user account. how can we word that differently?
```suggestion
				Short: "Scan a GitHub organization, single org repository, or repository in a user account",
```

🤔

PR #46, opened 2022-10-06
🐛 Fix missing cnquery dependencies and docs

czunker:


policy/policy.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.28.1
-// 	protoc        v3.21.7
+// 	protoc        v3.21.5
```
chris-rock: We should remove this change, you are using an old protoc compiler

.gitignore
```diff
@@ -0,0 +1 @@
+cnquery
```
chris-rock, line 1: Should this be `./cnquery`?

czunker: For me, only this or `/cnquery` works.

policy/policy.pb.go
czunker: Fixed.

PR #52, opened 2022-10-07
🧹 testing for mql bundle load

chris-rock:
Signed-off-by: Christoph Hartmann <chris@lollyrock.com>

### DIFF
--- a/policy/examples/directory/example1.mql.yaml
+++ b/policy/examples/directory/example1.mql.yaml
@@ -1,0 +1,16 @@
+policies:
+  - uid: example1
+    name: Example policy 1
+    version: 1.0.0
+    authors:
+      - name: Mondoo
+        email: hello@mondoo.com
+    specs:
+      - asset_filter:
+          query: platform.family.contains(_ == 'unix')
+        scoring_queries:
+          sshd-01: {}
+          sshd-02: {}
+          sshd-03: {}
+        data_queries:
+          sshd-d-1: 0
--- a/policy/examples/directory/queries/sshd-02.mql.yaml
+++ b/policy/examples/directory/queries/sshd-02.mql.yaml
@@ -1,0 +1,5 @@
+queries:
+  - uid: sshd-02
+    title: Configure the address family
+    severity: 40
+    query: sshd.config.params["AddressFamily"] == /inet|inet6|any/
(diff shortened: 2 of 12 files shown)


policy/examples/directory/example1.mql.yaml
```diff
@@ -0,0 +1,16 @@
...
+        email: hello@mondoo.com
+    specs:
+      - asset_filter:
+          query: platform.family.contains(_ == 'unix')
+        scoring_queries:
+          sshd-01: {}
+          sshd-02: {}
+          sshd-03: {}
+        data_queries:
+          sshd-d-1: 0
```
scottford-io, line 16: What is the purpose of `data_queries` with the introduction of Query Packs?

policy/examples/directory/queries/sshd-02.mql.yaml
```diff
@@ -0,0 +1,5 @@
+queries:
+  - uid: sshd-02
+    title: Configure the address family
+    severity: 40
+    query: sshd.config.params["AddressFamily"] == /inet|inet6|any/
```
scottford-io, line 5: Should our examples contain docs examples?

```suggestion
queries:
  - uid: sshd-02
    title: Configure the address family
    severity: 40
    docs:
      desc: |
        # Add an optional description.
      audit: |
        # Add optional steps to audit within an environment
      remediation: |
        # Add optional remediation steps
     refs: 
       # Add links to external references
       - title: External reference
         url: https://link-to-reference.com
    query: sshd.config.params["AddressFamily"] == /inet|inet6|any/
```

policy/examples/directory/example1.mql.yaml
arlimus: When you write a policy, you may likely want to collect data as well, to share context for security. For example: a policy about postgres will likely want to collect its configuration as well. I thought it was going to be more helpful to allow people to add their queries here, instead of forcing them into a separate query pack.

Ultimately both policies and query packs help collect useful queries for us. Some of these are for testing and some are for generating insights. If we think of these files as a way to get queries into our platform, then it's heplful to allow either type to be accessible.

policy/examples/directory/queries/sshd-02.mql.yaml
arlimus: That's a great idea! I only wanted to leave out the docs from the most basic example we have and agree we want to add them here. That said, I'd love to get some words in instead of the comments, because these examples would be awesome to test out and run

PR #59, opened 2022-10-10
✨ mqueries.RefreshChecksums + migrate tests

arlimus:
Also exposes a few more methods we want to share on the interface.

### DIFF
--- a/policy/mquery_test.go
+++ b/policy/mquery_test.go
@@ -1,0 +1,45 @@
+package policy
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMquery_Whitespaces(t *testing.T) {
+	mq := Mquery{
+		Query: "  mondoo { version \n}   \t\n  ",
+	}
+
+	mqexpect := Mquery{
+		Query: "mondoo { version \n}",
+	}
+
+	bundle, err := mq.RefreshChecksumAndType(nil)
+	assert.NoError(t, err)
+	assert.NotNil(t, bundle)
+
+	bundle, err = mqexpect.RefreshChecksumAndType(nil)
+	assert.NoError(t, err)
+	assert.NotNil(t, bundle)
+
+	assert.Equal(t, mqexpect.CodeId, mq.CodeId)
+}
+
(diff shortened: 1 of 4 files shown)


policy/mquery_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMquery_Whitespaces(t *testing.T) {
+	mq := Mquery{
+		Query: "  mondoo { version \n}   \t\n  ",
```
scottford-io, line 11: @arlimus should this be `mondoo` or `cnspec`?

policy/mquery_test.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	mqexpect := Mquery{
+		Query: "mondoo { version \n}",
```
scottford-io, line 15: Same with this...`mondoo` or `cnspec`?

policy/mquery_test.go
```diff
@@ -0,0 +1,45 @@
...
+	bundle, err = mqexpect.RefreshChecksumAndType(nil)
+	assert.NoError(t, err)
+	assert.NotNil(t, bundle)
+
+	assert.Equal(t, mqexpect.CodeId, mq.CodeId)
+}
+
+func TestMquery_CodeIDs(t *testing.T) {
+	mqAssetFilter := Mquery{
+		Query: "mondoo { version \n}",
```
scottford-io, line 31: Same with this...`mondoo` or `cnspec`?

policy/mquery_test.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	mqReg := Mquery{
+		Query: "mondoo { version \n}",
```
scottford-io, line 35: Same with this...`mondoo` or `cnspec`?

policy/mquery_test.go
chris-rock: 100% that we need to migrate the `mondoo` resource. I think the better name is `mql` since the resource is available in cnquery and cnspec.

Since this PR just adds tests, I propose we do this change in a followup PR

PR #77, opened 2022-10-12
🌈 prefix score indication on cli reports

arlimus:
image

^^ vs the previous version where it was added after the title


chris-rock (approved):
Merci

PR #103, opened 2022-10-15
⭐ restore scan -o report

arlimus:
Note: After the migration we are loosing some context from policy bundles, so we don't always have the asset policy available in the bundle. There is a new shared call in the resolved policy, which can help find the top policies given any bundle. This also means that we can hand it any random bundle we are about and given a resolved policy find the root policies that were pulled from the bundle. It helps render policies in order i.e. embedded policies are can be tackled further down. We only need this approach when looking at policies in their hierarchy, as for everything else we can just take the bundle and the results. Since the report displays policies in their hierarchy, it required these new calls.

image

### DIFF
--- a/cli/components/barchart.go
+++ b/cli/components/barchart.go
@@ -1,0 +1,214 @@
+package components
+
+import (
+	"fmt"
+	"math"
+	"strings"
+
+	"github.com/muesli/termenv"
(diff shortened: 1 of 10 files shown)


chris-rock (approved):
Super cool. Thank you @arlimus

PR #108, opened 2022-10-16
✨ specify scoring system via name

arlimus:
image


policy/policy.go
```diff
@@ -556,3 +559,27 @@ func IsPolicyMrn(candidate string) error {
...
+	var code int32
+	err := json.Unmarshal(data, &code)
+	if err == nil {
+		*s = ScoringSystem(code)
+	} else {
+		var name string
+		_ = json.Unmarshal(data, &name)
+
+		switch name {
+		case "lowest":
```
chris-rock: Why not "worst"?

chris-rock (approved):
human readable scoring system, yes

policy/policy.go
arlimus: Sounded like "worst scoring system". I've always regretted that name ever since choosing it, but I may also be wrong about it...

arlimus: Changed based on more feedback, see new screenshot. Thank you for raising it!

chris-rock (approved):
Thank you @arlimus

PR #110, opened 2022-10-16
🐛  do not use upstream when incognito is active

chris-rock:
fixes issue reported by @arlimus where `cnspec scan local -f policy/examples/example.mql.yaml` has failed when upstream was defined.


arlimus (approved):
Thank you for the rapid bugfix!

PR #122, opened 2022-10-17
📃 improve readme structure

arlimus:
Great feedback from @IronMania: We are mentioning custom policies very early in the process, where it's more helpful to talk about scaling and the shell, before diving into custom policies. Restructured the readme and moved the custom policy part to the back. We are going to create a shorter develop-centric policy-writing guide as well, which will serve as a more lightweight intro.

Additionally move  out of the  subfolder, so they are found more easily by just looking at the repo.


tas50 (approved):

arlimus:
A few more follow-ups from our conversation that are going to land very soon:

- [ ] we have great output in the tool; use screenshots and gifs in the readme to show that off
- [ ] update the cnspec-policies repo to also point people to the guide on writing custom policies
- overall: share more examples that show off MQL bits for smaller exciting use-cases (both cnquery+cnspec), as it's so easy to use, but not well-known

PR #124, opened 2022-10-18
🧹 enable bool assertions by default

jaym:


arlimus (approved):
Very cool!! Thank you Jay, glad to have this as the default

PR #128, opened 2022-10-18
🐛 set cnquery version when we build cnspec

chris-rock:
Before we have not properly set the build flags for cnquery


arlimus (approved):
nice! kudos :)

PR #131, opened 2022-10-19
🐛 Show vanished resources as unscored

czunker:
We now check for a specific error type, to check whether a k8s resource vanished.

### DIFF
--- a/policy/executor/internal/nodes.go
+++ b/policy/executor/internal/nodes.go
@@ -5,7 +5,7 @@ import (
 
 	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnquery/llx"
-	k8s_common "go.mondoo.com/cnquery/resources/packs/k8s"
+	"go.mondoo.com/cnquery/resources"
 	"go.mondoo.com/cnquery/types"
 	"go.mondoo.com/cnspec/cli/progress"
 	"go.mondoo.com/cnspec/policy"
@@ -320,7 +320,7 @@ func (nodeData *ReportingQueryNodeData) score() *policy.Score {
 		}
 
 		if cur.Data.Error != nil {
-			var k8sNotFoundErr *k8s_common.K8sResourceNotFound
+			var k8sNotFoundErr *resources.ResourceNotFound
 			if errors.As(cur.Data.Error, &k8sNotFoundErr) {
 				assetVanishedDuringScan = true
 			} else {
--- a/policy/executor/internal/nodes.go
+++ b/policy/executor/internal/nodes.go
@@ -320,7 +320,7 @@ func (nodeData *ReportingQueryNodeData) score() *policy.Score {
 		}
 
 		if cur.Data.Error != nil {
-			var k8sNotFoundErr *k8s_common.K8sObjectNotFound
+			var k8sNotFoundErr *k8s_common.K8sResourceNotFound
 			if errors.As(cur.Data.Error, &k8sNotFoundErr) {
 				assetVanishedDuringScan = true
 			} else {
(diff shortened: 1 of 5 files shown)


czunker:
I have a better idea of where this should be handled. I just have to find the spot in the code.

policy/executor/internal/nodes.go
```diff
@@ -315,6 +318,16 @@ func (nodeData *ReportingQueryNodeData) score() *policy.Score {
 			break
 		}
 
+		if cur.Data.Error != nil && strings.HasSuffix(cur.Data.Error.Error(), "not found") {
```
jaym: i feel like it would be possible for a lot of different errors to have a "not found" error that doesn't mean the asset went away

jaym: perhaps we can define some common errors and check for those. If the asset is disconnected, the provider should return that error type

policy/executor/internal/nodes.go
```diff
@@ -347,7 +360,16 @@ func (nodeData *ReportingQueryNodeData) score() *policy.Score {
 	}
 
 	if allFound {
-		if foundError {
+		if assetVanishedDuringScan {
```
preslavgerchev, line 359: In general, a suggestion: 
I feel like we should be able to mark assets themselves as ephemeral (can tie either with asset type or provider type maybe?) and act accordingly. 
I think a better way to fix this would be when starting the scan: if the asset does not exist then, we should be able to report accordingly upstream.

policy/executor/internal/nodes.go
czunker: I've changed the error message where it is created to be more precise: https://github.com/mondoohq/cnquery/pull/361

I think, now that I have the point where the message is created, I could also create a specific error type out of it.

policy/executor/internal/nodes.go
czunker: > we should be able to mark assets themselves as ephemeral

That sound like a much bigger topic

> I think a better way to fix this would be when starting the scan: if the asset does not exist then, we should be able to report accordingly upstream.

I'm not sure I understand your point completely. Would changing https://github.com/mondoohq/cnquery/pull/361 into a specific error type be what you had in mind?

preslavgerchev: I totally agree it's a bigger topic, for sure not something that can be done quickly.

I was thinking more along the lines of having a specific error type that can be thrown if the resource does not exist rather than relying on the message inside the error

czunker: I implemented a k8s specific error type in cnquery and check for it: https://github.com/mondoohq/cnquery/pull/361

policy/executor/internal/nodes.go
czunker: Changed.

czunker:
Tests will only pass when https://github.com/mondoohq/cnquery/pull/361 is merged and updated here.

policy/executor/internal/nodes.go
```diff
@@ -316,8 +320,13 @@ func (nodeData *ReportingQueryNodeData) score() *policy.Score {
 		}
 
 		if cur.Data.Error != nil {
-			allSkipped = false
-			foundError = true
+			var k8sNotFoundErr *k8s_common.K8sResourceNotFound
```
jaym: we should avoid making this code dependent on k8s specific logic. Can we create an error of type asset disconnected or something like that

czunker: What about this? https://github.com/mondoohq/cnquery/pull/380/files

czunker: Updated the PR to include the new cnquery. Now I check for the general error.

preslavgerchev (approved):

PR #142, opened 2022-10-21
⭐️ use upstream configuration if credentials are provided

chris-rock:
This allows the upstream communication and reporting of results as well as the use of the Mondoo Vulnerability Database. `cnspec` supports running the Vulnerability Policy and the Platform End-of-Life Policy.

Screenshot 2022-10-21 at 16 15 09

Screenshot 2022-10-21 at 16 14 48


arlimus (approved):
So awesome! For anyone running into this: The idea is for upstream to be any data collection destination that can follow the given protos. We added one by default that we maintain, but others are possible

PR #143, opened 2022-10-21
⭐️ new ascii art in readme

chris-rock:


arlimus (approved):
:+1: I think later on we want to have some nicer logos up top or even lead more with the examples.

PR #145, opened 2022-10-22
⭐️ allow user to upload policy bundles to query hub

chris-rock:
With this changed, authenticated users can upload policy bundles to Mondoo Platform

```
$ cnspec bundle upload example-policy.mql.yaml
→ loaded configuration from /Users/chris/.config/mondoo/mondoo.yml using source default
→ load policy bundle file=example-policy.mql.yaml
→ valid policy bundle
→ add policy bundle to space space=//captain.api.mondoo.app/spaces/musing-saha-952142
→ successfully added policies
```

### DIFF
--- a/apps/cnspec/cmd/bundle.go
+++ b/apps/cnspec/cmd/bundle.go
@@ -7,13 +7,19 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


arlimus (approved):
Yay, let's use that API!!

PR #161, opened 2022-11-03
🐛 use score-threshold param after scan

joelddiaz:
The --score-threshold param wasn't being used to determine whether or not the cnspec binary should have a non-zero exit based on the provided --score-threshold parameter.

Walk through the scan results to calculate the worst score and use that to determine whether or not cnspec should exit non-zero.


czunker (approved):
LGTM

PR #168, opened 2022-11-06
🧹 add cnspec version and build to client headers on upstream scan

vjeffrey:
much like we do for cnquery, https://github.com/mondoohq/cnquery/pull/431/files


chris-rock (approved):
Thank you @vjeffrey, lets see if we can reuse more code going forward.

PR #170, opened 2022-11-08
🐛 fix cnspec build

chris-rock:
follow up on #168


imilchev (approved):
Just tried building with this branch and it works! Thanks!

PR #172, opened 2022-11-08
🐛 Set container image platform before calling SynchronizeAssets

imilchev:
Discussed with @preslavgerchev and @jaym and we agreed that we should add a temporary fix for this problem until we have a more long term solution. Please see if you can come up with any other type of asset that might have the same issue so I can add it to the function that fixes the platform information

### DIFF
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -155,38 +155,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstreamConf
 		return nil, false, errors.New("could not find an asset that we can connect to")
 	}
 
-	// sync assets
-	if upstreamConfig.ApiEndpoint != "" && !upstreamConfig.Incognito {
-		log.Info().Msg("syncing assets")
-		upstream, err := policy.NewRemoteServices(upstreamConfig.ApiEndpoint, upstreamConfig.Plugins)
-		if err != nil {
-			return nil, false, err
-		}
-
-		resp, err := upstream.SynchronizeAssets(ctx, &policy.SynchronizeAssetsReq{
-			SpaceMrn: upstreamConfig.SpaceMrn,
-			List:     assetList,
-		})
-		if err != nil {
-			return nil, false, err
-		}
-		log.Debug().Int("assets", len(resp.Details)).Msg("got assets details")
-		platformAssetMapping := make(map[string]*policy.SynchronizeAssetsRespAssetDetail)
-		for i := range resp.Details {
-			log.Debug().Str("platform-mrn", i).Str("asset", resp.Details[i].AssetMrn).Msg("asset mapping")
-			platformAssetMapping[i] = resp.Details[i]
-		}
-
-		// attach the asset details to the assets list
-		for i := range assetList {
-			log.Debug().Str("asset", assetList[i].Name).Strs("platform-ids", assetList[i].PlatformIds).Msg("update asset")
-			platformId := assetList[i].PlatformIds[0]
-			assetList[i].Mrn = platformAssetMapping[platformId].AssetMrn
-			assetList[i].Url = platformAssetMapping[platformId].Url
-			// only needed for CI/CD, can be removed when fleet and CI/CD have the same platform MRN
-			assetList[i].PlatformIds = []string{platformAssetMapping[platformId].PlatformMrn}
-		}
-	} else {
+	if upstreamConfig.ApiEndpoint == "" || upstreamConfig.Incognito {
 		// ensure we have non-empty asset MRNs
 		for i := range assetList {
 			cur := assetList[i]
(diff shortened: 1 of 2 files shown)


policy/scan/local_scanner.go
```diff
@@ -545,3 +555,32 @@ func (s *localAssetScanner) UpdateFilters(filters *policy.Mqueries, timeout time
 
 	return queries, err
 }
+
+func setPlatforms(ctx context.Context, assetList []*asset.Asset, im inventory.InventoryManager, record bool) error {
+	for i := range assetList {
+		a := assetList[i]
+		if a.Platform.Kind == providers.Kind_KIND_CONTAINER_IMAGE {
```
preslavgerchev: Nit: can we put the kind in a list, so building upon it is easier?

chris-rock: This makes the scan much slower because we pull that image twice right?

chris-rock:
I do not think we should do that approach if we do not cache the image

policy/scan/local_scanner.go
imilchev: Yes, we pull the image twice now. One option to not pull twice is to re-use the connection from this function for the actual scan. However, that would cause memory issues when running this code in a container. We had that before and the container ended up using gigabytes of memory because it was stacking images on the disk.

The other option is to call `SynchronizeAssets` again after the scan is run for every asset that is affected. I think there was the intention to do this before since there is a commented out line in the code right now that is tracking assets that need to be re-synced

policy/scan/local_scanner.go
```diff
@@ -159,6 +162,14 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstreamConf
 		if err != nil {
 			return nil, false, err
 		}
+
+		// TODO: this is a temporary fix for populating the platform information for assets that only set it
+		// during a scan, e.g. container images. This should be removed once we start using MQL for retrieving
+		// platform information.
+		if err := setPlatforms(ctx, assetList, im, job.DoRecord); err != nil {
```
jaym: this is problematic. Opening a connection can be very expensive. I think it would be better if you opened the connection, get the platform id, call sync asset, then scan the asset.

I think ebs scanning is one of the things that will suffer from this. cc @vjeffrey

imilchev: I could do that. The only downside is that there will be 2 code paths that do asset synchronization

jaym: i'm not sure i understand why there would be 2

imilchev: I changed it. Not sure if that is exactly what you had in mind

vjeffrey: what is the original problem this is all fixing? is it that some assets do not get platform information until they are scanned?

preslavgerchev: @vjeffrey yup, previously we had a `SyncAssets` call after the scan was done that was populating that info (pre-v7). Thats no longer there

vjeffrey: i think i'm still unclear, bc there also other cases where we don't have the platform info, right? e.g. aws with ec2 discovery, we likely won't be able to connect directly to all the instances, so a bunch of them won't have the platform info unless we scan the instance directly.

imilchev: This is not the latest change. In the latest change the assets which platform changes are being sync-ed

vjeffrey: 👍  got it, so basically anything that doesn't get platform information during the first sync call never gets its platform information in the console, because we're no longer calling sync/set asset as part of the scan path. 

just tested this with ebs volume scanning, it definitely suffers from the same issue. a lot of stuff is going to suffer from this (ec2 instances that are discovered along with an aws account, for example)

policy/scan/local_scanner.go
```diff
@@ -261,7 +272,15 @@ func (s *LocalScanner) RunAssetJob(job *AssetJob) {
 					log.Warn().Err(err).Msg("failed to query platform information")
 				} else {
 					job.Asset.Platform = p
-					// resyncAssets = append(resyncAssets, assetEntry)
+					if upstream != nil {
+						_, err := upstream.SynchronizeAssets(job.Ctx, &policy.SynchronizeAssetsReq{
```
vjeffrey: is there a more minimal version of synchronise assets that we could call?

imilchev: Not that I could find. The goal is to have this temporarily until we start using MQL to retrieve platform information. That's why I added this as a quick fix until we plan out the details for the long term solution.

jaym: do we need to call synchornize assets twice for each asset? can we call it once after we have the connection and collected the platform information?

vjeffrey:
we had a mtg about this.

current idea:
- we already get the platform information on line 274ish, so instead of calling synchronise assets once with all the assets, we're going to call it for each asset, once we have the platform info (should also resolve a timeout that was observed)
- follow up issue will be to look into batching for that (issue will be added to internal bugs board)

imilchev:
I updated the code according to what we discussed. Also opened a follow-up issue for looking into batching `SynchronizeAssets` calls https://github.com/mondoohq/cnspec/issues/197

preslavgerchev (approved):

PR #178, opened 2022-11-10
⭐️ Release rootless containers

### LINKED ISSUE #176: Add non-root container images
**Is your feature request related to a problem? Please describe.**
The container images currently run as root by default:
```
docker run -ti --entrypoint=sh mondoo/cnspec                                                                                                                                                                                            / # id
uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
```

**Describe the solution you'd like**
They should run with a non-root user by default.

**Describe alternatives you've considered**
-

**Additional context**
-

imilchev:
Note that this also makes sure it is no longer required to manually update the `Dockerfile` for every release. Now the version is set via a build arg that is set by goreleaser

Fixes #176

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -95,6 +95,7 @@ dockers: # https://goreleaser.com/customization/docker/
       - "--label=org.opencontainers.image.title={{.ProjectName}}"
       - "--label=org.opencontainers.image.revision={{.FullCommit}}"
       - "--label=org.opencontainers.image.version={{.Version}}"
+      - "--build-arg=VERSION={{ .Version }}"
   - use: buildx
     goos: linux
     goarch: arm64
@@ -108,6 +109,7 @@ dockers: # https://goreleaser.com/customization/docker/
       - "--label=org.opencontainers.image.title={{.ProjectName}}"
       - "--label=org.opencontainers.image.revision={{.FullCommit}}"
       - "--label=org.opencontainers.image.version={{.Version}}"
+      - "--build-arg=VERSION={{ .Version }}"
   - use: buildx
     goos: linux
     goarch: arm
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -136,6 +139,66 @@ dockers: # https://goreleaser.com/customization/docker/
...
+      - "mondoo/{{ .ProjectName }}:{{ .Version }}-amd64-rootless"
+      - "mondoo/{{ .ProjectName }}:{{ .Major }}-amd64-rootless"
+      - "mondoo/{{ .ProjectName }}:latest-amd64-rootless"
+    build_flag_templates:
+      - "--platform=linux/amd64"
+      - "--label=org.opencontainers.image.created={{.Date}}"
+      - "--label=org.opencontainers.image.title={{.ProjectName}}"
+      - "--label=org.opencontainers.image.revision={{.FullCommit}}"
+      - "--label=org.opencontainers.image.version={{.Version}}"
+      - "--build-arg=VERSION={{ .Version }}-rootless"
```
czunker, line 157: Nice!

.goreleaser.yml
```diff
@@ -95,6 +95,7 @@ dockers: # https://goreleaser.com/customization/docker/
       - "--label=org.opencontainers.image.title={{.ProjectName}}"
       - "--label=org.opencontainers.image.revision={{.FullCommit}}"
       - "--label=org.opencontainers.image.version={{.Version}}"
+      - "--build-arg=VERSION={{ .Version }}"
```
czunker, line 98: Am I correct, this implies we always need a cnquery image of the same version?

So, e.g., a cnspec only bugfix like v7.3.1 wouldn't be possible without previously generating a v7.3.1 cnquery image?

imilchev: yes, it's seems like it. But from what I understood when we install cnquery and cnspec, they will be part of the same installer (e.g. the MSI for windows). If that is the case, I think this should be fine as well because we will always release them together?

My guess is that we can work with this for now and if we run into troubles releasing where we don't want to release both cnspec and cnquery, we can re-evaluate

czunker (approved):

.goreleaser.yml
chris-rock: I agree let`s see if we run into the issue where we need to use a different version.

chris-rock (approved):

PR #184, opened 2022-11-11
🐛 avoid progressbar on non-interactive CLI

arlimus:
it was printing into JSON output, try not to do that


chris-rock (approved):
Thank you @arlimus

PR #191, opened 2022-11-15
🧹 use impact instead of severity

chris-rock:


chris-rock:
Will be solved with a follow-up by @arlimus

Issue #198, opened 2022-11-17
Scan errors are not reported upstream


imilchev:
**Describe the bug**
When an error occurs while scanning an asset, the error is not reported upstream. The asset remains unscored and there is no message containing the scan error. This is something that Mondoo already does.

**To Reproduce**
1. Scan an asset with an error (we did it with EBS scanning)
2. Open the UI
3. See the asset is U and has no error message

**Expected behavior**
The asset should be X and the error should be there

czunker:
I can reproduce this with this command:
```
[ec2-user@ip-172-31-26-52 ~]$ ./cnspec --config creds.yaml scan aws --sudo ec2 ebs volume vol-0aa3885f0c175286b
→ loaded configuration from creds.yaml using source --config
→ using service account credentials
→ discover related assets for 1 asset(s)
! resolver does not support discovery target 'auto', the following are supported:  resolver="AWS EC2 EBS Resolver"
→ resolved assets resolved-assets=1
→ syncing assets
→ connecting to asset vol-0aa3885f0c175286b
! WARNING: you are using an early access feature feature=AWS_EC2_EBS status=experimental
→ validate exists volume={"AccountId":"921877552404","Id":"vol-0aa3885f0c175286b","PlatformId":"//platformid.api.mondoo.app/runtime/aws/ec2/v1/accounts/921877552404/regions/us-east-1/volumes/vol-0aa3885f0c175286b","Region":"us-east-1"}
! volume specified is not in available state
→ find recent snapshot
→ found snapshot snapshot={"Account":"921877552404","Id":"snap-0a10f84960c8edd0c","Region":"us-east-1"}
→ checking snapshot region scanner instance=us-east-1 snapshot=us-east-1
→ create volume
→ waiting for volume creation completion; sleeping 10 seconds state=creating
→ attach volume volume id=vol-061417dec6c385637
→ waiting for volume attachment completion state=in-use
→ create tmp scan dir
→ search for target volume
x unable to complete mount step error="target volume not found on instance"
→ unmount volume
x failed to unmount dir error="operation not permitted"
x failed to unmount dir error="operation not permitted"
x unable to unmount volume error="operation not permitted"
→ detach volume
→ waiting for volume detachment completion state=available
→ delete created volume
→ remove created dir
```

I also see the asset as unrated. But it seems to me this is intended. I understand the error messages such that there has no scan happened. So, the policy didn't create any errors, but the errors happened during the creation of the provider.

PR #199, opened 2022-11-18
🧹 Make consistent use of UpstreamConfig when initiating connections upstream

preslavgerchev:
use the upstream config always, that's what we base our logic on so the params should be coming from there too

Signed-off-by: Preslav <preslav@mondoo.com>


czunker (approved):
LGTM. Thanks for the cleanup.

chris-rock (approved):
Thank you @preslavgerchev

czunker:
Is this related to https://github.com/mondoohq/cnspec/issues/198 ?

chris-rock:
Unlikely, this PR cleans up use of variables and avoids side-effects. #198 is a reporting issue.

PR #208, opened 2022-11-22
🧹 fix cmd descriptions for run command

chris-rock:
aligns descriptions with the feedback from @misterpantz in https://github.com/mondoohq/cnquery/pull/497


misterpantz (commented):
beautiful :)

PR #209, opened 2022-11-24
Allow using upstream creds from scan job

preslavgerchev:
Allow the scanner to use credentials from an inventory job. Only works if that's enabled via `EnableJobCredentials`

### DIFF
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -43,7 +43,7 @@ type LocalScanner struct {
 	fetcher             *fetcher
 
 	// allows setting the upstream credentials from a job
-	useJobCredentials bool
+	allowJobCredentials bool
 	// for remote connectivity
 	apiEndpoint        string
 	spaceMrn           string
@@ -66,9 +66,9 @@ func WithPlugins(plugins []ranger.ClientPlugin) ScannerOption {
 	}
 }
 
-func EnableJobCredentials() ScannerOption {
+func AllowJobCredentials() ScannerOption {
 	return func(s *LocalScanner) {
-		s.useJobCredentials = true
+		s.allowJobCredentials = true
 	}
 }
 
@@ -476,9 +476,9 @@ func (s *LocalScanner) getUpstreamConfig(incognito bool, job *Job) resources.Ups
 	plugins := s.plugins
 	endpoint := s.apiEndpoint
 	spaceMrn := s.spaceMrn
-	jobCredentials := job.Inventory.Spec.UpstreamCredentals
 
-	if s.useJobCredentials && jobCredentials != nil {
+	jobCredentials := job.Inventory.Spec.UpstreamCredentals
+	if s.allowJobCredentials && jobCredentials != nil {
 		certAuth, _ := upstream.NewServiceAccountRangerPlugin(jobCredentials)
 		plugins = append(plugins, certAuth)
 		endpoint = jobCredentials.GetApiEndpoint()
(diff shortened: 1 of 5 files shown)


policy/scan/local_scanner.go
```diff
@@ -462,6 +468,31 @@ func (s *LocalScanner) HealthCheck(ctx context.Context, req *HealthCheckRequest)
...
+	if incognito {
+		return resources.UpstreamConfig{Incognito: true}
+	}
+
+	plugins := s.plugins
+	endpoint := s.apiEndpoint
+	spaceMrn := s.spaceMrn
+	jobCredentials := job.Inventory.Spec.UpstreamCredentals
+
+	if s.useJobCredentials && jobCredentials != nil {
```
chris-rock: semantically we could call it:

- `useJobCredentials` - in this case I would expect that all jobs need their own credentials, no fall back to default credentials
- `allowJobCredentials` - here job credentials are optional

Depending on what you want to achieve, we need to decide for the wording and then implement it accordinly

preslavgerchev: Good point. I went with the latter (`AllowJobCredentials`) as I feel that's a bit more flexible

chris-rock: Perfect, makes a lot of sense. We can restrict that further going forward if we need to

policy/scan/local_scanner.go
```diff
@@ -462,6 +468,31 @@ func (s *LocalScanner) HealthCheck(ctx context.Context, req *HealthCheckRequest)
...
+	}
+
+	plugins := s.plugins
+	endpoint := s.apiEndpoint
+	spaceMrn := s.spaceMrn
+
+	jobCredentials := job.Inventory.Spec.UpstreamCredentals
+	if s.allowJobCredentials && jobCredentials != nil {
+		certAuth, _ := upstream.NewServiceAccountRangerPlugin(jobCredentials)
+		plugins = append(plugins, certAuth)
```
chris-rock: I think this is problematic. If we have already added an authentication plugin, we add just another one. We need to split the authentication plugin from the rest of the plugins

preslavgerchev: Internally, ranger will always prefer the latest one as it builds out a map. However, I changed the plugins list to a map in the local scanner so that we can be very explicit about which header should be in the request

policy/scan/local_scanner.go
```diff
@@ -41,23 +42,38 @@ type LocalScanner struct {
...
 		s.apiEndpoint = apiEndpoint
-		s.plugins = plugins
 		s.spaceMrn = spaceMrn
 	}
 }
 
+func WithPlugins(plugins []ranger.ClientPlugin) ScannerOption {
+	return func(s *LocalScanner) {
+		for _, p := range plugins {
+			s.pluginsMap[p.GetName()] = p
```
chris-rock, line 66: nice approach

policy/scan/local_scanner.go
```diff
@@ -462,6 +479,43 @@ func (s *LocalScanner) HealthCheck(ctx context.Context, req *HealthCheckRequest)
 	}, nil
 }
 
+func (s *LocalScanner) getUpstreamConfig(incognito bool, job *Job) (resources.UpstreamConfig, error) {
+	if incognito {
+		return resources.UpstreamConfig{Incognito: true}, nil
+	}
+
+	pluginsMap := s.pluginsMap
```
chris-rock: Have you checked that this has no side-effect? I think we need to copy the keys, otherwise you manipulate the `s.pluginsMap` with the one-off credentials

preslavgerchev: Good catch, this is now a copied map so we do not touch the original one

chris-rock (approved):
Thank you @preslavgerchev

PR #210, opened 2022-11-24
✨Add --asset-name flag for scanning.

preslavgerchev:
Follow-up to the cnquery changes, allows cnspec scanning to set the asset name.


chris-rock (approved):
Woot

PR #220, opened 2022-12-02
🐛 Exit if sync assets fails

imilchev:
If an error is encountered during `SynchronizeAssets` currently cnspec panics. This PR fixes the issue


czunker (approved):
LGTM

PR #224, opened 2022-12-03
Replace Mondoo Client with cnspec in errors

tas50:
We had a few more still floating around


czunker (approved):
Nice catches.

PR #240, opened 2022-12-16
🧹 remove unhelpful warning messages

chris-rock:
Since we switch to unauthenticated as default, those warning messages are actually not helpful anymore. We display specifically when a service account is used.


arlimus (approved):
Fewer annoying messages

PR #249, opened 2022-12-27
⭐️ Add container edge releases workflow

imilchev:
Builds edge containers for every commit to `main`. Will use the latest edge release for cnquery as a base image


chris-rock (approved):
Great addition @imilchev

PR #253, opened 2022-12-28
⭐️ new`cnspec bundle lint` cmd 

chris-rock:
**Problem:**

As a developer of policies I want to verify that my developed bundle meets the requirements and conforms to best-practices.

**Solution:**

The new `cnspec bundle lint` command (previously called `cnspec bundle validate` ships with a set of rules:

- "MQL compile error"
- "UID is not valid"
- "Missing policy UID"
- "Missing policy name"
- "No unique policy UID"
- "Policy is missing checks"
- "Assigned query missing"
- "Policy version is missing"
- "Policy version is wrong"
- "Missing query UID"
- "Missing query title"
- "No unique query UID"
- "Unassigned Query"

A major improvement compared to the previous implementation is the detection of file name and line number. The allows the output to highlights the rule id and message with the file and line number. To see the new linting output, just run:

```
cnspec bundle lint policy.mql.yaml
```

Screenshot 2023-01-01 at 14 00 50

The new `cnspec bundle lint` also allows users to export the output as sarif with the `-o sarif` option. Best is to pipe the output into a file:

```
cnspec bundle lint -o sarif --output-file report.sarif policy.mql.yaml 
``` 

The report can then be viewed in Visual Studio Code and the [Sarif Extension](https://marketplace.visualstudio.com/items?itemName=MS-SarifVSCode.sarif-viewer)

Screenshot 2023-01-01 at 14 10 39

We also improved renamed `cnspec bundle upload` to `cnspec bundle publish`.


chris-rock:
cnspec bundle validate should also warn when queries are not assigned

jaym (approved):

PR #255, opened 2022-12-29
🧹 show where cnquery will be run from

joelddiaz:
Was getting wierd behavior and it was all because an old build of cnquery was being run instead of what was installed into /usr/local/bin.

Show the path when log level is debug


chris-rock (approved):
Thank you @joelddiaz

PR #256, opened 2022-12-30
🧹 Sort options for --output

jaym:
This prevents them from showing up in a random order when running help / building documentation


chris-rock (approved):
Thank you @jaym

PR #263, opened 2023-01-05
🌈 Refactor multi asset CLI summary

czunker:


czunker:
This is how it looks like for multiple k8s assets:

image


This is the output for a single asset:

image

tas50:
Tested this locally. The summary looks really great, but I'm not getting a link to the cluster when logged in.

czunker:
> Tested this locally. The summary looks really great, but I'm not getting a link to the cluster when logged in.

Before, we had asset URLs, which I think do not make sense here anymore.

I added a space URL instead:
image

czunker:
> @czunker Great job, could we try to put the two summaries (asset + score distribution) side by side at the bottom of the scan?

@arlimus Yes, we can:

image

image

czunker:
This is how the distribution for images looks like:
image

The last "1" is a CoreDNS image where we do not get the platform.

czunker:
> @czunker Those look great!
> 
> just a tiny nit-pick, thanks to your screenshot, I wonder if we want to align both columns to the left (i.e. remove the leading left space on both columns). It somehow looks misaligned on the right column because of the leading space. But maybe that's just me? What do you think @czunker @tas50 ?

This is how it looks like without the leading space:

image

And yes, seeing this, I agree. Makes it easier on the eyes.

arlimus:
Let's go with that!

arlimus (approved):
Thank you for adding this!

PR #264, opened 2023-01-06
⭐️ detect missing asset filter in policies

chris-rock:
This change detects missing asset filter:

```yaml
policies:
- uid: mondoo-azure-security
  name: Microsoft Azure Security by Mondoo
  version: 1.0.0
  specs:
  - scoring_queries:
      mondoo-azure-security-ensure-os-disk-are-encrypted: null
      mondoo-azure-security-ssh-access-restricted-from-internet: null
```

would create an error since the `spec` has no filter defined. The correct one would be:

```yaml
policies:
- uid: mondoo-azure-security
  name: Microsoft Azure Security by Mondoo
  version: 1.0.0
  specs:
  - asset_filter:
      query: |
         platform.name == "azure"
         platform.kind == "api"
    scoring_queries:
      mondoo-azure-security-ensure-os-disk-are-encrypted: null
      mondoo-azure-security-ssh-access-restricted-from-internet: null
```

### DIFF
--- a/internal/bundle/lint.go
+++ b/internal/bundle/lint.go
@@ -21,6 +21,7 @@ const (
 	policyUid                  = "policy-uid"
 	policyName                 = "policy-name"
 	policyUidUnique            = "policy-uid-unique"
+	policyMissingAssetFilter   = "policy-missing-asset-filter"
 	policyMissingAssignedQuery = "policy-missing-assigned-query"
 	policyMissingChecks        = "policy-missing-checks"
 	policyMissingVersion       = "policy-missing-version"
@@ -67,6 +68,11 @@ var rules = []Rule{
 		ID:          policyUidUnique,
 		Name:        "No unique policy UID",
 		Description: "Every policy uid must not be used twice in the same bundle and namespace",
+	},
+	{
+		ID:          policyMissingAssetFilter,
+		Name:        "Policy Spec is missing an asset filter",
+		Description: "Policy Spec has no asset filter defined",
 	},
 	{
 		ID:          policyMissingChecks,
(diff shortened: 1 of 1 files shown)


internal/bundle/lint.go
```diff
@@ -68,6 +69,11 @@ var rules = []Rule{
 		Name:        "No unique policy UID",
 		Description: "Every policy uid must not be used twice in the same bundle and namespace",
```
tas50:
```suggestion
		Description: "Every policy UID must not be used twice in the same bundle and namespace",
```

misterpantz:
```suggestion
		Description: "Do not use a policy UID twice in the same bundle and namespace.",
```

internal/bundle/lint.go
```diff
@@ -68,6 +69,11 @@ var rules = []Rule{
...
 	},
+	{
+		ID:          policyMissingAssetFilter,
+		Name:        "Policy Spec is missing an asset filter",
+		Description: "Policy Spec has no asset filter defined",
```
misterpantz:
```suggestion
		Description: "Policy Spec doesn't define an asset filter.",
```

internal/bundle/lint.go
```diff
@@ -332,6 +338,29 @@ func lintFile(file string) (*Results, error) {
...
+					location = Location{
+						File:   file,
+						Line:   spec.AssetFilter.FileContext.Line,
+						Column: spec.AssetFilter.FileContext.Column,
+					}
+				}
+
+				res.Entries = append(res.Entries, Entry{
+					RuleID:   policyMissingAssetFilter,
+					Message:  "Policy " + policy.Uid + " is missing asset filter",
```
misterpantz:
```suggestion
					Message:  "Policy " + policy.Uid + " doesn't define an asset filter.",
```

misterpantz (commented):
💜

atomic111 (approved):

PR #275, opened 2023-01-09
🧹 Use other lib to calculate width

czunker:
We already use this lib in cnquery. The other one was an additional dependency


chris-rock (approved):
Thank you @czunker

PR #278, opened 2023-01-10
🐛 Show asset url when only one asset is scanned

czunker:


czunker:
Before we only showed the space url with the summary.

Now we check whether we have only one asset, where it makes sense to show the asset url.

chris-rock (approved):
Thank you @czunker

PR #282, opened 2023-01-12
🌈 Add multi-asset progress bars

czunker:


czunker:
This requires an updated cnquery after https://github.com/mondoohq/cnquery/pull/771 is merged.

czunker:
Some examples how the output looks like:

Screenshot from 2023-01-12 08-49-33
Screenshot from 2023-01-12 08-49-14
Screenshot from 2023-01-12 08-48-00
Screenshot from 2023-01-12 08-47-22

czunker:
This currently only works in incognito mode, and needs a bugfix when working with upstream.

policy/scan/scan.proto
```diff
@@ -47,6 +47,8 @@ message Job {
   // Vault configuration + credentials query
 	// Report type configuration
   ReportType report_type = 22;
+
+  int32 progress_num_assets = 23;
```
chris-rock: This should be removed here

czunker: Good catch. Thanks.

I removed it.

chris-rock (approved):
Thank you @czunker

PR #284, opened 2023-01-12
🐛 avoid panic when no login/token available

joelddiaz:
running 'cnspec login' on a fresh system results in a panic

~  % cnspec login
→ no Mondoo configuration file provided. using defaults ! could not initialize certificate authentication error="agent credentials must be set"
panic: runtime error: invalid memory address or nil pointer dereference

avoid adding a nil client plugin so we get a more sensible error:

% ./cnspec login
→ no Mondoo configuration file provided. using defaults ! could not initialize certificate authentication error="agent credentials must be set" FTL failed to log in client error="failed to do request: Post \"https:///AgentManager/RegisterAgent\": http: no Host in request URL"


vjeffrey (approved):
i think we could do better with the err msg, but i know that wasn't really changed here, so +1

PR #301, opened 2023-01-23
🌠🛑 policy v2

arlimus:
WIP: the proto files need rebuilding after the cnquery PRs are rebased (see https://github.com/mondoohq/cnquery/pull/813 for the chain of PRs)

This change builds on top of the [query v2](https://github.com/mondoohq/cnquery/pull/713) and [properties](https://github.com/mondoohq/cnquery/pull/724) changes in  cnquery. It updates the structure for policies to be more flexible and standalone moving forward. 

It also introduces `impact` which will start deprecating `severity` moving forward. This is purely semantics, none of the underlying logic has changed. The only reason we are making this change now has been a ton of feedback from friends in the community, who pointed out that severity occupies a somewhat different meaning (that is very very tied to CVSS).

One of the biggest changes in this PR is the fact that we are deduplicating a lot of structures, most notable that of the `Mquery` (i.e. query metadata and MQL code). The core implementation now resides in cnquery and cnspec simply builds on top of it. A few others structures are affected as well, but should be largely unnoticeable.

We are also overhauling the property system and making it fully usable on the CLI. You can now specify properties via e.g.:

```
cnspec scan -f my.mql.yaml --props "homeFolder='/home/user'"
```


arlimus:
I have mostly been editing the current example to fit the new syntax:

```yaml
# This section lists all the policies that are part of this bundle.
# In our case, we only have 1 policy: example1
policies:
  - uid: example1
    name: Example policy 1
    version: "1.0.0"
    # If your policies has queries that have a "severity", you should set
    # the scoring system to "highest impact" (like in this case). It makes
    # sure that the lowest score (i.e. highest impact failure) is used as
    # the overall score for the report.
    scoring_system: highest impact
    authors:
      - name: Mondoo
        email: hello@mondoo.com
    # Specs are a way to specify all the queries (and other policies)
    # that we want to apply. Specs are grouped together and can be filtered.
    # This allows you to only apply a group of queries if the condition is met.
    specs:
      - checks:
          # These are checks that will be scored and contribute to the
          # final score of this policy.
          - uid: sshd-01
            title: Set the port to 22
            query: sshd.config.params["Port"] == 22
            # Impact are used for scoring. 100 = critical. 0 = informational.
            impact: 30

          - uid: sshd-02
            title: Configure the address family
            query: sshd.config.params["AddressFamily"] == /inet|inet6|any/
            impact: 40

            # Here we use a referenced query. You can put multiple policies
            # into a bundle, which may share theirtheir checks and data queries.
          - uid: shared

        queries:
          # These are queries which only collect data. They don't say
          # what you should or shouldn't do, they only provide insights.
          - uid: sshd-d-1
            title: Gather SSH config params
            query: sshd.config.params

          # Here is an example of a query that uses embedded properties. 
          # These allow users to fine-tune the policy.
          - uid: home-info
            mql: file(props.homeFolder) { * }
            title: Gather info about the user's home
            props:
              - uid: homeFolder
                mql: |
                  "/root"

        filter:
          # Here we specify that the queries in this spec are only applied
          # when the asset satisfies this condition:
          - mql: asset.family.contains('unix')

# These are all the queries that are part of this bundle. They are used
# by the policies specified above.
queries:
  # Every query can be identified by its UID
  # The title helps in printing it.
  - uid: shared
    title: Enable strict mode
    query: sshd.config.params["StrictModes"] == "yes"
    impact: 70
```

cli/reporter/print_compact.go
```diff
@@ -300,7 +301,7 @@ func (r *defaultReporter) printAssetSections(orderedAssets []assetMrnName) {
 
 	r.out.Write([]byte(NewLineCharacter))
 
-	var queries map[string]*policy.Mquery
+	var queries map[string]*explorer.Mquery
```
chris-rock, line 303: good to see that reused

policy/cnspec_policy.proto
```diff
@@ -33,16 +56,60 @@ message PolicySpec {
...
+
   string mrn = 1; // will be filled from UID on upload
+  string uid = 36; // uid is only used initially, to generate the MRN
   string name = 2;
   string version = 3;
+  string owner_mrn = 8;
+
+  repeated PolicySpec specs = 11;
+
+  bool is_public = 9;
```
chris-rock: I think we should deprecate is_public as well. That is not used anywhere anymore

policy/cnspec_policy.proto
```diff
@@ -33,16 +56,60 @@ message PolicySpec {
...
 
 /*
   Policy definition, re-usable or assigned to the user's environment
   - Contains all policies and overrides
   - Can limit to what it supports
 */
 message Policy {
+  // FIXME: DEPRECATED, remove in v9.0 vv
+  // This is replaced by the field "filters"
+  map<string,Mquery> v1_asset_filters = 7; // key == checksum, internal
```
chris-rock: I am not sure I missed it, do we have a custom unmarshaller to load old policies?

policy/cnspec_policy.proto
```diff
@@ -33,16 +56,60 @@ message PolicySpec {
...
+  string license = 21;
+  PolicyDocs docs = 41;
+
+  ScoringSystem scoring_system = 10;
+
+  repeated Author authors = 30;
+  int64 created = 32;
+  int64 modified = 33;
+  map<string,string> tags = 34;
+  repeated cnquery.explorer.Property props = 35;
```
chris-rock: When we re-use the old proto buf id, are we sure we can load the old protobuf format?

chris-rock (approved):
The new structure is so much better. Love it @arlimus

PR #302, opened 2023-01-23
⭐️ junit support as reporter

chris-rock:
**Problem**

I want to see my scan results in Jenkins

**Solution**

`cnspec` outputs the results as junit format now. To get the junit output, set the output format to junit and pipe the results to a file:

```
cnspec scan docker debian:10 --output junit > report.junit
```

The result looks like the following then:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
        <testsuite name="Policy Report for debian:10@edcf96f9d9d9" tests="85" failures="43" errors="0" id="0" time="">
                <testcase name="Ensure auditd is installed" classname="score">
                        <failure message="results do not match" type="fail"></failure>
                </testcase>
                <testcase name="Ensure no duplicate UIDs exist" classname="score"></testcase>
                <testcase name="Ensure root group is empty" classname="score"></testcase>
                <testcase name="Ensure no duplicate group names exist" classname="score"></testcase>
                <testcase name="Ensure source routed packets are not accepted" classname="score">
                        <failure message="results do not match" type="fail"></failure>
                </testcase>
                ...
                <testcase name="Ensure login and logout events are collected" classname="score">
                        <failure message="results do not match" type="fail"></failure>
                </testcase>
        </testsuite>
</testsuites>
```


vjeffrey:
works great locally!

vjeffrey (approved):

PR #306, opened 2023-01-25
🧹 Use the refactored multiprogress

czunker:
This depends on https://github.com/mondoohq/cnquery/pull/831

### DIFF
--- a/policy/executor/graph.go
+++ b/policy/executor/graph.go
@@ -6,7 +6,6 @@ import (
 	"go.mondoo.com/cnquery"
 	"go.mondoo.com/cnquery/cli/progress"
 	"go.mondoo.com/cnquery/llx"
-	"go.mondoo.com/cnquery/motor/asset"
 	"go.mondoo.com/cnquery/mqlc"
 	"go.mondoo.com/cnquery/resources"
 	"go.mondoo.com/cnspec/policy"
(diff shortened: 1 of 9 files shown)


policy/executor/graph.go
```diff
@@ -18,16 +18,16 @@ type GraphExecutor interface {
 }
 
 func ExecuteResolvedPolicy(schema *resources.Schema, runtime *resources.Runtime, collectorSvc policy.PolicyResolver, asset *asset.Asset,
-	resolvedPolicy *policy.ResolvedPolicy, features cnquery.Features, progressProg progress.Program,
+	resolvedPolicy *policy.ResolvedPolicy, features cnquery.Features, multiProgressBar progress.MultiProgress,
```
jaym: I might be missing something here, but why does the execution need to care that its a multi progress interface. It can only execute for one thing. Something else that cares about displaying multiple progress bars can deal with doing whatever necessary transform to make it work with multiple progress bars. For example:

```
type Progress interface {
  OnProgress(percent float64)
}

type MultiProgressAdapter struct {
  multi MultiProgress
  key string
}

func (m* MultiProgressAdapter) OnProgress(percentage float64) {
  m.multi.OnProgress(key, percentage)
} 

for _, asset := range assets {
  p := MultiProgressAdapter{key: asset.PlatformIds[0], multi: multiProgressBar)
  ExecuteResolvedPolicy(asset.Mrn, ..., p,...)
}
```

czunker: Thanks for explaining this to me. I re-worked the cnquery and cnspec PRs to re-use the extended progress.Progress interface.

policy/scan/local_scanner.go
```diff
@@ -255,33 +253,34 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstreamConf
 			select {
 			case <-ctx.Done():
 				log.Warn().Msg("request context has been canceled")
+				// When we scan concurrently, we need to call Errored(asset.Mrn) status for this asset
+				multiprogress.Close()
```
jaym, line 257: i might have missed this in the cnquery review, but is it ok for close to get called for each asset?

jaym (approved):

policy/scan/local_scanner.go
czunker: As I understood this part, currently, we cancel the complete scan. But yes, we need to change this as soon as we migrate to concurrent scans.

PR #308, opened 2023-01-26
🐛 Handle motor/sync error for progress bar

czunker:

### DIFF
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -313,6 +313,8 @@ func (s *LocalScanner) RunAssetJob(job *AssetJob) {
 				log.Error().Err(err).Msgf("failed to synchronize asset to Mondoo Platform %s", job.Asset.Mrn)
 			}
 		}
+		job.ProgressReporter.Score("X")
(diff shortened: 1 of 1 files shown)


policy/scan/local_scanner.go
```diff
@@ -313,6 +313,8 @@ func (s *LocalScanner) RunAssetJob(job *AssetJob) {
 				log.Error().Err(err).Msgf("failed to synchronize asset to Mondoo Platform %s", job.Asset.Mrn)
 			}
 		}
+		job.ProgressReporter.Score("X")
```
chris-rock, line 316: This code should be tied to the `job.Reporter.AddScanError(job.Asset, err)` because it always need to go together.

chris-rock (approved):

policy/scan/local_scanner.go
czunker: Will have a look at it in another PR.

czunker: Follow-up https://github.com/mondoohq/cnspec/pull/313

PR #311, opened 2023-01-27
🧹 Update the gifs + vhs tapes

tas50:
100% these tapes are for system, but this gives the next person a starting point and is still a significant improvement over the current manual setup


chris-rock (approved):
Thank you @tas50

PR #292, opened 2023-01-18
🐛 Fix conflicting file with mondoo client package

czunker:
Signed-off-by: Christian Zunker <christian@mondoo.com>

PR #318, opened 2023-02-01
🧹 Remove mondoo config example

czunker:
This is no longer part of the package.
It was previously removed from the build process with https://github.com/mondoohq/cnspec/pull/292.


chris-rock (approved):
Thank you @czunker

PR #322, opened 2023-02-05
🧹 make scoringspec v7=>v8 accessible

arlimus:
This helps with the migration of old v7 structs used in other places. So
far this method was not re-usable. Making it accessible helps with other
places that need to migrate scoring specs.


chris-rock (approved):
Thank you @arlimus

PR #323, opened 2023-02-06
🐛 re-activate and fix checksum generation on bundles

arlimus:
... using the updated cnquery methods to merge base queries and correctly re-generate their checksums.

The current approach had errors when it comes to base and embedded queries. That is because previously we would only deal with what we now call base queries and didn't have to worry too much about any overrides that were happening, because those were limited to scoring specs. However, the latter have now been replaced and merging queries has become the norm. In this world, we need correct checksums for all merged actions.

Additionally, we don't want to alter the structure of query overrides too much, i.e. we don't want to just pull the base query into the override and overwrite the policy group, since that would lose all context of a dependency existing. We will revisit some of this behavior when we get to version history of snapshotting in cnspec/cnquery, which is coming up soon.

### DIFF
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -360,7 +360,7 @@ func (p *Policy) updateAllChecksums(ctx context.Context,
 	if p.ScoringSystem == ScoringSystem_SCORING_UNSPECIFIED {
 		p.ScoringSystem = ScoringSystem_AVERAGE
 	}
-	executionChecksum = executionChecksum.Add(p.ScoringSystem.String())
+	executionChecksum = executionChecksum.AddUint(uint64(p.ScoringSystem))
 
 	// PROPS (must be sorted)
 	sort.Slice(p.Props, func(i, j int) bool {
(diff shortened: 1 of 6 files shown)


policy/policy.go
```diff
@@ -360,7 +360,7 @@ func (p *Policy) updateAllChecksums(ctx context.Context,
 	if p.ScoringSystem == ScoringSystem_SCORING_UNSPECIFIED {
 		p.ScoringSystem = ScoringSystem_AVERAGE
 	}
-	executionChecksum = executionChecksum.Add(p.ScoringSystem.String())
+	executionChecksum = executionChecksum.AddUint(uint64(p.ScoringSystem))
```
arlimus, line 363: fyi: we should never depend on the string here; the underlying type is an enum, use it

policy/resolver.go
```diff
@@ -273,7 +273,6 @@ func (s *LocalServices) CreatePolicyObject(policyMrn string, ownerMrn string) *P
 			Policies: []*PolicyRef{},
 			Checks:   []*explorer.Mquery{},
 			Queries:  []*explorer.Mquery{},
-			Filters:  &explorer.Filters{},
```
arlimus, line 276: fyi: when creating filters either create the whole thing (with items map) or none at all; here we auto-generate the first group, but without any filters. That should be expected and to be handled downstream, since groups may not have filters (since they can be parsed this way from bundles as well)

chris-rock (approved):

PR #330, opened 2023-02-07
🧹 remove internal reference to query isScored

arlimus:
It is a remnant of old code and not needed in this context (nor used).


chris-rock (approved):
Thank you @arlimus

PR #331, opened 2023-02-07
🐛 base query creation and lookup

arlimus:
After #330 

After the recent change on merging queries (where we now don't overwrite the object anymore) we have been missing the fully realized queries during the resolver step. Restore this by providing both the base queries (in the bundle) and the policy queries.

Fixes the basic query execution that failed after that changed merged in cnquery. Also has the nice side-effect that queries aren't (tried to be) created multiple times if they are referenced by multiple policies in a bundle.


chris-rock (approved):
Thank you @arlimus

PR #338, opened 2023-02-08
🐛 fix asset score aggregation

arlimus:
this is a problem that was introduced with the changes for policy v2. The unset value for weight is -1, not 0.


scottford-io (approved):
@arlimus Great stuff! Excited to keep improving our scoring!! ⭐ 🙌

PR #342, opened 2023-02-10
🐛+🧹 fix impact aggregation + scoringspec support

arlimus:
- the aggregation of impact values has caused e.g. the asset score to become an A, instead of what it actually should be. This is because the default Impact.Value was set to 0, which translates into an INFO score. To remedy this issue, we set the default impact.value to -1 as it should be (and as it is for parsing e.g. bundles)
- additionally, we were missing values for the scoringspec in reportingjobs, which is still in use for v7 clients and will be throughout v8. Make sure we continue to support them


chris-rock (approved):
Thank you @arlimus

PR #346, opened 2023-02-13
🧹 post-install script of package migrates mondoo service

chris-rock:
if a running mondoo service is detected, the cnspec service is used as a replacement

### DIFF
--- a/scripts/pkg/debian/postinstall.sh
+++ b/scripts/pkg/debian/postinstall.sh
@@ -6,29 +6,40 @@ cnspec --config /etc/opt/mondoo/mondoo.yml migrate
 
 if [ "$(cat /proc/1/comm)" = "init" ]
 then
+  # Upstart best-effort to just stop mondoo service and
+  # start cnspec service. Removing mondoo from upstart
+  # will not be scripted here.
+  # if mondoo service is running, stop it and start cnspec
+  test -n "$(initctl status mondoo | grep -o running)" && \
+    echo " -> Stop mondoo service; start cnspec service (init)" && \
+    initctl stop mondoo && \
+    initctl start cnspec
+
   test -n "$(initctl status cnspec | grep -o running)" && \
     echo " -> Restart cnspec service (init)" && \
     initctl restart cnspec
 
-  # if mondoo service is running, stop it and start cnspec
-  test -n "$(initctl status mondoo | grep -o running)" && \
-    echo " -> Replace mondoo service with cnspec service (init)" && \
-    initctl stop mondoo && \
-    initctl start cnspec
-
 elif [ "$(cat /proc/1/comm)" = "systemd" ]
 then
+  # if Mondoo service is currently running, make sure
+  # cnspec service is running to replace it
+  systemctl is-active --quiet mondoo && \
+    echo " -> Mondoo service running, replacing with cnspec (systemd)" && \
+    systemctl stop mondoo && \
+    systemctl daemon-reload && \
+    systemctl restart cnspec
+
+  # if Mondoo service is set up to run on boot
+  # replace it with cnspec to start on boot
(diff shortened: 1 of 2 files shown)


scripts/pkg/debian/postinstall.sh
```diff
@@ -9,12 +9,26 @@ then
   test -n "$(initctl status cnspec | grep -o running)" && \
     echo " -> Restart cnspec service (init)" && \
     initctl restart cnspec
+
+  # if mondoo service is running, stop it and start cnspec
+  test -n "$(initctl status mondoo | grep -o running)" && \
+    echo " -> Replace mondoo service with cnspec service (init)" && \
+    initctl stop mondoo && \
```
joelddiaz, line 15: Do we not need an equivalent to `systemctl disable mondoo` for non-systemd init so that the service doesn't start up on boot any more?

scripts/pkg/debian/postinstall.sh
```diff
@@ -9,12 +9,26 @@ then
...
     echo " -> Restart cnspec service (systemd)" && \
     systemctl daemon-reload && \
     systemctl restart cnspec
+
+  systemctl is-active --quiet mondoo && \
+      echo " -> Replace mondoo service with cnspec service (systemd)" && \
+      systemctl daemon-reload && \
+      systemctl stop mondoo && \
+      systemctl disable mondoo && \
+      systemctl start cnspec
```
joelddiaz: Do we need a `systemctl enable cnspec`?

scripts/pkg/debian/postinstall.sh
chris-rock: Is there a way to do that?

scripts/pkg/debian/postinstall.sh
chris-rock: Only if mondoo service is currently enabled. We probably need another check

scripts/pkg/debian/postinstall.sh
joelddiaz: Yes, I went ahead and added two sections into the postinstall script
1) if mondoo.service is running; stop it and start cnspec.service
2) if mondoo.service is enabled for boot; disable it and enable cnspec.service

So the end of the package upgrade should leave you with cnspec running if mondoo.service was running and cnspec.service enabled if mondoo.service was previously enabled.

czunker: I understand from this conversation that the service should also be enabled for Upstart. But from the comments above and in the commit message, I read this is no longer true. I assume comments and commit message are correct?

How would we inform the user, that the service changed?
When mondoo was active as a service, wouldn't this result in a failure after the next reboot? Mondoo is trying to start as a service, but the mondoo binary isn't present any longer.

scripts/pkg/debian/postinstall.sh
```diff
@@ -6,11 +6,36 @@ cnspec --config /etc/opt/mondoo/mondoo.yml migrate
...
+
 elif [ "$(cat /proc/1/comm)" = "systemd" ]
 then
+  # if Mondoo service is currently running, make sure
+  # cnspec service is running to replace it
+  systemctl is-active --quiet mondoo && \
+    echo " -> Mondoo service running, replacing with cnspec (systemd)" && \
+    systemctl stop mondoo && \
+    systemctl daemon-reload && \
+    systemctl restart cnspec
```
czunker, line 30: nit: This will restart cnspec twice. Should we change the ordering? Same for the Upstart part of the script.

scripts/pkg/debian/postinstall.sh
joelddiaz: We chatted about Upstart a bit yesterday (Tim and me), and we didn't want to spend too much time on Upstart since there really shouldn't be any users of it left at this point (unless you know of a supported distro that still uses it).

PR #354, opened 2023-02-14
🧹Compress windows amd64 binary with upx

jaym:
This takes the size from 200MB to 50MB. It doesn't work for arm.

Some downsides may be that the binaries now take longer to start. Its also possible that this is more likely to trigger antivirus software.


arlimus (approved):
one small step, 
one giant leap in filesize

PR #358, opened 2023-02-16
🐛 fix incorrect bundle caching and checksums

arlimus:
1: I noticed that with the deprecated policy example data queries
   weren't pulled from the bundle. It turns out the code was treating
   checks and queries differently. Fixed that.

   Additionally, the bundle didn't cache the base objects. Bundles have
   a global space to expose shared queries. This is used for both
   embed-only queries as well as queries that are actually shared.
   However, the code was pushing import objects (i.e. a query that only
   has the MRN and nothing else) into the shared query space, instead of
   the actual query.

2: Checksums weren't recalculated when base queries were pulled in.
   Again, in most v8 cases this wasn't noticeable, but with v7 code this
   broke the policy execution.

   This PR introduces both a fix to checksums when queries are merged
   (just refresh it) as well as a sanity-check since these missing
   checksums are both devestating to the resolved policy and go by
   unnoticed.

Now with the v7 policy:

image

And with the v8 policy:

image

### DIFF
--- a/policy/hub.go
+++ b/policy/hub.go
@@ -319,22 +319,26 @@ func (s *LocalServices) ComputeBundle(ctx context.Context, mpolicyObj *Policy) (
 			}
 		}
 
+		// For all queries and checks we are looking to get the shared objects only.
+		// This is because the embedded queries and checks are already part of the
+		// policy and what the bundle represents in its toplevel Queries field is
+		// the collection of shared content (not its overrides). So the section
+		// below is all about adding the shared content only.
+
 		for i := range group.Queries {
 			query := group.Queries[i]
-			base, err := s.DataLake.GetQuery(ctx, query.Mrn)
-			if err == nil {
-				query.Merge(base)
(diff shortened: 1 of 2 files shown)


policy/hub.go
```diff
@@ -319,22 +319,26 @@ func (s *LocalServices) ComputeBundle(ctx context.Context, mpolicyObj *Policy) (
...
+		// policy and what the bundle represents in its toplevel Queries field is
+		// the collection of shared content (not its overrides). So the section
+		// below is all about adding the shared content only.
+
 		for i := range group.Queries {
 			query := group.Queries[i]
-			base, err := s.DataLake.GetQuery(ctx, query.Mrn)
-			if err == nil {
-				query.Merge(base)
+			if base, _ := s.DataLake.GetQuery(ctx, query.Mrn); base != nil {
```
jaym, line 330: does GetQuery only return errors if the query cannot be found?

jaym (approved):

PR #374, opened 2023-02-23
🐛 stabilize the order of v7 conversions

arlimus:
and also start to re-use sortedKeys in a few other places to sort keys

### DIFF
--- a/policy/deprecated_v7.go
+++ b/policy/deprecated_v7.go
@@ -14,12 +14,13 @@ import (
 // the old proto structures) in v9.
 
 // DeprecatedV7Conversions will find any v7 pieces in the bundle and convert
-// them to V8+
+// them to v8+
 func (p *Bundle) DeprecatedV7Conversions() {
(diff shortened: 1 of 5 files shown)


policy/deprecated_v7.go
```diff
@@ -390,8 +392,10 @@ func (s *DeprecatedV7_ScoringSpec) ApplyToV8(ref *explorer.Mquery) {
 
 func (d *DeprecatedV7_PolicySpec) ToV8() *PolicyGroup {
 	policies := make([]*PolicyRef, len(d.Policies))
-	i := 0
```
czunker, line 393: nit: I think, you should also remove the `i++` at the end of the three loops. Otherwise, it is confusing to read. At least it was to me.

arlimus: great catch, I missed that while converting

chris-rock (approved):
Thank you @arlimus

PR #376, opened 2023-02-23
🧹 update cnquery

chris-rock:


arlimus (approved):

chris-rock:
we do this change as part of #375 since it fixes the proto changes

PR #381, opened 2023-02-23
🐛 don't throw errors if queries in bundles activate

arlimus:
This is lost in translation, where it correctly errors when we try to modify a query that doesn't exist, but it incorrectly errors if the query exists in the bundle and users try to activate it. Typically we don't run into this case, but the edge-case are policies that were written in v7 and have an explicit action ACTIVATE set.


chris-rock (approved):
Thank you @arlimus

PR #382, opened 2023-02-23
⭐ add group types

arlimus:
This is one of the major new features in v8 that allows to be more intentional with how groups in policies work. The 3 major types are:

1. **chapters** - which group checks and queries together similar to how chapters work in e.g. a CIS or NIST policy
2. **imports** - which focus on importing other policies only; you can use chapters to use embedded or imported checks and queries, while imports help with pulling policies in without altering the structure
3. **overrides** - a way to add exceptions and other changes to existing policies and queries and checks

These types can be expanded in the future (e.g. we were talking about policy-level variants before, not just query-level), but it covers the most important types for this major release.


chris-rock (approved):
Thank you @arlimus

PR #387, opened 2023-02-25
🐛 fix copying errors in goreleaser for UBI containers

imilchev:
I copied some things that were not supposed to be in this repo and the release didn't work because of that


chris-rock (approved):
Thank you @imilchev

PR #405, opened 2023-03-03
🐛 Save filters associated with queries

czunker:
Persist filters which are attached to queries, e.g. in the Kubernetes Inventory pack:
```
- uid: mondoo-kubernetes-pods-inventory
    name: Mondoo Kubernetes Pods Inventory
    version: 1.0.0
    tags:
      mondoo.com/platform: kubernetes
      mondoo.com/category: best-practices
    filters:
      - asset.platform == "k8s-pod"
    queries:
      - uid: k8s-pod
        title: Gather Pod Information
        query: |
          k8s.pod { * }
```

This fixes this error message in cnquery, when running a query pack:
```
x query.hub> failed to set query pack retrieved from upstream error="cannot compute filters for a querypack, unless it was compiled first" querypack=
```

### DIFF
--- a/policy/hub.go
+++ b/policy/hub.go
@@ -287,9 +287,11 @@ func (s *LocalServices) ComputeBundle(ctx context.Context, mpolicyObj *Policy) (
 		Props:    map[string]*explorer.Property{},
 	}
 
-	// we need to re-compute the asset filters
-	mpolicyObj.Filters = &explorer.Filters{
-		Items: map[string]*explorer.Mquery{},
+	if mpolicyObj.Filters == nil {
+		// we need to re-compute the asset filters
+		mpolicyObj.Filters = &explorer.Filters{
+			Items: map[string]*explorer.Mquery{},
+		}
 	}
 	bundleMap.Policies[mpolicyObj.Mrn] = mpolicyObj
 
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -110,8 +110,14 @@ func WaitUntilDone(resolver PolicyResolver, entity string, scoringMrn string, ti
 
 // RefreshLocalAssetFilters looks through the local policy asset filters and rolls them up
 func (p *Policy) RefreshLocalAssetFilters() {
-	p.Filters = &explorer.Filters{
-		Items: map[string]*explorer.Mquery{},
+	if p.Filters == nil {
+		p.Filters = &explorer.Filters{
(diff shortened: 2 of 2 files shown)


policy/policy.go
```diff
@@ -110,8 +110,10 @@ func WaitUntilDone(resolver PolicyResolver, entity string, scoringMrn string, ti
 
 // RefreshLocalAssetFilters looks through the local policy asset filters and rolls them up
 func (p *Policy) RefreshLocalAssetFilters() {
-	p.Filters = &explorer.Filters{
-		Items: map[string]*explorer.Mquery{},
+	if p.Filters == nil {
+		p.Filters = &explorer.Filters{
+			Items: map[string]*explorer.Mquery{},
```
chris-rock, line 115: I think we should separately check if items is nil

czunker: Added the check

czunker:
Seems this is only half the fix. They are also not returned.

policy/hub.go
```diff
@@ -287,9 +287,11 @@ func (s *LocalServices) ComputeBundle(ctx context.Context, mpolicyObj *Policy) (
 		Props:    map[string]*explorer.Property{},
 	}
 
-	// we need to re-compute the asset filters
-	mpolicyObj.Filters = &explorer.Filters{
-		Items: map[string]*explorer.Mquery{},
+	if mpolicyObj.Filters == nil {
```
arlimus, line 290: I'm curious why we check if filters are set here? 

ComputeBundle's original job was to (amongst other thigns) re-compute the policy filters from its PolicyGroups, so technically it should always clear out this structure first.

policy/policy.go
```diff
@@ -110,8 +110,14 @@ func WaitUntilDone(resolver PolicyResolver, entity string, scoringMrn string, ti
 
 // RefreshLocalAssetFilters looks through the local policy asset filters and rolls them up
 func (p *Policy) RefreshLocalAssetFilters() {
-	p.Filters = &explorer.Filters{
-		Items: map[string]*explorer.Mquery{},
+	if p.Filters == nil {
```
arlimus, line 113: Same as above comment, the job of this function was to create p.Filters and clear it out.

arlimus (changes requested):
Synced with @czunker , we need to approach this differently. He highlighted a great problem we have and may want to address in the proto now...

PR #410, opened 2023-03-06
🧹🏗  overhaul resolver

arlimus:
This is by no means the last iteration, in fact I still see problems with the overrides that v8 allows us to do now. However, this is a step forward in cleaning up the code and restructuring how the resolver works.

Follow-up: there is still a problem with generating checksums for policy bundles, which shows up for higher-level policies and might not be in this repo alone... Left a comment in the code.

1. PolicyRef removed props and filters. However, they are only commented out, because we fully intend to add them back in. The resolver needs to generate local checksums for these refs that can be properly chained in reporting jobs - which is not the case right now.
2. PolicyRef doesn't need graph nor content checksums (at least from the resolver's perspective). There might be a use-case for this down the line, but I removed it for now.
3. In the resovler cache the term checksum is problematic, because it can refer to many different things (including llx checksums). To avoid this, we are now calling the affected field Msum short for mquery checksum. It's only used internally and not exposed to developers...
4. The root job is no longer tracked in reporting jobs by checksum, since no child is allowed to modify it.
5. The whole code that attempted to track policy Mrns in reportingJobsByChecksum has been removed. This section needs revisiting during policy overrides.
6. We start to acknowledge that the same MRN can appear in multiple reporting jobs in their children (e.g. through policies, checks, and queries).
7. Some of the more complex parts of adding check and query jobs have been separated out. Needs more work too, but we want don't want to overdo abstractions in the resolver. It needs to land a balance between keeping the code simple and moving larger chunks to abstractions.
8. The same query checksum can land multiple reporting jobs, because it can report into multiple policies. Once again, there is more work to be done here, since v8 starts to enable overrides of all fields - which pre-v8 did not support (and thus the existing resolver never had to recognize).


chris-rock (approved):
Wooot. Great work @arlimus

PR #414, opened 2023-03-06
🐛 make ImpactToScoringSpec more compatible with v7

jaym:
We need to be able to create the from the impact without an action specified as it is only available in the first resolve. Without this, a second scan produces errored scores even if only 1 of n things errored.


chris-rock (approved):
Makes a lot of sense @jaym

PR #418, opened 2023-03-07
🐛 fix v7 custom bundle loading

arlimus:
Props were so far only loaded from raw v7 bundles, which generally use UIDs instead of MRNs. However, it didn't work on processed v7 bundles (like validated and compiled bundles) which are using MRNs instead.

Additionally, we were dragging around the properties field in policies, which in v7 only served to provide props to queries, but didn't do anything else (e.g. wiring / overrides). Because we can now directly embed properties into queries, we can fully remove them from policies during the conversion process.

To make the analysis less brittle, we now look for the basename of MRN properties (instead of a specific resource in the MRN). This is in line with v7 behavior, which also puts the name at the end.


chris-rock (approved):
Thank you @arlimus

PR #424, opened 2023-03-09
🐛 Fix progress for unscored assets

czunker:
Before, the progress bars looked like this, when unscored assets where included:
```
 kube-system/coredns                                ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────   0% score: U
 mondoo-operator/mondoo-client-scan-api             ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────   0% score: U
 mondoo-operator/mondoo-operator-controller-manager ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────   0% score: U
 mondoo-operator/mondoo-client-k8s-scan             ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% score: A
 mondoo-operator/mondoo-client-node-minikube        ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% score: A

 5/5 scanned                                        ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━──────────────────────────────────────────────────────────────────────────────────────────────────────────────  40%

```

Now, it looks like this:
```
 kube-system/coredns                                ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────  n/a score: U
 mondoo-operator/mondoo-client-scan-api             ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────  n/a score: U
 mondoo-operator/mondoo-operator-controller-manager ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────  n/a score: U
 mondoo-operator/mondoo-client-k8s-scan             ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% score: A
 mondoo-operator/mondoo-client-node-minikube        ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% score: A

 2/5 scanned 3/5 n/a                                ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100%

```

### DIFF
--- a/cli/reporter/print_compact.go
+++ b/cli/reporter/print_compact.go
@@ -373,20 +373,22 @@ func (r *defaultReporter) printAssetQueries(resolved *policy.ResolvedPolicy, rep
 		r.out.Write([]byte(NewLineCharacter))
 	}
 
-	r.out.Write([]byte("Controls:" + NewLineCharacter))
-	for id, score := range report.Scores {
-		_, ok := resolved.CollectorJob.ReportingQueries[id]
-		if !ok {
-			continue
-		}
+	if len(report.Scores) > 0 {
+		r.out.Write([]byte("Controls:" + NewLineCharacter))
+		for id, score := range report.Scores {
+			_, ok := resolved.CollectorJob.ReportingQueries[id]
+			if !ok {
+				continue
+			}
 
-		query, ok := queries[id]
-		if !ok {
-			r.out.Write([]byte("Couldn't find any queries for incoming value for " + id))
-			continue
-		}
+			query, ok := queries[id]
+			if !ok {
+				r.out.Write([]byte("Couldn't find any queries for incoming value for " + id))
+				continue
+			}
 
-		r.printControl(score, query, asset, resolved, report, results)
+			r.printControl(score, query, asset, resolved, report, results)
+		}
 	}
 }
 
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -595,7 +595,11 @@ func (s *localAssetScanner) run() (*AssetReport, error) {
 		return ar, err
 	}
 	s.ProgressReporter.Score(report.Score.Rating().Letter())
-	s.ProgressReporter.Completed()
+	if report.Score.Rating().Letter() == "U" {
+		s.ProgressReporter.NotApplicable()
+	} else {
+		s.ProgressReporter.Completed()
+	}
 
 	log.Debug().Str("asset", s.job.Asset.Mrn).Msg("scan complete")
 	ar.Report = report


policy/scan/local_scanner.go
```diff
@@ -595,7 +595,11 @@ func (s *localAssetScanner) run() (*AssetReport, error) {
 		return ar, err
 	}
 	s.ProgressReporter.Score(report.Score.Rating().Letter())
-	s.ProgressReporter.Completed()
+	if report.Score.Rating().Letter() == "U" {
```
chris-rock, line 598: Shouldn't this work more like https://github.com/mondoohq/cnquery/blob/a3f56da5a8af4f7e2f7ec46b04e781e0a0089c37/explorer/scan/local_scanner.go#L270-L274 and https://github.com/mondoohq/cnquery/blob/a3f56da5a8af4f7e2f7ec46b04e781e0a0089c37/explorer/scan/local_scanner.go#L311-L317?

czunker: I think, this is different. It is not an error, the asset matched a policy, but didn't get scored.

I used this policy to provoke this:
```
  - asset_filter:
      query: platform.name == "k8s-deployment"
    scoring_queries:
    data_queries:
```

@preslavgerchev In which situation did you see the problem with the progress bar?

preslavgerchev: I tested this with the `Kubernetes Best Practices by Mondoo` policy, however it had most of its controls disabled or ignored, leading to the situation you described above (matching filter, no queries).

i can provide the exact policy (with ignored queries) if needed

chris-rock: I am not 100% sure I understand what we are solving here then. In case policies applied it was scored. Feels like the issue is with the policy itself.

chris-rock: Preslav showed me the issue. I agree that we need to fix the overall progress, but reading the score itself feels wrong. The progress bar should not care about the rating at all. We can move forward with this approach as mitigation but we need a better solution for that handling going forward.

cli/reporter/print_compact.go
```diff
@@ -373,20 +373,22 @@ func (r *defaultReporter) printAssetQueries(resolved *policy.ResolvedPolicy, rep
 		r.out.Write([]byte(NewLineCharacter))
 	}
 
-	r.out.Write([]byte("Controls:" + NewLineCharacter))
-	for id, score := range report.Scores {
-		_, ok := resolved.CollectorJob.ReportingQueries[id]
-		if !ok {
-			continue
-		}
+	if len(report.Scores) >= 0 {
```
preslavgerchev: was this meant to be `> 0` instead of `>= 0` ? else the check does not make a lot of sense

czunker: Oh, good catch. Thanks.

preslavgerchev (approved):

policy/scan/local_scanner.go
czunker: Follow-up is https://github.com/mondoohq/cnspec/pull/439

PR #432, opened 2023-03-10
🪚 yac-it: yaml and context structures

arlimus:
Add a lightweight library to create YAML structs and file context. This is very useful for cases where YAML metadata is missing while JSON metadata exists. It also provides file context for all parsed objects. We may externalize this library in the future, but for now it's here to solve a problem...

We will apply this shortly to overhaul `cnspec bundle format`, which has used a manually generated file structure.


chris-rock (approved):
Yeah!

PR #434, opened 2023-03-10
🧹  edited example bundle comments, changed one check

misterpantz:
Fixed typos
Made comments more readable

### DIFF
--- a/examples/example.mql.yaml
+++ b/examples/example.mql.yaml
@@ -7,9 +7,9 @@ policies:
   - uid: example1
     name: Example policy 1
     version: "1.0.0"
-    # If your policy has queries with severities specify the
+    # If your policy has checks with impacts, specify the
     # "highest impact" scoring system. This system uses the lowest
-    # score (highest impact failure) as the overall report score.
+    # score (the highest impact failure) as the overall report score.
     scoring_system: highest impact
     authors:
       - name: Mondoo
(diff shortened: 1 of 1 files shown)


examples/example.mql.yaml
```diff
@@ -2,30 +2,29 @@
...
   - uid: example1
     name: Example policy 1
     version: "1.0.0"
-    # If your policies has queries that have a "severity", you should set
-    # the scoring system to "highest impact" (like in this case). It makes
-    # sure that the lowest score (i.e. highest impact failure) is used as
-    # the overall score for the report.
+    # If your policy has queries with severities specify the
+    # "highest impact" scoring system. This system uses the lowest
+    # score (highest impact failure) as the overall report score.
```
tas50: We could really use a simple example here

misterpantz: Can you recommend one?

tas50: A smaller version of this scenario:

- You have a policy with 3 with failing queries:
  - 1 with impact of 10
  - 1 with impact of 30
  - 1 with impact of 90 (pretty critical)
- Your policy scores a 10 (100-90) and thus is an F

misterpantz: Thanks, that helps me. I changed the text based on this... but I didn't provide an example. I think that's a lot of info for a sample file. I'll make sure to describe it in the docs.

arlimus (approved):
Awesome improvements, thank you!! these examples are more :star2:

PR #441, opened 2023-03-14
🐛 avoid accidentally merging queries in compile phase

arlimus:
My previous pr https://github.com/mondoohq/cnspec/pull/425 introduced an error that merged base queries into their parents, which caused any updates to base queries not get registered as a change. This PR fixes the issue by keeping track of a merged query object during the compile step, which avoids writing unnecessary information into the policy/group check or query.

### DIFF
--- a/policy/bundle.go
+++ b/policy/bundle.go
@@ -549,8 +549,9 @@ func (c *bundleCache) error() error {
 }
 
 func (c *bundleCache) compileQueries(queries []*explorer.Mquery, policy *Policy) error {
+	mergedQueries := make([]*explorer.Mquery, len(queries))
 	for i := range queries {
-		c.precompileQuery(queries[i], policy)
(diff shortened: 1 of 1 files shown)


policy/bundle.go
```diff
@@ -596,7 +600,7 @@ func (c *bundleCache) precompileQuery(query *explorer.Mquery, policy *Policy) {
 	if policy == nil {
 		c.lookupQuery[query.Mrn] = query
 	} else if existing, ok := c.lookupQuery[query.Mrn]; ok {
-		query.AddBase(existing)
+		query = query.Merge(existing)
```
arlimus, line 609: AddBase modifies the query in the policy group - we don't want that. Merge doesn't modify it, but we need the merged query a bit later for the compile step, so we don't want to drop it.

jaym (approved):

PR #443, opened 2023-03-14
🐛 return partial bundle maps on compile

arlimus:
The previous behavior (before e4c0b5d ) was to return the partial bundle, even if we had errors. So it returns both the error and the bundle. Restore tothis behavior. We should revisit it down the line...


arlimus:
Closing, it may have adverse side-effects to do this so early.

Issue #448, opened 2023-03-15
cnspec bundle format should update `scoring_system`


tas50:
**Is your feature request related to a problem? Please describe.**
Right now you can define `scoring_system` in a policy with a number or the friendly value. When we run `cnspec bundle format` on a policy we should write out the friendly format, which is much more newcomer friendly.

**Describe the solution you'd like**
Give

```yaml
policies:
  - uid: example1
    name: Example policy 1
    version: "1.0.0"
    scoring_system: 2
```

Running `cnspec bundle format example.yml` would result in:

```yaml
policies:
  - uid: example1
    name: Example policy 1
    version: "1.0.0"
    scoring_system: highest impact
```

PR #450, opened 2023-03-16
🧹 clean up old ScoringSystem

arlimus:
... and instead start to use cnquery ScoringSystem. They were both identical (with one enum mising), so no need to carry both around with us. See https://github.com/mondoohq/cnquery/pull/1076

Note: requires the above PR to merge + go.mod to be updated for that cnquery

Fixes https://github.com/mondoohq/cnspec/issues/448 (because it uses the new cnquery ScoringSystem which has a `MarshalYAML` that returns the string version of the scoring system)


chris-rock (approved):
Thank you @arlimus

PR #455, opened 2023-03-19
🧹 re-use cnquery AssetMatchError + add resolver tests

arlimus:
Only after https://github.com/mondoohq/cnquery/pull/1081 + update to go deps.


chris-rock (approved):
Nice cleanup

PR #458, opened 2023-03-20
🐛 bundle fmt

chris-rock:
- handle case where v7 structs where loaded in v8 parse
- order fields for yaml formatting


arlimus (approved):
nice ordering

PR #466, opened 2023-03-20
🧹 update the default policy bundle

chris-rock:


atomic111 (approved):
@chris-rock thanks for the fix

PR #469, opened 2023-03-21
🐛 fixes an issue where a list of typed docs where not parsed properly

chris-rock:


arlimus (approved):
thank you the fix, weird that this didn't work...

PR #471, opened 2023-03-23
Add ScoreDistrubition.Error

preslavgerchev:
We already have a score rating for an `Error` so I figured it's easy to put the `error` assets in the `ScoreDistribution` too

Not sure if there was a good reason in the past to not add that there, so let me know if that is the case


policy/report.go
```diff
@@ -79,6 +79,8 @@ func (sd *ScoreDistribution) Add(score *Score) {
 		sd.D++
 	case ScoreRating_failed:
 		sd.F++
+	case ScoreRating_error:
```
chris-rock, line 82: to cover all cases we also would need to catch the case where the score is unrated

chris-rock (approved):
Great coverage improvement @preslavgerchev

PR #472, opened 2023-03-24
⭐ Add Policy Preview

czunker:


policy/cnspec_policy.proto
```diff
@@ -106,6 +106,9 @@ message Policy {
   cnquery.explorer.Filters computed_filters = 43;
 
   QueryCounts query_counts = 42;
+
+  // Holds the information whether the policy is assinged in preview or active state
+  PolicyAssignmentType assignmentType = 47;
```
arlimus, line 111: This field is unnecessary and shouldn't be set in the policy at all. The concept of preview is a Mondoo-only (upstream) concept, that is done via its assignment in the managing policy (not in scope of cnspec).

arlimus (changes requested):
Policy preview is not an OSS concept and thus shouldn't be introduced here. Anyone that wants to run policies effectively unscored can set their action to `IGNORE`. I recommend closing this here.

PR #474, opened 2023-03-26
Add functions to remove add and remove ratings from score dist

preslavgerchev:
There's already a function to add, based on the `Score` struct. This PR adds a `Remove` function and adds functions to do the same, this time based solely on `ScoreRating`.

### DIFF
--- a/policy/report.go
+++ b/policy/report.go
@@ -66,9 +66,16 @@ func (s *Stats) Add(score *Score) {
 }
 
 func (sd *ScoreDistribution) Add(score *Score) {
-	sd.Total++
+	sd.AddRating(score.Rating())
+}
 
-	switch score.Rating() {
+func (sd *ScoreDistribution) Remove(score *Score) {
+	sd.RemoveRating(score.Rating())
+}
+
+func (sd *ScoreDistribution) AddRating(scoreRating ScoreRating) {
+	sd.Total++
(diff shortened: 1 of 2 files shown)


policy/report.go
```diff
@@ -66,9 +66,16 @@ func (s *Stats) Add(score *Score) {
 }
 
 func (sd *ScoreDistribution) Add(score *Score) {
-	sd.Total++
+	sd.AddRating(score.Rating())
```
chris-rock, line 70: do we need to catch the case where score is nil?

preslavgerchev: That makes sense, however looking at the `Rating()` function on the score it already handles a nil score and returns `ScoreRating_Unrated`. 

Seeing as how that function is used in a few places I'd prefer to not introduce that check right now since I do not know how scores are being passed in wherever that function is being used, it could lead to a change in behaviour.

chris-rock: Makes sense, please add a comment to the function. Essentially it allows nil scores and counts them accordingly. With is all good. So if the caller cares about that it need to handle nil values themselves.

chris-rock (approved):

policy/report.go
preslavgerchev: also updated the tests so that if we change this in the future, there will be a failing test to indicate that

PR #477, opened 2023-03-27
🐛 fix bundle fmt command to properly convert v7 to v8

imilchev:
`scoring_system` completely disappeared when running `cnspec bundle fmt`. This PR fixes the problem. I confirmed this by taking a v7 policy and running `cnspec bundle fmt` for it with latest `main` and with this branch. The only difference between the output of the 2 is the added `scoring_system`.


internal/bundle/fmt.go
```diff
@@ -69,7 +70,9 @@ func DeprecatedV7_ToV8(data []byte) ([]byte, error) {
 	// this step will unfortunately not produce well-formatted YAML at all
 	// because the proto structures don't have the yaml tags (only the
 	// yac-it structures do) ...
-	interim, err := Format(v8)
+	// for the same reason we convert between the 2 types using JSON since both
```
chris-rock, line 73: can you add a test case  for it so that we see this works properly

imilchev: I extended the existing test to verify the presence of `scoring_system`

internal/bundle/fmt_test.go
```diff
@@ -57,6 +58,7 @@ queries:
       - filters: platform.family.contains(_ == 'unix')
         checks:
           - uid: query1
+    scoring_system: 2
```
chris-rock, line 61: can we also test that the value exists?

chris-rock (approved):
Thank you @imilchev

PR #480, opened 2023-03-30
🐛 Handle the scoring of ignore policies

czunker:
This fixes the handling of ignored policies.
The policy controls are scored, but the score isn't added to the asset


policy/resolver.go
```diff
@@ -404,14 +404,16 @@ func (s *LocalServices) tryResolve(ctx context.Context, policyMrn string, assetF
 		return nil, err
 	}
 
-	var rp *ResolvedPolicy
-	rp, err = s.DataLake.CachedResolvedPolicy(ctx, policyMrn, allFiltersChecksum, V2Code)
+	// var rp *ResolvedPolicy
+	_, err = s.DataLake.CachedResolvedPolicy(ctx, policyMrn, allFiltersChecksum, V2Code)
```
czunker: The caching does currnetly not work. This is WIP and needs a fix.

policy/policy.go
```diff
@@ -433,6 +433,7 @@ func (p *Policy) updateAllChecksums(ctx context.Context,
 			}
 
 			executionChecksum = executionChecksum.Add(ref.Mrn)
+			executionChecksum = executionChecksum.Add(ref.Action.String())
```
arlimus: Go for `AddUint` here, faster and smaller to do

czunker: Fixed.

policy/resolver.go
czunker: Fixed

arlimus (approved):
lgtm, thank you!

arlimus (approved):
Thanks for the update!

policy/policy.go
```diff
@@ -433,6 +433,7 @@ func (p *Policy) updateAllChecksums(ctx context.Context,
 			}
 
 			executionChecksum = executionChecksum.Add(ref.Mrn)
+			executionChecksum = executionChecksum.AddUint(uint64(ref.Action))
```
joelddiaz, line 436: Does the line right above this have any purpose if the result in `executionChecksum` is immediately overwritten?

czunker: It's not overwritten, it get's additional data added:
https://github.com/mondoohq/cnquery/blob/323c116a72e37534ebeafc2ccfe9c6bc71f79e7b/checksums/checksum.go#L22

Another example for this in this file:
https://github.com/mondoohq/cnspec/blob/e730ccf05254e433192557aa73eb98effc123688/policy/policy.go#L366-L376

joelddiaz: Wow...I completely misread these two lines of. Just pretend I never said anything :)

Issue #482, opened 2023-03-31
Scanning with a query pack only shows an asset as errored


czunker:
**Describe the bug**
When I scan an asset with only a query pack enabled, it gets shown as errored.

**To Reproduce**

1. Only enable a matching query pack in the UI
2. Scan asset
3. Note the error

**Expected behavior**
It should show as unscored.

**Screenshots or CLI Output**
```
cnspec scan local
→ discover related assets for 1 asset(s)
→ resolved assets resolved-assets=1

Asset: X1
---------

Data queries:
os.hostname: "X1"
platform.name: "ubuntu"
users.where.list: [

...
... 1044 more lines ...

Controls:


Scanned 1 assets

Ubuntu 22.04.2 LTS
    X X1

```

![image](https://user-images.githubusercontent.com/827818/229094233-71fba741-4406-4aba-a402-17a1712365aa.png)

PR #487, opened 2023-04-04
🐛 Fix the scoring when scaning with inventory packs

czunker:
The score for inventory packs is now unscored and no longer error.

Fixes: #482

### DIFF
--- a/policy/executor/internal/nodes.go
+++ b/policy/executor/internal/nodes.go
@@ -530,27 +530,22 @@ func (nodeData *ReportingJobNodeData) score() (*policy.Score, error) {
 		return s, nil
 	}
 
-	s := &policy.Score{
-		Type: policy.ScoreType_Unscored,
+	calculator, err := policy.NewScoreCalculator(nodeData.scoringSystem)
+	if err != nil {
+		return nil, err
 	}
-	if nodeData.scoringSystem != explorer.ScoringSystem_DATA_ONLY {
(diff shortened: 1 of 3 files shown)


czunker:
Now, it looks like this:
```
... 215 more lines ...



Scanned 1 assets

Ubuntu 22.04.2 LTS
    U X1

For detailed output, run this scan with "-o full".
```

policy/executor/internal/nodes.go
```diff
@@ -530,22 +530,27 @@ func (nodeData *ReportingJobNodeData) score() (*policy.Score, error) {
 		return s, nil
 	}
 
-	calculator, err := policy.NewScoreCalculator(nodeData.scoringSystem)
-	if err != nil {
-		return nil, err
+	s := &policy.Score{
+		Type: policy.ScoreType_Unscored,
```
preslavgerchev: Do we need to fill in the other props, such as `datacompletion` and `datatotal` here?

Edit: looks like those values seem to be correct in the db after the scan

preslavgerchev:
I just tested this out too, works great!

policy/executor/internal/nodes.go
czunker: Good catch. This lead to an even simpler solution. We can re-use the avg calculator. That also takes care of the rest.

preslavgerchev (approved):

PR #493, opened 2023-04-05
🧹 Add test for policy with ignore action

czunker:
Add unit test to check policies which gets modified with action "IGNORE".


czunker:
Follow-up to https://github.com/mondoohq/cnspec/pull/480

jaym (approved):

PR #495, opened 2023-04-05
🐛 Fix the checks and queries checksumming for policies

jaym:
There is a path that throws away data instead of merging queries. It's not clear why it was like that, however when changing a check from ignored to disabled wasn't invalidating the policy


chris-rock (approved):
Thank you @jaym

PR #497, opened 2023-04-06
📄 add documentation on client-specific proxy

misterpantz:
- add documentation on client-specific proxy setting


chris-rock:
We do not use this flag, it is `api-proxy`

PR #501, opened 2023-04-07
⭐️ shared reports

joelddiaz:


apps/cnspec/cmd/scan.go
```diff
@@ -374,8 +376,33 @@ This example connects to Microsoft 365 using the PKCS #12 formatted certificate:
 		}
 
 		logger.DebugDumpJSON("report", report)
+		// TODO: ONLY FOR TESTING PURPOSES REMOVE ME
+		if fileName := os.Getenv("DEBUG_SAVE_REPORT_COLLECTION_FILE"); fileName != "" {
```
chris-rock: this should automattically happen when `log.Debug().Enabled()` is enabled. See logging helper function we have here https://github.com/mondoohq/cnquery/blob/main/logger/debug.go

apps/cnspec/cmd/scan.go
```diff
@@ -388,6 +415,27 @@ This example connects to Microsoft 365 using the PKCS #12 formatted certificate:
 	},
 })
 
+func askToUploadReport() bool {
+	reader := bufio.NewReader(os.Stdin)
```
chris-rock: This should move to a cli component. If it gets more complex also consider using https://github.com/charmbracelet/bubbletea/tree/master/examples#result

apps/cnspec/cmd/scan.go
```diff
@@ -374,8 +376,33 @@ This example connects to Microsoft 365 using the PKCS #12 formatted certificate:
...
+		// ^^
 		printReports(report, conf, cmd)
 
+		// if no config set, and wasn't explicitly incognito, offer to upload
+		// report for web UI viewing
+		opts, optsErr := cnspec_config.ReadConfig()
+		if optsErr != nil {
+			log.Fatal().Err(optsErr).Msg("could not load configuration")
+		}
+		if opts.GetServiceCredential() == nil && !viper.GetBool("incognito") {
```
chris-rock: we also need to check for tty since it would not be great to ask in CI/CD:

```
isTTY := isatty.IsTerminal(os.Stdout.Fd())
```

policy/cnspec_policy.proto
```diff
@@ -662,3 +662,13 @@ message PurgeAssetsConfirmation {
   map<string, string> errors = 2;
 }
 
+// **********       Public Report        **************
+
+service PublicReport {
```
chris-rock: service name is just `Reporting`

policy/cnspec_policy.proto
```diff
@@ -662,3 +662,13 @@ message PurgeAssetsConfirmation {
...
+  rpc StoreReport(cnspec.policy.v1.ReportCollection) returns(PublicReportID) {}
+} 
+
+message PublicReportID {
```
chris-rock: this should be just `ReportID`

policy/cnspec_policy.proto
```diff
@@ -662,3 +662,13 @@ message PurgeAssetsConfirmation {
   map<string, string> errors = 2;
 }
 
+// **********       Public Report        **************
```
chris-rock: Shared reporting belong into upstream/upstream.proto similar to https://github.com/mondoohq/cnquery/tree/main/upstream

apps/cnspec/cmd/scan.go
joelddiaz: Yes, I just wanted a quick way to get the binary protobuf for testing w/o having to convert the json. I'll remove this entirely

apps/cnspec/cmd/scan.go
joelddiaz: Yeah, that would be unfortunate for CI/CD. Thanks

cli/components/report_sharing.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/rs/zerolog/log"
+	"go.mondoo.com/cnspec/policy"
+	"go.mondoo.com/cnspec/upstream"
+	"go.mondoo.com/ranger-rpc"
+)
+
+const sharedReportUrl = "https://report.api.mondoo.com"
+
+func UploadSharedReport(report *policy.ReportCollection, reportUrl string, proxy *url.URL) error {
```
chris-rock: Can this move to upstream package?

cli/components/report_sharing.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		log.Error().Err(err).Msg("error uploading shared report")
+		return err
+	}
+
+	fmt.Printf("View your report at https://TBD.COM/space/overview?spaceId=%s\n", reportId.Id)
+	return nil
+}
+
+func AskToUploadReport() bool {
```
chris-rock: can we reduce the cli component to get question as input and it returns a true / false answer?

policy/cnspec_policy.proto
```diff
@@ -287,6 +288,7 @@ message Asset {
   string name = 18;
   string url = 19;
   string platformName = 20;
+  cnquery.motor.platform.v1.Platform platform = 21;
```
chris-rock, line 291: If we add platform here, we should not use platform name anymore

policy/cnspec_policy.proto
```diff
@@ -286,7 +287,7 @@ message Asset {
   string mrn = 1;
   string name = 18;
   string url = 19;
-  string platformName = 20;
```
chris-rock, line 289: the field needs to be reserved to make sure it is non-breaking

apps/cnspec/cmd/scan.go
```diff
@@ -383,6 +395,21 @@ This example connects to Microsoft 365 using the PKCS #12 formatted certificate:
 		logger.DebugDumpJSON("report", report)
 		printReports(report, conf, cmd)
 
+		// if we are in an interactive terminal, and no config, and not explicty incognito, and user responds "yes" to upload offer
+		opts, optsErr := cnspec_config.ReadConfig()
```
chris-rock: This is not required. We can just check for conf.IsIncognito

upstream/upstream.go
```diff
@@ -0,0 +1,37 @@
...
+		return err
+	}
+
+	reportId, err := sharedReportClient.StoreReport(context.Background(), report)
+	if err != nil {
+		log.Error().Err(err).Msg("error uploading shared report")
+		return err
+	}
+
+	fmt.Printf("View your report at https://TBD.COM/space/overview?spaceId=%s\n", reportId.Id)
```
chris-rock: Functions not located in cli packages should not print to cli. This needs to move to scan command and I would expect that `StoreReport` function also returns a report url

chris-rock (approved):
Thank you Joel. I added a few comments and pushed changes to address those.

cli/components/question.go
```diff
@@ -0,0 +1,31 @@
...
+	"strings"
+
+	"github.com/rs/zerolog/log"
+)
+
+func AskAYesNoQuestion(msg string) bool {
+	reader := bufio.NewReader(os.Stdin)
+
+	for {
+		fmt.Printf("%s [(y)es/(n)o]: ", msg)
```
chris-rock, line 16: The output can be improved. Debian apt uses a simple approach:

```
Do you want to continue? [Y/n]
```

In addition the current implementation does not handle cases where users presses a non allowed value. Similar to debian, it should make clear what the default value is. I would recommend we just use the default pattern established by debian

policy/cnspec_policy.proto
chris-rock: I added the reserved field

apps/cnspec/cmd/scan.go
chris-rock: I changed the implementation to be more simple

cli/components/question.go
chris-rock: We improve that in a follow-up PR.

PR #502, opened 2023-04-07
🐛 return err if it's not nil in SetBundleMap

imilchev:
Seems like a bug. It causes weird behaviour when errors happen since they are ignored


chris-rock (approved):
Thank you @imilchev

PR #503, opened 2023-04-07
⭐️ Make policy resolver take the end date of groups into account

jaym:
The policy groups had an end date added, but we never wired the resolver up to look at that field when making the resolved policy


chris-rock (approved):
Great addition @jaym

PR #506, opened 2023-04-10
🐛 Fix checksumming of policies with variants

jaym:
The execution checksum was only taking into account the code id of the query that specified it had variants. This wasn't changing if one of the variants since that query itself has no code.

### DIFF
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -472,9 +472,12 @@ func (p *Policy) updateAllChecksums(ctx context.Context,
 			}
 
 			contentChecksum = contentChecksum.Add(check.Checksum)
-			executionChecksum = executionChecksum.
-				Add(check.CodeId).
-				AddUint(check.Impact.Checksum())
+
+			var err error
+			executionChecksum, err = variantsExecutionChecksum(check, executionChecksum, true, getQuery)
+			if err != nil {
+				return err
+			}
 		}
 
 		// DATA (must be sorted)
@@ -507,7 +510,12 @@ func (p *Policy) updateAllChecksums(ctx context.Context,
 			}
 
 			contentChecksum = contentChecksum.Add(query.Checksum)
-			executionChecksum = executionChecksum.Add(query.CodeId)
+
+			var err error
+			executionChecksum, err = variantsExecutionChecksum(query, executionChecksum, false, getQuery)
+			if err != nil {
+				return err
+			}
 		}
 
 		// FILTERs (also sorted)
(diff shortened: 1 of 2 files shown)


policy/policy.go
```diff
@@ -681,3 +689,21 @@ func (s *GroupType) UnmarshalJSON(data []byte) error {
 
 	return errors.New("failed to unmarshal group type: " + str)
 }
+
+func variantsExecutionChecksum(q *explorer.Mquery, c checksums.Fast, includeImpact bool, getQuery func(ctx context.Context, mrn string) (*explorer.Mquery, error)) (checksums.Fast, error) {
+	// TODO: this function needs to be safe against cycles
```
imilchev: Do you mean that the same `CodeId` shouldn't be added to the checksum multiple times? If yes, then I guess this change should be sufficient:
```go
func variantsExecutionChecksum(q *explorer.Mquery, c checksums.Fast, includeImpact bool, getQuery func(ctx context.Context, mrn string) (*explorer.Mquery, error), added map[string]struct{}) (checksums.Fast, error) {
	if _, ok := added[q.CodeId]; ok {
		return c, nil
	}

	c = c.Add(q.CodeId)
	visited[q.CodeId] = struct{}{}
	if includeImpact {
		c = c.AddUint(q.Impact.Checksum())
	}

	for _, ref := range q.Variants {
		if v, err := getQuery(context.Background(), ref.Mrn); err == nil {
			return variantsExecutionChecksum(v, c, includeImpact, getQuery, added)
		} else {
			return 0, errors.New("cannot find dependent composed query '" + ref.Mrn + "'")
		}
	}
	return c, nil
}
```

jaym: its not really about being added so much as we dont want cycles because then the function wouldn't terminate

jaym: also, its possible that there is not a code id because there is no code

policy/policy.go
```diff
@@ -681,3 +689,21 @@ func (s *GroupType) UnmarshalJSON(data []byte) error {
...
+	c = c.
+		Add(q.CodeId)
+	if includeImpact {
+		c = c.AddUint(q.Impact.Checksum())
+	}
+
+	for _, ref := range q.Variants {
+		if v, err := getQuery(context.Background(), ref.Mrn); err == nil {
+			return variantsExecutionChecksum(v, c, includeImpact, getQuery)
```
jaym: this should not return

policy/policy.go
jaym: There are other things before this code runs which assume there are no cycles. So this problem probably needs to be fixed far before we get here

chris-rock (approved):
Thank you @jaym Looks good to me

PR #508, opened 2023-04-11
🐛 Check for cycles in variants

jaym:
Without this, we run into stack overflows if we are given a policy with queries that have cycles

After #506


chris-rock (approved):
Amazing catch @jaym

PR #510, opened 2023-04-12
🐛 Return useful URL for assets scanned in CICD

### LINKED ISSUE #484: The link in multi-asset CI scans goes to the wrong location
**Describe the bug**
This is going to require some conversation about what the possible right thing could be, but let's start with the current problem:

If you scan a Kubernetes manifest it is automatically a 2+ asset scan. The manifest file is an asset, and each of the resources within that file are also assets. When we scan with cnspec in a non-CI scenario the link given at the end of the scan is the space, which is less than ideal, but it works. When we scan in a CI environment like GH Actions we're giving users that same link. The problem is CI assets never show up in the space so this link takes them to a place where they will never be able to see the asset.

Where we could potentially link them instead:

- The manifest file on the explorer tab which would give them quick access to each repo
- The CI project page this scan would be a part of.

czunker:
This is the frontend fix.

Fixes #484


czunker:
The backend part is https://gitlab.com/mondoolabs/mondoo/-/merge_requests/4001/

PR #511, opened 2023-04-12
Switch the asset type in ReportCollection to asset.Asset. 

preslavgerchev:
Using the cnquery `motor.Asset` lets us pass around full data, that was gathered on those assets during discovery


policy/cnspec_policy.proto
```diff
@@ -415,10 +415,10 @@ message Reports {
 }
 
 message ReportCollection {
-  map<string, Asset> assets = 1;
```
joelddiaz, line 418: Is there really much use for the policy Asset type/message any more? I see it used in various places, but (from a quick glance) they seem like they could also just use the motor.Asset type?
Anyway, not really specific to the purpose of the PR. Just asking

preslavgerchev: 🤷  probably not, seeing as its a trimmed wrapper around motor.Asset
Would have to look everywhere to see if there's any special usage

joelddiaz (approved):

PR #512, opened 2023-04-12
🐛 Use MRNs for variant cycle detection

jaym:
There are many paths in which the UID is zero'd out if a MRN is calculated. Instead, the detection is moved to a point at which MRNs have been calculated, but before any compilation of queries happens

### DIFF
--- a/policy/bundle.go
+++ b/policy/bundle.go
@@ -513,51 +513,19 @@ func (p *Bundle) ExtractDocs(out io.Writer, noIDs bool, noCode bool) {
 	}
 }
 
-func (p *Bundle) ensureNoCyclesInVariants() error {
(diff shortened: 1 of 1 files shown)


policy/bundle.go
```diff
@@ -513,51 +513,19 @@ func (p *Bundle) ExtractDocs(out io.Writer, noIDs bool, noCode bool) {
...
-			if base != nil {
-				return base
-			}
-			return queryMap[mrn]
+	queriesMap := map[string]*explorer.Mquery{}
+	for _, q := range queries {
+		if q.Mrn == "" {
+			// This should never happen. This function is called after all
+			// queries have their MRNs set.
+			panic("BUG: expected query MRN to be set for variant cycle detection")
```
chris-rock, line 523: I like the explicit handling

chris-rock (approved):
Works like a charm @jaym

PR #516, opened 2023-04-13
pass in default ranger client if using incognito.

preslavgerchev:
Fixes nil panics if trying to fetch a vuln report when using incognito.
The vuln report requires upstream communication and it does that via the http client in the UpstreamConfig


vjeffrey (approved):
tested incognito and non-incognito scans on here, everything is behaving as expected

PR #517, opened 2023-04-14
🐛 Initialize endpoint and client when using incognito.

preslavgerchev:
If there's a present config, use it even with the incognito flag. That lets cnspec communicate upstream to analyze assets (if needed).
The missing endpoint and space mrn checks are now only done if cnspec is not running incognito since then it is assumed that those will not be needed, thus checking for them does not make sense.
The following will work:

`cnspec scan container fedora:31 --incognito`
`cnspec scan container fedora:31 --incognito --config bad-config`


chris-rock:
That makes sense. It is still incognito since it will not send results upstream.

chris-rock (approved):
Thank you @preslavgerchev

PR #518, opened 2023-04-17
Don't report to query mrn when it has variants. Use the code id

jaym:
Without this, we end up with some queries reporting by mrn and other reporting by code id.

Requires https://github.com/mondoohq/cnquery/pull/1154

### DIFF
--- a/policy/resolver.go
+++ b/policy/resolver.go
@@ -909,6 +909,9 @@ func (cache *policyResolverCache) addCheckJob(ctx context.Context, check *explor
 	// ^^
 
 	if len(check.Variants) != 0 {
+		// Its easier to deal with things if we dont mix query code ids with
+		// query mrns.
+		queryJob.QrId = check.CodeId
 		err := cache.addCheckJobVariants(ctx, check, queryJob)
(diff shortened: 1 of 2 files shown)


policy/resolver.go
```diff
@@ -909,6 +909,9 @@ func (cache *policyResolverCache) addCheckJob(ctx context.Context, check *explor
 	// ^^
 
 	if len(check.Variants) != 0 {
+		// Its easier to deal with things if we dont mix query code ids with
+		// query mrns.
+		queryJob.QrId = check.CodeId
```
jaym, line 914: I cant figure out the correct place to put this. There is other code that changes qrid from mrn to code id for checks, but I dont think it runs for these queries

arlimus (approved):
I'll approve this for now, but I'm not clear on what the problem is this solves. Variant queries do not contain code themselves, so having them assigned code IDs is very confusing.

jaym:
I've reverted the change to give variants code ids at https://github.com/mondoohq/cnquery/pull/1163 so this cannot work

PR #534, opened 2023-04-26
🧹Add tests for creating upstream config.

preslavgerchev:
Adds tests that verify an upstream config can be created when using `WithUpstream` and. when using job credentials, enabled by the `AllowJobCredentials`


czunker (approved):
Nice, more tests. Thanks, @preslavgerchev

PR #540, opened 2023-04-28
⭐️ Score errors as failures

jaym:
This changes the way checks get scored when they're errored. Instead of ignoring them as part of the policy score, they will be scored no different than if they were failing.

This is behind a feature flag "ErrorsAsFailures".

Example policy `errors.yaml`:
```
policies:
  - uid: test-pol
    name: Errors as Failures Test
    specs:
      - asset_filter:
          query: platform.family.contains(_ == 'unix')
        scoring_queries:
          my-test1:
          my-test2:
queries:
  - uid: my-test1
    title: erroring check with sev 50
    severity: 50
    query: |
      file("/doesnotexit").content != ""
  - uid: my-test2
    title: /etc exists
    severity: 100
    query: |
      file("/etc").exists == true
```

```
❯ MONDOO_FEATURES="ErrorsAsFailures" go run apps/cnspec/cnspec.go scan local -f errors.yaml
```

```
Asset: mars
-----------

Checks:
! Error: erroring check with sev 50
✓ Pass:  A 100  /etc exists

Scanned 1 assets

Arch Linux
    B mars
```

### DIFF
--- a/policy/executor/graph.go
+++ b/policy/executor/graph.go
@@ -30,6 +30,10 @@ func ExecuteResolvedPolicy(schema *resources.Schema, runtime *resources.Runtime,
 		builder.WithProgressReporter(progressReporter)
 	}
 
+	if features.IsActive(cnquery.ErrorsAsFailures) {
+		builder.WithFeatureFlagFailErrors()
+	}
+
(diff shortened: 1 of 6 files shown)


vjeffrey (approved):
works as advertised! :sparkles:

Issue #539, opened 2023-04-27
Add "Always / neVer" options and a flag to the cnspec report upload options


charlesjohnson:
**Is your feature request related to a problem? Please describe.**
At the end of cnspec runs, cnspec prompts the user:
```
Do you want to upload this report to Mondoo's web-based reporting service to view the results in a browser? [Y/n]
```

But if I want to wrap cnspec in a script, I can't answer a prompt every time. I need to be able to specify the behavior I want.

**Describe the solution you'd like**
1. Add an option to the config that defaults to "always attempt to upload report" or "never attempt to upload a report."
2. If no config option is specified, update the displayed prompt options from `[Y/n]` to 
```
[Yes* / No / Always / neveR ] 
```
(Or similar)

3. Add cli flags for `--upload-report` and `--skip-report-upload`, which will override the config option.
4. If there is no TTY present, the option flag should let me upload anyway (for instance, from CI)

chris-rock:
I recommend we update this to be:

```
Do you want to view or share this report to via Mondoo's reporting service to view the results in a browser? [Y/n]
```

and then we also add a flag that allows users to force the setting via:

```
cnspec scan --share-report
```

chris-rock:
If the user sets the flag, the report is uploaded even in non TTY environments:

```
cnspec scan --share-report
```

If that is not set then the user is asked:

```
Do you want to view or share this report results via Mondoo's reporting service in a browser? [Y/n]
```

To disable the the sharing and not being asked, users either

- `--share-report=false` or set the 
- set the config with the following settings

```
share_report: false # only applicable if used unauthenticated 
```

When a user `cnspec scan --share-report` and has disabled the reporting in the config, the cli flag has precedence and therefore the report will be shared.

PR #543, opened 2023-05-01
⭐️ configurable report sharing

chris-rock:
Users can now simply share a report via:

```
cnspec scan --share-report
```

If that is not set then the user is asked:

```
Do you want to view or share this report results via Mondoo's reporting service in a browser? [Y/n]
```

To disable the the sharing and not being asked, users either

- use the `--share-report=false` flag
- set the config with the following settings

```
share_report: false # only applicable if used unauthenticated 
```

When a user runs `cnspec scan --share-report` and has disabled the reporting in the config, the cli flag has precedence and therefore the report will be shared.

### DIFF
--- a/apps/cnspec/cmd/config/config.go
+++ b/apps/cnspec/cmd/config/config.go
@@ -23,5 +23,8 @@ type CliConfig struct {
 
 	// Asset Category
 	Category               string `json:"category,omitempty" mapstructure:"category"`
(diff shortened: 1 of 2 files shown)


chris-rock:
fixes #539

vjeffrey (approved):
works as advertised!

charlesjohnson:
Resolves #539

PR #546, opened 2023-05-01
🧹 improve linter for variants

### LINKED ISSUE #526: cnspec bundle publish fails when using variants
**Describe the bug**
`cnspec bundle publish` fails to upload a policy if it leverages variants because it believes checks are missing an `asset_filter`.

```
$ cnspec bundle publish policies/okta-security.mql.yaml
→ loaded configuration from /Users/path/.config/mondoo/test.yaml using source $MONDOO_CONFIG_PATH
→ load policy bundle file=policies/okta-security.mql.yaml
  RULE ID                      LEVEL  FILE                    LINE  MESSAGE                         
  policy-missing-asset-filter  error  okta-security.mql.yaml  65    Policy mondoo-okta-security-v2  
                                                                    doesn't define an asset         
                                                                    filter.                         
FTL invalid policy bundle
```

chris-rock:
fixes #526

With the new v8 and variants we support queries to have their own asset filter. Therefore policy groups do not necessarily need an asset filter anymore. We still need to see how this shapes out long term but it definitely makes sense not to show the warning when variants are uses since they need to have an asset filter.

### DIFF
--- a/internal/bundle/lint.go
+++ b/internal/bundle/lint.go
@@ -340,7 +340,8 @@ func lintFile(file string) (*Results, error) {
 		for j := range policy.Groups {
 			group := policy.Groups[j]
 
-			if group.Filters == nil || len(group.Filters.Items) == 0 {
+			// issue warning if no filters are assigned, but do not show the warning if the policy has variants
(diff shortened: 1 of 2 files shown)


imilchev (approved):
LGTM

PR #550, opened 2023-05-04
🧹 allow specifying custom http timeout for serve-api cmd

imilchev:
The default HTTP timeout is 30s but sometimes that's not enough. For example, if the mondoo-operator needs to clean thousands of assets, it might take more than 30s. The operator itself sets the timeout for its client to 50minutes or so but the scan API terminates the request on 30s.

Added a flag to allow extending this timeout when needed


chris-rock:
Since this is only a temporary fix. The real solution is that the server needs to respond faster. Happy to chat about how we achieve that. Therefore please mark the flag as hidden since it is nothing we should advertise as properly supported.

chris-rock (approved):
Thank you @imilchev

PR #560, opened 2023-05-09
✨ variants inherit missing fields from parents

arlimus:
Allow variant queries to pull common fields from their parents, if they have not been specified. For example: title, descriptions, and properties.


chris-rock (approved):
Great addition @arlimus

PR #561, opened 2023-05-09
⭐️ propagate props from base to variants so we do not have to define multiple times

chris-rock:
Followup in #560


scottford-io (approved):
Worked like a charm @chris-rock ⭐ 🌟 !!!!

arlimus (approved):
yeah great catch, sorry @chris-rock !

PR #574, opened 2023-05-17
🐛 fix the recording

chris-rock:
depends on https://github.com/mondoohq/cnquery/pull/1217

This issue was reported by @mm-weber. The recordings where working for `cnspec shell` but not for `cnspec scan`. This brings back the ability to record the data (os provider only) for a policy.


mm-weber (approved):
Works as expected. LGTM

PR #576, opened 2023-05-18
🧹 Add no-lint flag to `cnspec bundle publish`

jaym:
There was no way to ingore linting errors. Sometimes the linter produces incorrect results


apps/cnspec/cmd/bundle.go
```diff
@@ -165,16 +167,19 @@ var policyPublishCmd = &cobra.Command{
...
+			// render cli output
+			os.Stdout.Write(result.ToCli())
 
-		if result.HasError() {
-			log.Fatal().Msg("invalid policy bundle")
+			if result.HasError() {
+				log.Fatal().Msg("invalid policy bundle")
+			}
 		}
 		log.Info().Msg("valid policy bundle")
```
chris-rock: At this point we are not sure it is valid. Compile could fail server-side. I suggest we only show that when the linting was completed successfully.

chris-rock (approved):
Great addition @jaym

apps/cnspec/cmd/bundle.go
jaym: I'm not sure I understand. Is the ask to always upload the bundle, regardless of what the lint says?

chris-rock: If you call --no-lint we essentially not verify the bundle on client side, so we cannot say at this point that the bundle is actually valid.

PR #575, opened 2023-05-18
🐛 Assign the computed filters if recomputed

jaym:
This is causing panics:
```
go.mondoo.com/cnspec/policy.(*LocalServices).ComputeBundle(0xc020fd25a8, {0xd45b998, 0xc01c77c780}, 0xc020b5e780)
	/go/pkg/mod/go.mondoo.com/cnspec@v0.0.0-20230516164323-ea5aaaad968d/policy/hub.go:405 +0x8f4
```

PR #577, opened 2023-05-19
🐛 Prevent panic for missing ComputedFilters

czunker:
This shouldn't happen, but does.
We currently do not know the reason, so first prevent panic and later try to fix the actual issue.


vjeffrey:
is this the same as https://github.com/mondoohq/cnspec/pull/575 ?

PR #578, opened 2023-05-19
🐛 Fix linters ability to decode protobuf enums

jaym:


internal/yac-it/yac-it.go
```diff
@@ -20,6 +24,10 @@ const disclaimer = `// Code generated by yac-it. DO NOT EDIT.
 const prefix = `
 import (
 	"gopkg.in/yaml.v3"
+	"encoding/json"
+	"errors"
+	"go.mondoo.com/cnquery/explorer"
+    "go.mondoo.com/cnspec/policy"
```
chris-rock: can we fix the whitespace issue here?

internal/bundle/bundle.yac.go
```diff
@@ -6,19 +6,46 @@
 package bundle
 
 import (
-	"go.mondoo.com/cnquery/explorer"
-	"go.mondoo.com/cnspec/policy"
 	"gopkg.in/yaml.v3"
+	"encoding/json"
+	"errors"
+	"go.mondoo.com/cnquery/explorer"
+    "go.mondoo.com/cnspec/policy"
```
chris-rock: This file is auto-generated but should be formatted once

internal/yac-it/yac-it.go
jaym: fixed

internal/bundle/bundle.yac.go
jaym: fixed

chris-rock (approved):
Thank you @jaym

PR #584, opened 2023-05-22
🌠 compliance frameworks

arlimus:
This is the first step in introducing Compliance Frameworks.

**Why Compliance?** Because the vast majority of security testing and reporting is used for either one of two purposes: hardening your tech to protect against attackers AND to adhere to mostly regulatory compliance. 

Security is served well with policies and prioritized results. Policies, checks, and additional data queries provide everything we need.

Compliance, however, requires a lot of structured evidence-gathering and result aggregation across a lot of technologies and policies. Additionally, it puts a much stronger emphasis on exception-handling.

**Why compliance frameworks?** An alternate route to this approach could have been to just tag all security checks with identifiers from common compliance frameworks. This is commonly good enough... until something changes. Versioning, in this model, is tedious at best and can lead to outright incorrect results at its worst. 

Together with a lack of versioning, we have seen toolsprawl and aggregation cause annoying headaches in the relationship of security and compliance.

Today we are recognizing Compliance Frameworks as a codified artifact, which derives its information from security policies, its checks and queries, as well as other controls. As such, it can be versioned, scaled, and extended.

The model that is introduced today allows users to extend frameworks and create bindings between new policies and existing frameworks. Unlike policies, which are fully maintained by their authors, compliance frameworks are an openly extensible object, which allows other users to create new bindings.

**Please note: This capability is experimental** until the release of v9 in the upcoming weeks. We appreciate your help in shaping this out tremendously! Feedback very welcome!

image

Follow-up:
- add testing
- re-introduce policy summary in compact CLI output, so we can also print the framework in the summary


policy/cnspec_policy.proto
```diff
@@ -377,15 +484,28 @@ message ReportingJob {
   // This is replaced by the new spec, which now references cnquery
   map<string,DeprecatedV7_ScoringSpec> deprecated_v7_spec = 4;
   // ^^
+  // FIXME: DEPRECATED, remove in v10.0 vv
+  // This is replaced by the new type field, which now carries more info
+  bool deprecated_v8_is_data = 8;
+  // ^^
+
+  enum Type {
```
chris-rock, line 492: Do we want this enum to be embedded or standalone?

policy/cnspec_policy.proto
```diff
@@ -422,6 +542,23 @@ message ReportCollection {
   map<string, ResolvedPolicy> resolved_policies = 5;
 }
 
+message ComplianceReport {
```
chris-rock: should this be named `ComplianceReport` to be consistent?

policy/cnspec_policy.proto
```diff
@@ -422,6 +542,23 @@ message ReportCollection {
...
+  string scoring_mrn = 1;
+  string entity_mrn = 2;
+  ComplianceScore score = 3;
+  repeated ComplianceScore controls = 4;
+}
+
+message ComplianceScore {
```
chris-rock: Should this be `FrameworkScore`?

policy/cnspec_policy.proto
```diff
@@ -550,6 +690,7 @@ service PolicyResolver {
   rpc StoreResults(StoreResultsReq) returns (Empty) {}
 
   rpc GetReport(EntityScoreReq) returns (Report) {}
+  rpc GetComplianceReport(EntityScoreReq) returns (ComplianceReport) {}
```
chris-rock: maybe `GetFrameworkReport` to be consistent?

chris-rock (approved):
Great addition @arlimus works like a charm. I added feedback naming.

policy/cnspec_policy.proto
arlimus: I was thinking of keeping this embedded, since it's limited to ReportJobs. However, if we end up ever using it in other places, I agree and we would want to externalize it. I'm not sure about it

policy/cnspec_policy.proto
arlimus: Great idea to naming it `FrameworkReport` i'll update it

policy/resolver.go
```diff
@@ -55,22 +61,24 @@ func (s *LocalServices) Assign(ctx context.Context, assignment *PolicyAssignment
 		}
 	}
 
-	// assign policy locally
-	deltas := map[string]*PolicyDelta{}
+	s.DataLake.EnsureAsset(ctx, assignment.AssetMrn)
```
jaym, line 64: error check needed?

chris-rock:
`cnspec scan -f examples/compliance.mql.yaml`

arlimus:
Great catch @czunker !! I'll do it as a follow-up.
Ditto @jaym , i'll add it in the next step.

policy/resolver.go
```diff
@@ -86,18 +94,22 @@ func (s *LocalServices) Unassign(ctx context.Context, assignment *PolicyAssignme
...
 
-	deltas := map[string]*PolicyDelta{}
+	policyActions := map[string]explorer.AssignmentDelta_Action{}
 	for i := range assignment.PolicyMrns {
 		policyMrn := assignment.PolicyMrns[i]
-		deltas[policyMrn] = &PolicyDelta{
-			PolicyMrn: policyMrn,
-			Action:    PolicyDelta_DELETE,
-		}
+		policyActions[policyMrn] = explorer.AssignmentDelta_ADD
```
czunker, line 100: This also looks wired to me. This is the function `Unassign` and we have the action `ADD`. Before, we had the action `DELETE`

policy/resolver.go
```diff
@@ -55,22 +61,24 @@ func (s *LocalServices) Assign(ctx context.Context, assignment *PolicyAssignment
...
+	frameworkActions := map[string]explorer.AssignmentDelta_Action{}
+	for i := range assignment.FrameworkMrns {
+		frameworkMrn := assignment.FrameworkMrns[i]
+		frameworkActions[frameworkMrn] = explorer.AssignmentDelta_ADD
+	}
 
-	_, err := s.DataLake.MutatePolicy(ctx, &PolicyMutationDelta{
-		PolicyMrn:    assignment.AssetMrn,
-		PolicyDeltas: deltas,
-		Action:       assignment.Action,
```
czunker, line 73: I think, we need to get this back somehow. Before this PR, it could have had the values `explorer.Action_IGNORE` or `explorer.Action_ACTIVATE`.

PR #588, opened 2023-05-26
✨ Add exceptions to controls

czunker:
This also extend the IGNORE action for checks to include a valid timestamp. ~~For this, it depends on https://github.com/mondoohq/cnquery/pull/1238~~


examples/compliance.mql.yaml
```diff
@@ -29,6 +29,18 @@ frameworks:
       title: Limit access to system configuration
     - uid: compliance-x-03
       title: Harden systems to security recommendations
+    - uid: compliance-x-04
+      title: Only use ciphers with sufficient bits
+  - title: exception-1 # something auto-generated
+    type: 4 # IGNORED/SNOOZED
```
preslavgerchev, line 35: is it both or just one? would be nice to have 2 groups in here with the 2 examples

policy/resolver.go
```diff
@@ -1378,13 +1386,30 @@ func mquery2executionQuery(query queryLike, props map[string]*llx.Primitive, pro
...
 		return found
 	}
 
+	// If we ignore this control, we have to transfer this info to the impact var,
+	// which is used to inform how to aggregate the results of all child jobs.
+	impact := &explorer.Impact{}
+	validUntil := ""
+	if frameworkGroup, ok := frameworkGroupByControlMrn[controlMrn]; ok {
+		if frameworkGroup.EndDate != 0 {
+			time.Unix(frameworkGroup.EndDate, 0).Format(time.RFC3339)
```
preslavgerchev: this looks unassigned

policy/resolver.go
```diff
@@ -1378,13 +1386,30 @@ func mquery2executionQuery(query queryLike, props map[string]*llx.Primitive, pro
...
+		}
+		if frameworkGroup.Type == GroupType_IGNORED {
```
preslavgerchev, line 1401: if this is disabled, should it also have a changed impact?

examples/compliance.mql.yaml
czunker: It's just one. The type is actually called `IGNORE`, but at some places we use the word `SNOOZE`. As I understood, going forward `SNOOZE` will win.

policy/resolver.go
czunker: No, when it is disabled, it should not have a job at all. You can also see this in the tests. The one with a `DISABLE` group has less jobs than the `IGNORE` test. Both have the same framework otherwise.

policy/resolver.go
czunker: Nice catch. Thanks. I fixed it and moved it also inside the other if block. No need to do the time conversion when we do not need it later.

preslavgerchev (approved):

PR #595, opened 2023-05-30
🐛 Set action according to function called

czunker:
When we want to unassign a policy/framework, we need to set the action to delete


czunker:
No longer needed.

PR #600, opened 2023-05-30
🧹 set minimum go version to 1.20.4

vjeffrey:
just like we did in cnquery, https://github.com/mondoohq/cnquery/pull/1258/files


arlimus (approved):
Thank you!

PR #601, opened 2023-06-01
⭐ simplify compliance structure

arlimus:
- Instead of unpacking separate from embedded compliance maps, remove support for embedde compliance maps altogether. This increases verbosity a bit at the benefit of being much simpler to understand and edit.
- Extend FrameworkGroups with override capabilities for exceptions


policy/cnspec_policy.proto
```diff
@@ -354,22 +358,25 @@ message Control {
...
+  cnquery.explorer.Action action = 41;
   bool manual = 50; // indicates that this is a manual control
 }
 
 message FrameworkMap {
   string mrn = 1;
-  repeated cnquery.explorer.ObjectRef referenced_policies = 2;
-  repeated ControlMap controls = 3;
+  string uid = 2;
+  string framework = 20;
```
preslavgerchev: nit:
maybe `framework_owner` here? to indicate better the relationship

arlimus: sure, let's give it a shot. We will want to get more feedback from users anyway before this is final

preslavgerchev (approved):

PR #604, opened 2023-06-02
call prepare framework

jaym:
I couldn't find where this was getting called and was expecting it be called before attempting to store in the data lake


preslavgerchev (approved):

arlimus (approved):
Kudos @jaym !

PR #608, opened 2023-06-04
:bug: fix fail if config file not found

mariuskimmina:
Requires https://github.com/mondoohq/cnquery/pull/1246


mariuskimmina:
Outdated

PR #609, opened 2023-06-04
🧹 remove unused flag for report output

chris-rock:
The pager wasn't working as expecting and the new defaults with web reporting and compact view do not require us to have that anymore. Lets remove it.


cli/reporter/print_report.go
```diff
@@ -76,17 +75,7 @@ func (r *reportRenderer) print() error {
 		}
 	}
 
-	if r.usePager && pager.Supported(r.pager) {
```
preslavgerchev, line 79: can we also remove those reportRenderer props: `usePager` and `r.pager`? looks like we will not be needing those

chris-rock: Yes, good call

chris-rock: Code updated and ready for re-review

czunker (approved):
LGTM Thanks @chris-rock

PR #621, opened 2023-06-16
🧹 make sharing report optional

chris-rock:
This changes the default behavior for sharing reports. Users need to pass the `--share-report` option explicitly.

### DIFF
--- a/apps/cnspec/cmd/scan.go
+++ b/apps/cnspec/cmd/scan.go
@@ -413,7 +413,7 @@ This example connects to Microsoft 365 using the PKCS #12 formatted certificate:
 		}
 
 		if conf.IsIncognito && shareReport == false {
-			otherReportOptionsMsg += "To generate a private web-based report use `--share`. "
+			otherReportOptionsMsg += "To create a web-based report with a private URL using Mondoo's reporting service, use `--share`."
 		}
 
 		if otherReportOptionsMsg != "" {
--- a/apps/cnspec/cmd/scan.go
+++ b/apps/cnspec/cmd/scan.go
@@ -340,6 +340,8 @@ This example connects to Microsoft 365 using the PKCS #12 formatted certificate:
 		cmd.Flags().StringP("output", "o", "compact", "Set output format: "+reporter.AllFormats())
 		cmd.Flags().BoolP("json", "j", false, "Set output to JSON (shorthand).")
 		cmd.Flags().Bool("share-report", false, "create sharable web-based reports when cnspec is unauthenticated. Defaults to false.")
+		cmd.Flags().MarkHidden("share-report")
+		cmd.Flags().Bool("share", false, "create sharable web-based reports when cnspec is unauthenticated. Defaults to false.")
 	},
 	CommonPreRun: func(cmd *cobra.Command, args []string) {
 		// multiple assets mapping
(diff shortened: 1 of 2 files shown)


arlimus (approved):
Thank you! I'm glad this is getting optional

charlesjohnson:
I disliked the prompt-every-run approach & I'm stoked to see it gone, but we _know_ that people who try cnspec as their introduction to Mondoo get to the CLI report and then stop there. If we're not going to offer up web-based reporting as the way to get them into the rest of our system, then I'd really like us to think about how we can better incentivize people to opt-in to the rest of what we have to offer.

apps/cnspec/cmd/scan.go
```diff
@@ -397,21 +395,28 @@ This example connects to Microsoft 365 using the PKCS #12 formatted certificate:
...
-		// if we are in an interactive terminal, running in incognito mode, and user responds "yes" then offer report viewer
-		if shareReport == nil && isatty.IsTerminal(os.Stdout.Fd()) && conf.IsIncognito {
-			response := cnspec_components.AskAYesNoQuestion("Do you want to view or share these scan results in a browser using Mondoo's reporting service?")
-			shareReport = &response
+		otherReportOptionsMsg := ""
+		if conf.Output == "compact" {
+			otherReportOptionsMsg += "For detailed CLI output, use `--output full`. "
+		}
+
+		if conf.IsIncognito && shareReport == false {
```
charlesjohnson, line 415: Is Incognito the default when cnspec is run without a Platform config?

apps/cnspec/cmd/scan.go
```diff
@@ -397,21 +395,28 @@ This example connects to Microsoft 365 using the PKCS #12 formatted certificate:
...
+			otherReportOptionsMsg += "To share the report with others use the `--share-report`. "
```
charlesjohnson:
```suggestion
			otherReportOptionsMsg += "For detailed web-based output using Mondoo's reporting service, use `--share-report`."
```

apps/cnspec/cmd/scan.go
```diff
@@ -397,21 +395,28 @@ This example connects to Microsoft 365 using the PKCS #12 formatted certificate:
 		printReports(report, conf, cmd)
 
 		// handle report sharing
-		var shareReport *bool
+		var shareReport bool
```
charlesjohnson, line 403: Should we rename this to 'web-report' for clarity?

apps/cnspec/cmd/scan.go
chris-rock: incognito is the default

apps/cnspec/cmd/scan.go
chris-rock: Maybe:

```
For detailed web-based report using Mondoo's free reporting viewer, use `--web-report`.
```

charlesjohnson:
Added this to the v9 card: https://www.notion.so/mondoo/251941963f1a4aa98c2b04f8489201a2?v=bf10d965a0b5445f96828a3138708068&p=9d25200d6726457b92cb8a6ebb005007&pm=s

charlesjohnson:
Created an internal decision doc to talk through this choice.

apps/cnspec/cmd/scan.go
chris-rock: Simplified the flag to `--share`

apps/cnspec/cmd/scan.go
chris-rock: with the current assumption that we go with --share the current name is fine

apps/cnspec/cmd/scan.go
```diff
@@ -397,21 +400,28 @@ This example connects to Microsoft 365 using the PKCS #12 formatted certificate:
...
+		otherReportOptionsMsg := ""
+		if conf.Output == "compact" {
+			otherReportOptionsMsg += "For detailed CLI output, use `--output full`. "
+		}
 
-		if viper.IsSet("share-report") {
-			shareReportFlag := viper.GetBool("share-report")
-			shareReport = &shareReportFlag
+		if conf.IsIncognito && shareReport == false {
+			otherReportOptionsMsg += "To generate a private web-based report use `--share`. "
```
charlesjohnson:
```suggestion
			otherReportOptionsMsg += "To create a web-based report with a private URL using Mondoo's reporting service, use `--share`."
```

arlimus (approved):
Awesome, thank you! glad to change this

PR #628, opened 2023-06-23
⭐️ Randomize the start time of scanning in serve mode

jaym:
This makes sure that scans get randomly distributed throughout the timer time in an effort to spread out load against the api server.


chris-rock (approved):
Thank you @jaym

PR #632, opened 2023-06-26
🧹 pluralize framework and frameworkmap resource

jaym:
queries, policies, assets, and controls are pluralized. framework and frameworkmap should be for consistency


arlimus (approved):
yes! kudos :)

PR #636, opened 2023-06-27
Preserve the framework dependencies in the reporting structure

jaym:
There are major changes here:

First, compliance frameworks were not connected to the root reporting job. On the backend, we iterate from the root reporting job to figure out all reporting jobs that get saved. Without this, we don't save the compliance frameworks or controls

Second change is the structure of the framework dependencies is preserved in the reporting jobs. For example, asset frameworks point to the space framework which point to the global framework which point to some actual frameworks. This keeps it consistent with the way handle policies.

One issue I ran into was that we have space/asset frameworks and space/asset policies. I chose to not create separate reporting jobs for those because having multiple reporting jobs with the same query is likely to break something. Instead, the policies would just get attached to the existing policy jobs, but the impact is set to unscored

reporting_structure

### DIFF
--- a/policy/framework.go
+++ b/policy/framework.go
@@ -24,8 +24,6 @@ type ResolvedFramework struct {
 	// E.g. ReportSources[controlA] = [check123, check45]
 	// E.g. ReportSources[frameworkX] = [controlA, ...]
 	ReportSources map[string][]string
-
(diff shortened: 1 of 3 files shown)


arlimus (approved):
Looks good, thank you for wiring this up @jaym !

PR #639, opened 2023-06-29
:broom:  Incorporate framework maps into the framework checksum 

jaym:
With this, frameworks graph checksums will change if any of their
framework maps change. This allows us to correctly pick up changes to
the map during building of the resolved policy. Without this, we'd just
keep picking up the cached resolved policy when the framework maps
change because the graph checksum wouldn't change


policy/resolver_test.go
```diff
@@ -418,6 +429,35 @@ framework_maps:
...
+
+		rpInitial, err := srv.Resolve(context.Background(), &policy.ResolveReq{
+			PolicyMrn:    "asset1",
+			AssetFilters: []*explorer.Mquery{{Mql: "true"}},
+		})
+		require.NoError(t, err)
+		require.NotNil(t, rpInitial)
+
+		bFrameworkUpdate := parseBundle(t, bundleStr)
+		bFrameworkUpdate.Frameworks[0].Groups[0].Controls = bFrameworkUpdate.Frameworks[0].Groups[0].Controls[:2]
```
czunker, line 448: Do you need to explicitly update the checksums after this?

policy/resolver_test.go
```diff
@@ -353,9 +353,18 @@ frameworks:
       title: control2
     - uid: control3
       title: control3
+- uid: framework2
+  name: framework1
```
czunker: nit: Does it help to also change the name here?

policy/resolver_test.go
jaym: checksums should get recalculated later

policy/resolver_test.go
```diff
@@ -572,9 +614,24 @@ framework_maps:
...
+		var frameworkJob *policy.ReportingJob
+		for _, rj := range rp.CollectorJob.ReportingJobs {
+			if rj.QrId == "//test.sth/framework/mondoo-ucf" {
+				frameworkJob = rj
+				break
+			}
+		}
+		require.NotNil(t, frameworkJob)
 		require.Equal(t, frameworkJob.Type, policy.ReportingJob_FRAMEWORK)
-		require.Equal(t, explorer.ScoringSystem_IGNORE_SCORE, frameworkJob.ChildJobs["wzZosRUzLRQ="].Scoring)
```
czunker: More asking out of curiosity as I have the same kind of tests in another PR:

Aren't the checksums stable? Or is this just the easier way in the long run when we change the input?

czunker (approved):

policy/resolver_test.go
jaym: they're not that stable. In this case they seem to have changed. only mrn is stable. so i decided to make it invariant to the checksum

PR #640, opened 2023-06-30
Add compliant and non-compliant asset count.

preslavgerchev:
Add # of compliant and failed assets to the ControlScore struct


czunker (approved):
LGTM

PR #645, opened 2023-07-07
✨ Add handling of rejected exceptions

czunker:
When exceptions are marked as rejected, they will be skipped.


czunker:
This is currently based on https://github.com/mondoohq/cnspec/pull/639

czunker:
The diff around `TestResolve_Exceptions` looks a bit strange. I merged the existing tests into one bigger test to not Copy&Paste the framework bundle again and again.

Then I added new test cases by modifying the bundle directly:
- resolve with ignored control and validUntil
- resolve with expired validUntil
- resolve with rejected disable exception

policy/cnspec_policy.proto
```diff
@@ -343,7 +343,11 @@ message FrameworkGroup {
   string title = 24;
   PolicyGroupDocs docs = 25;
   repeated cnquery.explorer.Author authors = 26;
-  repeated cnquery.explorer.Author approvers = 27;
+  repeated cnquery.explorer.Author reviewers = 27;
+  // only applies to GroupType IGNORED/DISABLED
+  // Only if true ignore the group
+  // Which means that unapproved groups get handled as if they were approved
+  bool rejected = 28;
```
preslavgerchev: do we have a use case for approved exception too? if thats true, it might better to define this as an enum state instead of a bool so we can also express that

czunker: We currently only care about `rejected`, but still this is a nice idea. That will make parts of the code easier to read and more clear. Thanks.

czunker: I changed the type. Could you please have another look?

policy/cnspec_policy.proto
```diff
@@ -343,7 +349,11 @@ message FrameworkGroup {
   string title = 24;
   PolicyGroupDocs docs = 25;
   repeated cnquery.explorer.Author authors = 26;
-  repeated cnquery.explorer.Author approvers = 27;
+  repeated cnquery.explorer.Author reviewers = 27;
+  // only applies to GroupType IGNORED/DISABLED
+  // Only if true ignore the group
```
preslavgerchev: this comments needs slight update, 'only if true ignore the group'

czunker: Good catch. :+1:  Is fixed now.

preslavgerchev (approved):
thank you @czunker

PR #646, opened 2023-07-07
🐛 Check frameworks when assigning and bailing early

jaym:


policy/resolver.go
```diff
@@ -38,7 +38,7 @@ type AssetMutation struct {
 // 2. asset is local (via incognito mode) but policy is upstream
 // 3. asset and policy are upstream
 func (s *LocalServices) Assign(ctx context.Context, assignment *PolicyAssignment) (*Empty, error) {
-	if len(assignment.PolicyMrns) == 0 {
+	if len(assignment.PolicyMrns)+len(assignment.FrameworkMrns) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "a policy mrn is required")
```
preslavgerchev: nit:
```suggestion
		return nil, status.Error(codes.InvalidArgument, "a policy or framework mrn is required")
```

preslavgerchev (approved):

PR #652, opened 2023-07-12
🧹 Use ObjectRef for the framework map owner field

jaym:
Having a field that could have a uid or mrn could end up being ambiguous. Using ObjectRef makes it more consistent with the way other fields handle owner or ref


arlimus (approved):
Looks great @jaym thank you!

PR #658, opened 2023-07-19
🧹 disable cgo since we do not need it

chris-rock:


chris-rock:
That prevents including the sqlite loading

PR #660, opened 2023-07-20
🐛 Use action when calculating framework graph checksums

jaym:
Without this, changing a framework from preview to active produces no change in the framework checksums, and thus we cant store such updates


arlimus (approved):
great catch!

PR #664, opened 2023-07-26
🧹 Don't unnecessarily recalculate framework dep checksums

jaym:
This causes us to load frameworks to calculate checksums we already have. If checksums need to be recalculated, they get cleared first


vjeffrey (approved):

vjeffrey:
thank you @jaym, makes sense!

PR #666, opened 2023-07-27
🧹 do not reorder queries and checks in a policy when updating checksums

imilchev:
This PR makes sure we are not changing the orders of queries and checks when we update checksums


arlimus (approved):
nice job! nit: could you add a comment as to why we copy it in the go-code there, something along the lines of `// copy checks to keep the original order and only sort it for the purpose of checksum generation`

PR #673, opened 2023-08-01
🐛 The controlmap control uids also need to be cleared

jaym:
Otherwise, you cannot compile a bundle twice. And when you run locally, the bundle gets compiled during validation and then when it gets set in the data store


chris-rock (approved):
Thank you @jaym

PR #678, opened 2023-08-03
🧹 Use average scoring for frameworks

jaym:
Frameworks get controls. Controls are pass or fail (0 or 100). By changing the scoring system to average, the value of the framework score is the completion


chris-rock (approved):
Thank you @jaym

PR #680, opened 2023-08-04
🧹 Change how controls are scored

jaym:
Any child of a control that errors is converted into a fail. Any child that is skipped or unscored is converted to pass. This way, controls always have a result


chris-rock (approved):
Thank you @jaym

PR #681, opened 2023-08-04
🧹 Modify example compliance.mql.yaml to be compatible with Mondoo cloud

jaym:
Exceptions cannot be managed inline with the framework in this manner.


chris-rock (approved):
Thank you @jaym

PR #687, opened 2023-08-08
🐛 Fix check control resolver bug

jaym:
When we have controls that map to check, the resolver creates a reporting job from the query by its code id to query by its mrn. And then maps that to the control. However, there was a bug in this translation where we would pick a random check mrn from the bundle that had that code id. Since its possible for multiple checks to have the same id in a bundle, and not all of those checks are part of an active policy, this would end up picking a mrn from a policy that wasn't active.

### DIFF
--- a/policy/resolver_test.go
+++ b/policy/resolver_test.go
@@ -340,7 +340,16 @@ policies:
       mql: 1 == 1
     - uid: check-pass-2
       mql: 2 == 2
-
+- uid: policy-inactive
+  groups:
+  - filters: "false"
+    checks:
+    - uid: inactive-fail
+      mql: 1 == 2
+    - uid: inactive-pass
+      mql: 1 == 1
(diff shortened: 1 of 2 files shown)


policy/resolver_test.go
```diff
@@ -441,6 +452,10 @@ framework_maps:
 		rjTester.requireReportsTo(controlMrn("control4"), frameworkMrn("framework1"))
 		rjTester.requireReportsTo(frameworkMrn("framework1"), frameworkMrn("parent-framework"))
 		rjTester.requireReportsTo(frameworkMrn("parent-framework"), "root")
+
+		require.Nil(t, rjTester.queryIdToReportingJob[queryMrn("inactive-fail")])
```
chris-rock, line 456: nice test @jaym

chris-rock (approved):
This looks awesome! Thank you @jaym

PR #688, opened 2023-08-08
Do not report scores for frameworks without any controls implemented

preslavgerchev:
If the framework has no children, remove it from the reporting jobs. This ensures we do not report framework scores for frameworks that have 0 controls reporting.


czunker:
Do we still need this PR? The linked issue is already closed.

preslavgerchev:
Yep, we still need to fix that, although in a totally different way than what I have here. I will close this PR and re-open once it's ready

PR #690, opened 2023-08-13
🧹 remove internal vault since it is easier to configure via inventory file

chris-rock:
brings https://github.com/mondoohq/cnquery/pull/1500 to cnspec


chris-rock:
once https://github.com/mondoohq/cnquery/pull/1500 is merged, we need to update the cnquery dependency

atomic111 (approved):
@chris-rock thanks for the enhancement.

PR #712, opened 2023-09-05
🌠 V9

arlimus:
This is one of the biggest updates to cnspec since we first released it!

It uses v9 of cnquery, which provides modular provider plugins which gives 3 big advantages:

1. **The binaries are 90% smaller**, from 277mb => 30mb, i.e. 90% reduction in size
2. **Both cnquery and cnspec share the same providers**, so no more duplication, i.e. another 50% smaller
3. **You can now write your own providers!** They are not mandatory to be in our managed repo anymore. In fact, as long as you adhere to the same interface (which is Terraform plugin) **you can write providers in any language and even keep them internal** (which for some custom systems is incredibly useful). Of course **we ♥️ providers** so feel free to add suggestions and things you'd love to see.

Checklist of related PRs:
- [x] https://github.com/mondoohq/cnquery/pull/1635
- [x] update testutils for cnquery to work on easy import
- [x] fix tests


chris-rock (approved):
Wooot

PR #719, opened 2023-09-12
🐛 Pass in the correct incognito value when calling RunIncognito.

preslavgerchev:


czunker (approved):
Nice catch. Thanks @preslavgerchev

vjeffrey (approved):

PR #724, opened 2023-09-15
🧹 Dump resolved policy DOT file if TRACE env var is set.

preslavgerchev:
Same as https://github.com/mondoohq/cnquery/pull/1742


chris-rock (approved):
Thank you @preslavgerchev

PR #741, opened 2023-09-21
🐛 make filter queries optional

arlimus:
Currently filter queries that error during the compile stage will cause the execution to fail. However, with the new providers model, certain contents are now optional, thus the local compiler (given locally available schemas) will throw errors. We simply eliminate these filters from the result list.

Example use-case: Have a scanner that only scans github. When it encounters filters, it will get a long list of technologies that are unsupported (like clouds and k8s). These can now simply be eliminated and moved on.


chris-rock (approved):
Thank you @arlimus

PR #748, opened 2023-09-25
✨ rename fleet => inventory

arlimus:
This better matches the more commonly used definition. We will keep fleet working for the next  major version and then slowly deprecate it.

Analogous to: https://github.com/mondoohq/cnquery/pull/1920


chris-rock (approved):
Thank you @arlimus

PR #762, opened 2023-09-27
🐛 Initialize policy remote services instead of queryhub remote services.

preslavgerchev:
Make sync assets go through the `/PolicyResolver` endpoint instead of `/QueryConductor`


vjeffrey (approved):

vjeffrey:
thank you!

PR #782, opened 2023-10-02
⚡ re-use runtimes across providers

arlimus:
depends on cnquery https://github.com/mondoohq/cnquery/pull/2015


chris-rock (approved):
Thank you for the cleanup

PR #784, opened 2023-10-02
🐛 fix GarbageCollectAssets authentication for local scanner

imilchev:
The `GarbageCollectAssets` call was throwing HTTP 403 because the plugins were not set correctly


chris-rock (approved):
Thank you @imilchev

PR #794, opened 2023-10-03
🧹 apply runtime env labels for discovered assets

imilchev:
this is needed for ci/cd assets

### DIFF
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -303,7 +303,11 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 
 	runtimeEnv := execruntime.Detect()
 	var runtimeLabels map[string]string
-	if runtimeEnv != nil {
+	// If the runtime is an automated environment and the root asset is CI/CD, then we are doing a
+	// CI/CD scan and we need to apply the runtime labels to the assets
+	if runtimeEnv != nil &&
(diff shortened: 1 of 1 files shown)


policy/scan/local_scanner.go
```diff
@@ -301,9 +301,21 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
...
+	runtimeEnv := execruntime.Detect()
+	var runtimeLabels map[string]string
+	if runtimeEnv != nil {
+		runtimeLabels = runtimeEnv.Labels()
+	}
+
 	justAssets := []*inventory.Asset{}
 	for _, asset := range assets {
 		asset.asset.KindString = asset.asset.GetPlatform().Kind
+		for k, v := range runtimeLabels {
```
vjeffrey, line 317: do we not wanna check if its an automated env before applying all these? (saw that we have that check in v8)

vjeffrey: `	if opts.AutoDetectCICDCategory && runtimeEnv.IsAutomatedEnv() || opts.Category == "cicd" {`

imilchev: we cannot check this in the same way in this part of the code. I am now checking if the root asset is ci/cd. This should indicate that we are doing a ci/cd scan

benr (approved):

PR #800, opened 2023-10-04
🐛 copy labels from root asset to child assets

imilchev:
This is something we used to do in v8. It is required to get the k8s admission controller to work as well


imilchev:
not needed anymore

PR #802, opened 2023-10-04
✨ re-use provider instance for all assets when scanning k8s

imilchev:
do not use ephemeral runtimes for discovered k8s assets. Needed to optimize the memory usage for k8s since we scan hundreds of assets there

### DIFF
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -281,9 +281,17 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 	for i := range assetCandidates {
 		candidate := assetCandidates[i]
 
-		runtime, err := providers.Coordinator.EphemeralRuntimeFor(candidate.asset)
-		if err != nil {
-			return nil, false, err
+		var runtime *providers.Runtime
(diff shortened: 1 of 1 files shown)


policy/scan/local_scanner.go
```diff
@@ -443,7 +451,9 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 			})
 
 			// shut down all ephemeral runtimes
-			runtime.Close()
+			if asset.Connections[0].Type != "k8s" {
```
czunker: I'm, not sure we will remember why this is there, when we look at it later.

Perhaps we better expose https://github.com/mondoohq/cnquery/pull/2050/files#diff-7263c3839c73627bf39872d6eafc82a803d9b052ece899a5a14fb5cf8497d844R40 and check for it here?

My current gut feeling is, that we will re-use this also for container registries, AWS, ...

imilchev: if seems like this condition isn't needed at all since the shutdown thing already checks that only ephemeral providers are killed

czunker (approved):
LGTM. Thanks @imilchev

PR #804, opened 2023-10-04
🧹 Add err to the reporter if we cant connect to an asset.

preslavgerchev:
This is an attempt at getting better insight into scan errors. Until now, we ignore errors if we cant connect to an asset. Those errors are now logged into the reporter which is then returned if no assets are present instead of returning `nil, false, nil`

### DIFF
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -249,6 +249,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 			Asset:    resolvedAsset,
 			Upstream: upstream,
 		}); err != nil {
+			reporter.AddScanError(resolvedAsset, err)
 			log.Error().Err(err).Msg("unable to connect to asset")
 			continue
 		}
@@ -292,6 +293,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 			Upstream: upstream,
 		})
 		if err != nil {
+			reporter.AddScanError(candidate.asset, err)
 			log.Error().Err(err).Msg("unable to connect to asset")
 			continue
 		}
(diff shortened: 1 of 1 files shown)


imilchev (approved):

policy/scan/local_scanner.go
```diff
@@ -249,6 +249,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 			Asset:    resolvedAsset,
 			Upstream: upstream,
 		}); err != nil {
+			reporter.AddScanError(resolvedAsset, err)
```
czunker, line 252: Is there a way, we can test this?

As these are connect errors, I'm not sure these will end up as intended here: https://github.com/mondoohq/cnspec/blob/5438e55fcc06668ea7c35466664d96b6c97e8ea6/policy/scan/aggregate_reporter.go#L46-L47

I worked at something similar today and had no assert information.

preslavgerchev: that's a good point, we will end up with errs being overwritten here. need to go back to the drawing board on this one

chris-rock (commented):
I like to see a test for it and also see how this renders with to json

policy/scan/local_scanner.go
```diff
@@ -292,6 +293,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 			Upstream: upstream,
 		})
 		if err != nil {
+			reporter.AddScanError(candidate.asset, err)
```
vjeffrey, line 296: can we collect the errors and return errors.Join ?

imilchev:
i think that's no longer needed because of #1080. Please close it if that is the case

chris-rock:
superseeded by #1080

PR #822, opened 2023-10-06
🧹 shutdown coordinator after scan

imilchev:
Make sure all providers are killed after scan finishes


czunker (approved):
LGTM. Thanks @imilchev

PR #824, opened 2023-10-06
🧹 run 1 scan at a time via scan API

imilchev:
Running scans in parallel causes issues with killing providers from previous scans. This is a temporary fix for the scan API such that only 1 scan runs at a time. We can re-write the code to kill only the specific provider used in the current scan, instead of all providers at a later stage.

Running 1 scan at a time isn't really an issue at the moment since only the `mondoo-operator` is using the scan API. There we always run 1 cluster scan at a time. The things that could run next to it are single resource scans that are coming via the internal queue. They are not time-sensitive and aren't blocking anything.

### DIFF
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -58,6 +58,7 @@ type LocalScanner struct {
 	// allows setting the upstream credentials from a job
 	allowJobCredentials bool
 	disableProgressBar  bool
+	scanMutex           sync.Mutex
 }
 
 type ScannerOption func(*LocalScanner)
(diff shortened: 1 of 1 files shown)


policy/scan/local_scanner.go
```diff
@@ -130,6 +131,8 @@ func (s *LocalScanner) Schedule(ctx context.Context, job *Job) (*Empty, error) {
 }
 
 func (s *LocalScanner) Run(ctx context.Context, job *Job) (*ScanResult, error) {
+	s.scanMutex.Lock()
+	defer s.scanMutex.Unlock()
```
mariuskimmina, line 140: Maybe we should leave a comment here so that we know in the future why this was necessary? Maybe just a link to this PR?

mariuskimmina (approved):
One small suggestion, overall LGTM

policy/scan/local_scanner.go
imilchev: added a comment

PR #841, opened 2023-10-15
⭐ wire data queries to controls

arlimus:
This creates the internal wiring to collect data queries for controls. It connects each datapoint that a query produces to all the reporting job of the control that it is mapped to.

Follow-up: I noticed that we don't do a decent job on printing just yet, e.g. we only print if a control passes or fails bound to its score. That's why there isn't a good place to add datapoints yet, considering we already print all data that is executed for a policy. Definitely looking to improve the CLI experience.

@tas50 @jaym @preslavgerchev


jaym (approved):
this case should probably be added to the tests in `resolver_test.go`

examples/compliance.mql.yaml
```diff
@@ -19,6 +19,10 @@ policies:
       title: SSH config editing should be limited to admins
       mql: sshd.config.file.permissions.mode == 0644
       impact: 100
+    queries:
```
preslavgerchev, line 22: do we need do add query pack dependencies to support this in the maps?

similar to `policy_dependencies` but for query packs instead

policy/resolver.go
```diff
@@ -1578,6 +1588,41 @@ func (s *LocalServices) jobsToControls(cache *frameworkResolverCache, framework
...
+			for controlMrn := range targets {
+				controlJob := ensureControlJob(cache, nuJobs, controlMrn, framework, frameworkGroupByControlMrn)
+				controlJob.ChildJobs[queryJob.Uuid] = nil
+				queryJob.Notify = append(queryJob.Notify, controlJob.Uuid)
+				err := connectDatapointsToReportingJob(execQuery, controlJob, job.Datapoints)
+				if err != nil {
+					return err
+				}
+				// feed the datapoints into the query's reporting job too, not just the control job
+				err = connectDatapointsToReportingJob(execQuery, queryJob, job.Datapoints)
```
preslavgerchev: not sure if needed, but i thought we might want to feed the datapoints nto the new reporting jobs too @arlimus

examples/compliance.mql.yaml
arlimus: oh good point! I'm not sure on the wiring we are using internally, but essentially we just want to point people to the MRN of the policy or querypack they want to activate at the end of the day

policy/resolver.go
arlimus: I think we'd only need to add them to the queryJob, right? Because the queryjob feeds into the control job, so the control job gets its data from that if I'm not mistaken

preslavgerchev: Good point, I moved this to only report to the query job: datapoint -> query job -> control job

examples/compliance.mql.yaml
preslavgerchev: I added a `query_pack_dependencies` field to the proto. We use the policy/pack deps when compiling the framework maps to ensure that the policies/packs are in the bundle.

Note that packs come back as policies in the bundle but we only reference them by `uid` so that should be fine

arlimus:
Looks great, thank you for the updates!

PR #845, opened 2023-10-16
:broom: Set correct filter for `cnspec bundle init` policy

mm-weber:
From `platform.` to `asset.`

Not sure if that's the right place.

I want to modify the `example-policy.mql.yml` created when using `cnspec bundle init`


chris-rock (approved):

chris-rock:
Great addition @mm-weber

PR #853, opened 2023-10-17
🧹 create a resilient v8 and v9+ vuln report

arlimus:
We recently hotfixed an issue that caused the report not to show on v9+, because older v8 code-paths were taken: https://github.com/mondoohq/cnspec/pull/850

When the upstream reporting eventually switches over to use `asset` instead of `platform`, this hotfix will fail. This PR create a more resilient approach to the vulnerability report, supporting both v8 and v9+


chris-rock (approved):
Thank you @arlimus

PR #854, opened 2023-10-17
🐛 Fix some typos

czunker:


czunker:
Came across some typos while debugging something else.

PR #867, opened 2023-10-20
🐛 do not print error if vulns are not supported

arlimus:
Many providers do not support vulnerability reports, eg: Anything that requests APIs or files like Terraform/K8s. We support it for things like OSs and OS-related targets.

Example OLD:

```bash
...
✕ Fail:         Verify shielded VM is enabled on compute instances
✓ Pass:         Legacy metadata endpoints enabled

Vulnerabilities:
error: Could not find the vulnerability report.

Scanned 1 asset

Terraform HCL
    C Terraform Static Analysis directory pass
```

Example NEW:

```bash
✓ Pass:         VM disk encryption keys should not be provided in plaintext
. Skipped:      Service accounts should not have roles assigned with excessive privileges


Scanned 1 asset

Terraform HCL
    C Terraform Static Analysis directory pass
```

### DIFF
--- a/cli/reporter/junit.go
+++ b/cli/reporter/junit.go
@@ -163,10 +163,7 @@ func assetMvdTests(r *policy.ReportCollection, assetMrn string, assetObj *invent
 	}
 
 	rawResults := results.RawResults()
-	value, err := getVulnReport(rawResults)
-	if err != nil {
-		return nil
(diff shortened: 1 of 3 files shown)


chris-rock (approved):
Thank you @arlimus

PR #868, opened 2023-10-20
🧹 simplify default provider install

arlimus:
Requires https://github.com/mondoohq/cnquery/pull/2306

Uses provider IDs instead of other (confusing) means, after I finally cleaned up my original code.


chris-rock (approved):
Nice simplification

PR #873, opened 2023-10-20
⭐️ Add conditions to the systemd unit file

benr:
This is in response to https://github.com/mondoohq/installer/issues/279

Adding these conditions will cause the systemd unit to fail is a more elegant way than simply failing because the files aren't present, in the case that the service wasn't registered.  Because we install the unit file with the packages there isn't a direct interface to the user to instruct them to register before instantiating the service.


chris-rock (approved):
Thank you @benr

PR #879, opened 2023-10-24
🐛 Fix creation of reporting job for missing or non-matching data queries.

preslavgerchev:
This PRs fixes 2 issues:
 1. If we have a framework that references a query but that query is not part of the bundle because the pack or policy are not activated, it will not be present during execution. Add a guard + test to ensure we only add reporting jobs for present queries
 2. If there's an active query but it does not match the asset (wrong asset filters), it will not be part of the execution queries. Note it is still part of the bundle, since the pack or policy is active but its not a query that the asset will run. Add a guard + test

### DIFF
--- a/policy/resolver.go
+++ b/policy/resolver.go
@@ -1594,8 +1594,12 @@ func (s *LocalServices) jobsToControls(cache *frameworkResolverCache, framework 
 				continue
 			}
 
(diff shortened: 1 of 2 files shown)


arlimus (approved):
Nice catch on this!!

PR #885, opened 2023-10-25
🐛 fix --asset-name flag

### LINKED ISSUE #858: 🐛 --asset-name is not rendered in Mondoo Platform
**Describe the bug**
`cnspec scan <target> --asset-name myasset` does not render in Mondoo Platform. This used to work in v8. I tested with `terraform-plan` and `aws`. 

**To Reproduce**

`cnspec scan aws --asset-name aws-demo-account`

**Expected behavior**

**Screenshots or CLI Output**
If applicable, add screenshots or the CLI output to help explain your problem.

**Desktop (please complete the following information):**

- OS: [e.g. macOS]
- OS Version: [e.g. 13.0]
- Browser if applicable: [e.g. Chrome, Firefox]
- Browser Version: [e.g. 106]

**Additional context**

mariuskimmina:
Closes #858 

```
cnspec scan ssh root@192.168.122.158 --ask-pass --config demo.agent.credentials.json --asset-name IAMTHEASSETNAMENOW
```

And here the result in console:

image

### DIFF
--- a/apps/cnspec/cmd/scan.go
+++ b/apps/cnspec/cmd/scan.go
@@ -96,6 +96,7 @@ To manually configure a policy, use this:
 		viper.BindPFlag("inventory-domainlist", cmd.Flags().Lookup("inventory-domainlist"))
 		viper.BindPFlag("policy-bundle", cmd.Flags().Lookup("policy-bundle"))
 		viper.BindPFlag("detect-cicd", cmd.Flags().Lookup("detect-cicd"))
+		viper.BindPFlag("asset-name", cmd.Flags().Lookup("asset-name"))
 		viper.BindPFlag("category", cmd.Flags().Lookup("category"))
 		viper.BindPFlag("score-threshold", cmd.Flags().Lookup("score-threshold"))
 		viper.BindPFlag("share", cmd.Flags().Lookup("share"))
--- a/apps/cnspec/cmd/scan.go
+++ b/apps/cnspec/cmd/scan.go
@@ -225,7 +225,7 @@ func getCobraScanConfig(cmd *cobra.Command, runtime *providers.Runtime, cliRes *
 	if err != nil {
 		log.Fatal().Err(err).Msg("failed to parse asset-name")
 	}
-	if assetName != "" {
+	if assetName != "" && cliRes.Asset != nil {
 		cliRes.Asset.Name = assetName
 	}
 
--- a/apps/cnspec/cmd/scan.go
+++ b/apps/cnspec/cmd/scan.go
@@ -221,6 +221,14 @@ func getCobraScanConfig(cmd *cobra.Command, runtime *providers.Runtime, cliRes *
 		optAnnotations[k] = v
 	}
 
+	assetName, err := cmd.Flags().GetString("asset-name")
+	if err != nil {
+		log.Fatal().Err(err).Msg("failed to parse asset-name")
+	}
+	if assetName != "" {
+		cliRes.Asset.Name = assetName
+	}
+
 	inv, err := inventoryloader.ParseOrUse(cliRes.Asset, viper.GetBool("insecure"), optAnnotations)
 	if err != nil {
 		log.Fatal().Err(err).Msg("failed to parse inventory")
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -381,7 +381,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 			assets[i].asset.Url = platformAssetMapping[platformMrn].Url
 		}
 	} else {
-		// ensure we have no non-empty asset MRNs
+		// ensure we have non-empty asset MRNs
 		for i := range assets {
 			cur := assets[i]
 			if cur.asset.Mrn == "" {
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -16,6 +16,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog/log"
 	"github.com/segmentio/ksuid"
+	"github.com/spf13/viper"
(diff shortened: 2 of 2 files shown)


mariuskimmina:
Marking as draft for now as I am not sure this works properly in all cases

mariuskimmina:
Alternatively, we could also do something like this

```
  // if the flag --asset-name was used and there is only one asset, override the name
  if len(assets) == 1 {
		assetName := viper.GetString("asset-name")
    if assetName != "" {
      assets[0].asset.Name = viper.GetString("asset-name")
    }
  }
```

I am not really sure how `--asset-name` should behave if there is more than 1 asset

policy/scan/local_scanner.go
```diff
@@ -374,14 +375,20 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 		}
 
 		// attach the asset details to the assets list
+		// also override asset name if flag --asset-name was used
 		for i := range assets {
 			log.Debug().Str("asset", assets[i].asset.Name).Strs("platform-ids", assets[i].asset.PlatformIds).Msg("update asset")
 			platformMrn := assets[i].asset.PlatformIds[0]
 			assets[i].asset.Mrn = platformAssetMapping[platformMrn].AssetMrn
 			assets[i].asset.Url = platformAssetMapping[platformMrn].Url
+			assetName := viper.GetString("asset-name")
```
preslavgerchev: we should pass that in from the caller of this func, we dont want to use viper here

ideally you want to attach this name to the root asset (cliAsset)

policy/scan/local_scanner.go
```diff
@@ -391,6 +398,10 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 					return nil, false, multierr.Wrap(err, "failed to generate a random asset MRN")
 				}
 				cur.asset.Mrn = x.String()
+				assetName := viper.GetString("asset-name")
```
preslavgerchev: same comment as above

policy/scan/local_scanner.go
preslavgerchev: on the question of how to work with this when there is more than a single asset i think we should either:
1. ignore the flag in that case
2. only apply it to the primary asset, targeted for connection. we return an asset and an inventory from the connect call. we should probably only apply the flag to the asset and not the inventory

@chris-rock any thoughts?

policy/scan/local_scanner.go
preslavgerchev: actually, i don't think this is even needed. the `assets` list gets modified. above as well and the name gets assigned

policy/scan/local_scanner.go
```diff
@@ -381,7 +381,7 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 			assets[i].asset.Url = platformAssetMapping[platformMrn].Url
 		}
 	} else {
-		// ensure we have non-empty asset MRNs
+		// ensure we have no non-empty asset MRNs
```
vjeffrey: i think the comment was ok as it was...we want to make sure the mrns are not empty, adding no there makes it seem like we want empty mrns

mariuskimmina: I added it because I thought that's the intention to **not** have non-empty mrns - maybe here it shines through that english is not my native language but to me it sounded wrong without the no

mariuskimmina: Changed it back

policy/scan/local_scanner.go
```diff
@@ -395,6 +395,12 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 		}
 	}
 
+	// ensure that the root asset has keept his name from --asset-name
+	if job.Inventory.Spec.Assets[0].Name != assets[0].asset.Name {
```
preslavgerchev: i dont think we can guarantee the inventory's asset ordering here. should just attaching this to the `cliAsset` not be enough?

apps/cnspec/cmd/scan.go
```diff
@@ -220,6 +221,14 @@ func getCobraScanConfig(cmd *cobra.Command, runtime *providers.Runtime, cliRes *
 		optAnnotations[k] = v
 	}
 
+	assetName, err := cmd.Flags().GetString("asset-name")
+	if err != nil {
+		log.Fatal().Err(err).Msg("failed to parse asset-name")
+	}
+	if assetName != "" {
+		cliRes.Asset.Name = assetName
```
preslavgerchev, line 229: is it guaranteed that the `cliAsset` is not nil?

policy/scan/local_scanner.go
mariuskimmina: It wasn't no, when just doing it on the cli asset it didn't work at all

apps/cnspec/cmd/scan.go
mariuskimmina: Not sure, I'll add a check

policy/scan/local_scanner.go
```diff
@@ -395,6 +395,13 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 		}
 	}
 
+	// if there is exactly one asset, assure that the --asset-name is used
+	// TOOD: make it so that the --asset-name is set for the root asset only even if multiple assets are there
+	if len(assets) == 1 {
```
preslavgerchev: we should be doing this earlier, this is post SynchronizeAssets

there's a check for `len(assets == 0)` somewhere up, just before the sync call.  You can also use `assetList` as that loads up the job inventory:
```
if len(assets) == 1 && assetsList[0].Name != assets[0].Name) {
 ...
 }
 ```
also do please write a comment that this is a temp fix

preslavgerchev:
lgtm!

preslavgerchev (approved):

PR #888, opened 2023-10-25
🐛 Fix adding ignored checks in the resolver

preslavgerchev:
Previously, we used to annotate ignored checks with the `MODIFY` action, we now have a `IGNORE` action too. Clean up of all v7 deprecated code, related to the scoring spec and add a test for a disabled query.

I still need to write a test for a policy with an ignored query


vjeffrey:
well, the ignored checks work great with a smaller sample policy, so we know this def improves things

vjeffrey (approved):

PR #889, opened 2023-10-25
bump cnquery to v9.3.3

mariuskimmina:


cnspec
vjeffrey, line 1: looks like a cnspec binary landed in here

mariuskimmina: opps, good catch, it's gone now

vjeffrey (approved):

PR #893, opened 2023-10-26
Add a test for resolving a policy with an ignored query.

preslavgerchev:
Not sure if this is the best way to write that test, but only way I could have it get an asset policy with the ignored check in.

### DIFF
--- a/policy/resolver_test.go
+++ b/policy/resolver_test.go
@@ -264,6 +264,79 @@ policies:
 	}
 }
 
+func TestResolve_IgnoredQuery(t *testing.T) {
+	b := parseBundle(t, `
+owner_mrn: //test.sth
+policies:
+- uid: policy-1
+  owner_mrn: //test.sth
+  groups:
+  - type: chapter
+    filters: "true"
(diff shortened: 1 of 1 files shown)


policy/resolver_test.go
```diff
@@ -264,6 +264,79 @@ policies:
...
+	}, []*policy.Bundle{b})
+
+	bundleMap, err := b.Compile(context.Background(), schema, nil)
+	require.NoError(t, err)
+
+	// we get the asset policy, manually apply the ignored action and store it back
+	ap, err := srv.DataLake.GetRawPolicy(context.Background(), "asset1")
+	require.NoError(t, err)
+	require.NotNil(t, ap)
+	ap.Groups[0].Checks = []*explorer.Mquery{
```
jaym: does setting action in the policy directly not work like we do in `TestResolve_DisabledQuery`?

preslavgerchev: yep, had. to play a bit with it but got it working

jaym (approved):

PR #907, opened 2023-10-27
🐛 make junit exporter more robust

chris-rock:
This change adds a new test to the junit exporter to deal with empty report data. This case can happen when no asset was scanned at all.

It also clean up the test case and simplifies the test setup for junit report testing so that we explicitly can test 3 score results: pass, skip and error


vjeffrey (approved):
so many tests!!! ⭐️ 🎉

PR #913, opened 2023-10-30
🐛 Regularly check in with upstream

### LINKED ISSUE #894: `cnspec` does not report check-in to upstream as one might expect
**Describe the bug**
In the platform, we show a "Last Check-in" for managed clients. But this does not get updated when a scan happens or the service is restarted.

From the cnspec service logs:
```
Oct 26 06:38:56 ubuntu-jammy cnspec[8606]: → next scan in 1h0m40.42483421s
Oct 26 07:39:36 ubuntu-jammy cnspec[8606]: → discover related assets for 1 asset(s)
Oct 26 07:39:36 ubuntu-jammy cnspec[8606]: → synchronize assets
Oct 26 07:40:23 ubuntu-jammy crontab[9836]: (root) LIST (root)
Oct 26 07:40:26 ubuntu-jammy cnspec[8606]: → scan for asset ubuntu-jammy (Ubuntu 22.04.1 LTS) completed
Oct 26 07:40:26 ubuntu-jammy cnspec[8606]: → next scan in 1h20m15.867470213s
```

But in the UI:
image

**To Reproduce**

1. Create a Ubuntu VM
2. Create a workstation integration in the UI
3. Install the client in the VM as suggested by the UI
4. See the client report in the UI
5. Wait some time and see scans happening
6. Note that the UI does not show an updated "Last Check-in"

**Expected behavior**
I'd expect the client to check-in with a scan.

**Desktop (please complete the following information):**

- OS: Ubuntu 22.04 VM, but also reported on Windows
- cnspec 9.3.3

**Additional context**
I've not checked the code so far, whether we have an interval set or something like that.
Perhaps we automatically check-in every 24 hours?

czunker:
Fixes #894


czunker:
This requires https://github.com/mondoohq/cnquery/pull/2419 to work.

apps/cnspec/cmd/serve.go
```diff
@@ -83,6 +83,10 @@ var serveCmd = &cobra.Command{
 			hc := backgroundjob.NewHealthPinger(ctx, client.HttpClient, client.ApiEndpoint, 5*time.Minute)
 			hc.Start()
 			defer hc.Stop()
+
+			checkin := backgroundjob.NewCheckinPinger(ctx, client.HttpClient, client.ApiEndpoint, conf.runtime.UpstreamConfig, 2*time.Hour)
```
chris-rock, line 83: We do we need health and checkin?

czunker: The health check only checks the availability of the upstream API. No data is updated upstream, but it is also used for clock skew detection.

The check-in reports the current version and other data and this is updated upstream.

We could combine this into one API call. I separated it because of the clock skew detection and this way we can use different intervals.

chris-rock: I think we should keep the checkin call and not do the health check. I like the separation but I do not see any benefit of doing the health check in addition to the checkin.

czunker: :+1: Changed it.

apps/cnspec/cmd/backgroundjob/checkin.go
```diff
@@ -0,0 +1,154 @@
...
+		endpoint:   endpoint,
+		httpClient: httpClient,
+		mrn:        config.Creds.Mrn,
+		creds:      config.Creds,
+	}
+}
+
+func (c *checkinPinger) Start() {
+	// determine information about the client
+	sysInfo, err := sysinfo.GatherSystemInfo()
```
imilchev, line 47: IIRC this call is starting an OS provider. Please double-check that it is getting killed

apps/cnspec/cmd/backgroundjob/checkin.go
```diff
@@ -0,0 +1,154 @@
...
+	} else {
+		return errors.New("no credentials configured")
+	}
+
+	client, err := upstream.NewAgentManagerClient(c.endpoint, c.httpClient, plugins...)
+	if err != nil {
+		return errors.Wrap(err, "could not connect to mondoo platform")
+	}
+
+	name := viper.GetString("name")
```
imilchev: i think this isn't used anywhere at the moment

apps/cnspec/cmd/backgroundjob/checkin.go
czunker: Yes, this uses the os provider. And yes it gets killed. I checked with `ps`. As soon as the actual scan finishes, all started plugins are gone.

apps/cnspec/cmd/backgroundjob/checkin.go
czunker: Good catch. Removed it.

apps/cnspec/cmd/backgroundjob/checkin.go
imilchev: I think that's because there is a coordinator shutdown invoked after a scan is done. However, the scans run every ~1h and this is every ~2h + potential jitter. I guess that means we can have 1 OS provider running for a while until a scan happens

imilchev: We also have the provider heartbeat implemented though so I guess that fine. Just wanted to point out the behaviour here

PR #915, opened 2023-10-30
🧹 remove deprecated v7 structs

chris-rock:
This change removes all deprecated v7 structs. We do not need them in v9 anymore.


policy/testdata/policybundle-deps.mql.yaml
```diff
@@ -1,57 +1,48 @@
...
-  - asset_filter:
-      query: platform.name == "debian"
-    scoring_queries:
-      //policy.api.mondoo.app/queries/1.1.1.1-mounting-freevxfs-filesystems-disabled:
-        id: //policy.api.mondoo.app/queries/1.1.1.1-mounting-freevxfs-filesystems-disabled
-      //policy.api.mondoo.app/queries/1.1.1.2-mounting-jffs2-filesystems-disabled:
-        id: //policy.api.mondoo.app/queries/1.1.1.2-mounting-jffs2-filesystems-disabled
-      //policy.api.mondoo.app/queries/1.1.1.3-mounting-hfs-filesystems-disabled:
-        id: //policy.api.mondoo.app/queries/1.1.1.3-mounting-hfs-filesystems-disabled
+  - computed_filters: platform.name == "debian"
```
jaym: i feel like `computed_filters` was an internal thing and not to be set explicitly

chris-rock: I have not changed those values, I just reformatted it. It is a great call to remove those though.

PR #922, opened 2023-11-01
🧹 Include mrn in ErrVariantCycleDetected.

preslavgerchev:
Makes for looking through a problematic bundle easier


czunker (approved):
:+1: Thanks @preslavgerchev

PR #924, opened 2023-11-02
🐛 Use agent MRN to check in upstream

czunker:
Previously, I used the serivceaccount MRN. That does not work on all cases. We need to use the agent MRN.


czunker:
This is a fix to https://github.com/mondoohq/cnspec/pull/913

preslavgerchev (approved):

PR #926, opened 2023-11-02
🐛 Ensure that policy group checks are also properly cleaned if queries are missing

### LINKED ISSUE #925: asset filter in policy not working
i run a docker debian container for testing with cnspec version 9.5.1.

I do `cnspec bundle init` to create the example policy `example-policy.mql.yaml`

I execute the the scan with the example policy

```bash
cnspec scan -f example-policy.mql.yaml
→ no provider specified, defaulting to local. Use --help to see all providers.
→ no Mondoo configuration file provided, using defaults
! No credentials provided. Switching to --incognito mode.

 aebc0940f38d ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────    X score


Asset: aebc0940f38d
-------------------

error: rpc error: code = InvalidArgument desc = asset doesn't support any policies


Scanned 1 asset

Debian GNU/Linux 12 (bookworm)
    X aebc0940f38d
```

preslavgerchev:
We only were checking the `queries` part and not the `checks`, which mean that any policy grp w/o queries would have been skipped. Fixes #925 

- [x] I want to add a test for this

### DIFF
--- a/policy/bundle.go
+++ b/policy/bundle.go
@@ -579,7 +579,8 @@ func (c *bundleCache) removeFailing(res *Bundle) {
 		for j := range policy.Groups {
 			group := policy.Groups[j]
 			group.Queries = explorer.FilterQueryMRNs(c.removeQueries, group.Queries)
-			if len(group.Queries) != 0 {
+			group.Checks = explorer.FilterQueryMRNs(c.removeQueries, group.Checks)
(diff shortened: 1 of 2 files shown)


mariuskimmina (approved):
Love to see the added test :+1:

PR #935, opened 2023-11-06
✨ Add FromQueryPackBundle, allowing conversion of an explorer bundle to a policy bundle.

preslavgerchev:
This is more of an internal feature, but a function to let us merge a cnquery (explorer) bundle and a policy bundle together.

### DIFF
--- a/policy/bundle.go
+++ b/policy/bundle.go
@@ -120,7 +120,7 @@ func aggregateFilesToBundle(paths []string) (*Bundle, error) {
 			return nil, errors.Wrap(err, "could not load file: "+path)
 		}
 
-		mergedBundle = aggregateBundles(mergedBundle, bundle)
+		mergedBundle = Merge(mergedBundle, bundle)
 	}
 
 	return mergedBundle, nil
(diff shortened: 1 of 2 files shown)


policy/bundle.go
```diff
@@ -1106,3 +1106,18 @@ func translateGroupUIDs(ownerMrn string, policyObj *Policy, uid2mrn map[string]s
 
 	return nil
 }
+
+// Takes a query pack bundle and does the following:
+// 1. Merges all queries from the two bundles together
+// 2. Merges all props from the two bundles together
+// 3. Puts all query packs in the policy bundle
+// 4. Convert all query packs to policies with a data only scoring system. Packs are still kept
+func (p *Bundle) MergeQueryPackBundle(bundle *explorer.Bundle) {
```
arlimus: @preslavgerchev What if we changed this into a more generic `Merge(other *Bundle)` and instead have another `FromQueryPack(*explorer.Bundle)` function? It does require one more function, but they would be more specialized in their use-cases i.e. merge bundles together and convert any explorer.Bundle that we may receive. We currently have the `aggregateBundles` function in this file, which could be converted + exported into a `Merge(other *Bundle)` function. It also needs to combine the `Packs` field which is missing right now.

arlimus (changes requested):
Super useful addition! I have a suggestion that splits the new call slightly and replaces an old internal-only call

policy/bundle.go
preslavgerchev: Sounds good!
Then I assume we could chain this into something like `bundle.Merge(FromQueryPack(explorerBundle))` where needed. I will work on that

preslavgerchev:
@arlimus i've added the new `FromQueryPackBundle` func, this should be rfr again 🙌

arlimus (approved):
Awesome!! Thank you so much

PR #937, opened 2023-11-07
🧹 add CompilerConfig on all compile-related calls

arlimus:
See https://github.com/mondoohq/cnquery/pull/2510

@ Reviewers: I recommend merging this post-release, since it will require changes in all upstream projects that depend on it. This is the accompanying cnspec update, but there is at least one more place to update.


chris-rock (approved):
Thank you @arlimus

PR #941, opened 2023-11-08
Be more clear on the CLI that we're showing compliance controls

tas50:
Controls alone is too generic of a heading.


mariuskimmina (approved):
Sounds good to me

PR #943, opened 2023-11-09
✨ policy exceptions v2

imilchev:
We want to support a new way of defining exceptions for checks in policies. The new way is almost identical to compliance framework exceptions.

Example exception old:
```yaml
policies:
  - uid: sshd-server-policy
    name: SSH Server Policy
    version: 1.0.0
    groups:
      - filters: asset.family.contains("darwin")
        checks:
          - uid: sshd-score-01
          - uid: ignored-query
          - uid: deactivate-query
  - uid : asset-policy
    groups:
      - checks:
          - uid: ignored-query
            action: 4
          - uid: deactivate-query
            action: 2
        policies:
          - uid: sshd-server-policy
queries:
  - uid: sshd-score-01
    title: Ensure SSH MaxAuthTries is set to 4 or less
    mql: true == true
  - uid: ignored-query
    title: Ignored query
    mql: asset.name == "test"
  - uid: deactivate-query
    title: Deactivate query
    mql: asset.name == "test2"
```

New example:
```yaml
policies:
  - uid: sshd-server-policy
    name: SSH Server Policy
    version: 1.0.0
    groups:
      - filters: asset.family.contains("darwin")
        checks:
          - uid: sshd-score-01
          - uid: ignored-query
          - uid: deactivate-query
  - uid : asset-policy
    groups:   
      - checks:
          - uid: ignored-query
        type: 4
        policies:
          - uid: sshd-server-policy
      - checks:
          - uid: deactivate-query
        type: 5
        policies:
          - uid: sshd-server-policy
queries:
  - uid: sshd-score-01
    title: Ensure SSH MaxAuthTries is set to 4 or less
    mql: true == true
  - uid: ignored-query
    title: Ignored query
    mql: asset.name == "test"
  - uid: deactivate-query
    title: Deactivate query
    mql: asset.name == "test2"
```


policy/resolver.go
```diff
@@ -977,13 +977,12 @@ func (cache *policyResolverCache) addCheckJob(ctx context.Context, check *explor
 		cache.global.reportingJobsByUUID[uuid] = queryJob
 		cache.global.reportingJobsByMsum[check.Checksum] = append(cache.global.reportingJobsByMsum[check.Checksum], queryJob)
 		cache.childJobsByMrn[check.Mrn] = append(cache.childJobsByMrn[check.Mrn], queryJob)
+		ownerJob.ChildJobs[queryJob.Uuid] = impact
 	}
 
 	// local aspects for the resolved policy
 	queryJob.Notify = append(queryJob.Notify, ownerJob.Uuid)
 
-	ownerJob.ChildJobs[queryJob.Uuid] = impact
```
imilchev, line 985: I don't understand why we need to change the impact for a job that already exists... I think this shouldn't be the case. If we really want to change it, then we should be calling `modifyCheckJob` instead of `addCheckJob` but I am not sure

arlimus: Good catch! There is a problem that we never fully explored that you are really exposing here:
What if a check is deactivate, and later re-activate? 
ie: it is first added (action=unspecified/activate)
then it is deactivated (action=deactivate)
and then another top-level policy comes and wants to actually set it on AND/OR make a modification. We have assumed - so far - that the action for activate and the action for modify are almost interchangeable. However, now that the impact is getting modified, it can lead to some nasty side-effects. This happens because if we ignore a check, the impact is set to `impact.Scoring = explorer.ScoringSystem_IGNORE_SCORE`. So any call later on that modifies this overwrites the impact's scoring system.

Let's solve this step by step.
And the first step is where I think you are spot on: let's remove this line and keep it only for explicit modifications.

Next, I think we should (1) remove the really outdated `Action` field from impact (it should only be used for v7 compatibility, otherwise all actions are bound to the mquery) and (2) review the action field, because it is overloaded right now.

imilchev: added #949 and #950 for the follow ups. Also added automated tests for the cases used to validate whether exceptions work as expected

policy/scan/local_scanner_test.go
```diff
@@ -100,3 +105,224 @@ func TestCreateAssetList(t *testing.T) {
...
+		RemoveFailing: true,
+	})
+	s.Require().NoError(err)
+
+	s.job.Bundle = bundle
+	s.job.PolicyFilters = []string{"asset-policy"}
+	bundleMap := bundle.ToMap()
+
+	ctx := context.Background()
+	scanner := NewLocalScanner(WithRecording(providers.NullRecording{}))
```
czunker: nit: `WithRecording(providers.NullRecording{})` should no longer be necessary. That's the default now.

policy/resolver.go
```diff
@@ -904,20 +904,20 @@ func (s *LocalServices) policyGroupToJobs(ctx context.Context, group *PolicyGrou
...
-		if check.Action == explorer.Action_IGNORE {
+		if check.Action == explorer.Action_IGNORE || group.Type == GroupType_IGNORED {
 			impact.Scoring = explorer.ScoringSystem_IGNORE_SCORE
 			impact.Action = explorer.Action_IGNORE
 		}
 
 		cache.global.propsCache.Add(check.Props...)
 
-		if check.Action == explorer.Action_UNSPECIFIED || check.Action == explorer.Action_ACTIVATE {
+		if (check.Action == explorer.Action_UNSPECIFIED || check.Action == explorer.Action_ACTIVATE) && group.Type != GroupType_IGNORED {
```
czunker, line 914: I'm not 100% sure about this.

I think, for security exceptions, "IGNORE" means, that it still get's executed, but not scored. I'd expected "DISABLE" here.

policy/scan/local_scanner_test.go
```diff
@@ -100,3 +105,224 @@ func TestCreateAssetList(t *testing.T) {
...
+		p := full.ResolvedPolicies[k]
+
+		// Get the code id for all the executed queries
+		executedQueries := []string{}
+		for qCodeId := range p.ExecutionJob.Queries {
+			executedQueries = append(executedQueries, qCodeId)
+		}
+
+		expectedQueries := []string{
+			bundleMap.Queries["//local.cnspec.io/run/local-execution/queries/ignored-query"].CodeId,
```
czunker, line 251: I think, I'm missing something here.

Regarding the comment above about "IGNORE"/"DISABLE":
"ignored query" has an unspecified action, but is part of an ignored group. With the above code I'd not expected it to be included.

czunker (commented):
Thanks for all the tests @imilchev 

Code is looking good to me. There is only one thing I currently do not understand. I'll try to test this manually.

czunker:
I caused a panic during testing:
```
cnspec scan local -f policy/scan/testdata/exceptions.mql.yaml 
→ loaded configuration from /etc/opt/mondoo/mondoo.yml using source default
→ using service account credentials
! Scanning with local bundles will switch into --incognito mode by default. Your results will not be sent upstream.
panic: invalid score report

goroutine 94 [running]:
go.mondoo.com/cnspec/v9/policy/executor/internal.(*ReportingJobNodeData).consume(0xc0001a5540, {0xc000014c00, 0xc}, 0xc001467d40)
	/home/christian/workspace/mondoo/github.com/cnspec/policy/executor/internal/nodes.go:442 +0x19a
go.mondoo.com/cnspec/v9/policy/executor/internal.(*GraphExecutor).Execute(0xc001493680)
	/home/christian/workspace/mondoo/github.com/cnspec/policy/executor/internal/graph.go:137 +0x832
go.mondoo.com/cnspec/v9/policy/executor.ExecuteResolvedPolicy({0x17e3e80, 0xc000fde1c0}, {0x17e6de0?, 0xc000eb11a0}, {0xc0004c9700, 0x3a}, 0xc000bcfc40?, {0xc001004408, 0x2, 0x8}, ...)
	/home/christian/workspace/mondoo/github.com/cnspec/policy/executor/graph.go:47 +0x479
go.mondoo.com/cnspec/v9/policy/scan.(*localAssetScanner).runPolicy(0xc001387d48)
	/home/christian/workspace/mondoo/github.com/cnspec/policy/scan/local_scanner.go:996 +0x509
go.mondoo.com/cnspec/v9/policy/scan.(*localAssetScanner).run(0xc001387d48)
	/home/christian/workspace/mondoo/github.com/cnspec/policy/scan/local_scanner.go:777 +0x33
go.mondoo.com/cnspec/v9/policy/scan.(*LocalScanner).runMotorizedAsset.func1(0x17e3e80?, 0xc000a3a1c0?)
	/home/christian/workspace/mondoo/github.com/cnspec/policy/scan/local_scanner.go:628 +0x217
go.mondoo.com/cnspec/v9/internal/datalakes/inmemory.WithDb({0x17e3e80?, 0xc000a3a1c0?}, 0x7fdfc97d7a68?, 0xc000bcfdf8)
	/home/christian/workspace/mondoo/github.com/cnspec/internal/datalakes/inmemory/inmemory.go:52 +0x42
go.mondoo.com/cnspec/v9/policy/scan.(*LocalScanner).runMotorizedAsset(0xc00100e0f0?, 0x15d0478?)
	/home/christian/workspace/mondoo/github.com/cnspec/policy/scan/local_scanner.go:604 +0x75
go.mondoo.com/cnspec/v9/policy/scan.(*LocalScanner).RunAssetJob(0xc0004c6850, 0xc000fe4c80)
	/home/christian/workspace/mondoo/github.com/cnspec/policy/scan/local_scanner.go:568 +0x3fb
go.mondoo.com/cnspec/v9/policy/scan.(*LocalScanner).distributeJob.func2()
	/home/christian/workspace/mondoo/github.com/cnspec/policy/scan/local_scanner.go:492 +0x105
created by go.mondoo.com/cnspec/v9/policy/scan.(*LocalScanner).distributeJob in goroutine 1
	/home/christian/workspace/mondoo/github.com/cnspec/policy/scan/local_scanner.go:472 +0x1749
```
I only removed the `action` field from the `ignored-query` in above policy.

The panic does not happen with cnspec v9.7.0. So, my guess is this PR introduced/triggered it.

The panic also happens when I only change the `action` field from 4 to 0.

czunker:
> Thanks for all the tests @imilchev
> 
> Code is looking good to me. There is only one thing I currently do not understand. I'll try to test this manually.

I tried policy groups with "IGNORED" and "DISABLED", that changes the score as expected. But, from the code I don't understand why this happens.

policy/scan/local_scanner_test.go
imilchev: it doesn't seem to be the default. I added this because if it's not there, then I get panic on recording stuff

policy/resolver.go
imilchev: this check makes sure that when we have a group with action `IGNORE` we do not call `addCheckJob` but `modifyCheckJob`. The exception groups do not add new checks, they only modify existing ones

policy/scan/local_scanner_test.go
imilchev: yeah, the resolver code is complex. That's why I added tests for the old exception scenarios and the new ones. If you revert my changes to the resolver, the old test cases still pass. So that is the expected behaviour

policy/scan/local_scanner_test.go
czunker: Nice catch! I only fixed it for cnquery: https://github.com/mondoohq/cnquery/pull/2332

I'll add it to cnspec as well.

policy/resolver.go
czunker: Thanks for the explanation. Makes sense to me.

policy/scan/local_scanner_test.go
czunker: Thanks for the explanation. Makes sense to me.

czunker:
> I caused a panic during testing:
> 
> ```
> cnspec scan local -f policy/scan/testdata/exceptions.mql.yaml 
> → loaded configuration from /etc/opt/mondoo/mondoo.yml using source default
> → using service account credentials
> ! Scanning with local bundles will switch into --incognito mode by default. Your results will not be sent upstream.
> → discover related assets for 1 asset(s)
> panic: invalid score report
> 
> goroutine 94 [running]:
> go.mondoo.com/cnspec/v9/policy/executor/internal.(*ReportingJobNodeData).consume(0xc0001a5540, {0xc000014c00, 0xc}, 0xc001467d40)
> 	/home/christian/workspace/mondoo/github.com/cnspec/policy/executor/internal/nodes.go:442 +0x19a
> go.mondoo.com/cnspec/v9/policy/executor/internal.(*GraphExecutor).Execute(0xc001493680)
> 	/home/christian/workspace/mondoo/github.com/cnspec/policy/executor/internal/graph.go:137 +0x832
> go.mondoo.com/cnspec/v9/policy/executor.ExecuteResolvedPolicy({0x17e3e80, 0xc000fde1c0}, {0x17e6de0?, 0xc000eb11a0}, {0xc0004c9700, 0x3a}, 0xc000bcfc40?, {0xc001004408, 0x2, 0x8}, ...)
> 	/home/christian/workspace/mondoo/github.com/cnspec/policy/executor/graph.go:47 +0x479
> go.mondoo.com/cnspec/v9/policy/scan.(*localAssetScanner).runPolicy(0xc001387d48)
> 	/home/christian/workspace/mondoo/github.com/cnspec/policy/scan/local_scanner.go:996 +0x509
> go.mondoo.com/cnspec/v9/policy/scan.(*localAssetScanner).run(0xc001387d48)
> 	/home/christian/workspace/mondoo/github.com/cnspec/policy/scan/local_scanner.go:777 +0x33
> go.mondoo.com/cnspec/v9/policy/scan.(*LocalScanner).runMotorizedAsset.func1(0x17e3e80?, 0xc000a3a1c0?)
> 	/home/christian/workspace/mondoo/github.com/cnspec/policy/scan/local_scanner.go:628 +0x217
> go.mondoo.com/cnspec/v9/internal/datalakes/inmemory.WithDb({0x17e3e80?, 0xc000a3a1c0?}, 0x7fdfc97d7a68?, 0xc000bcfdf8)
> 	/home/christian/workspace/mondoo/github.com/cnspec/internal/datalakes/inmemory/inmemory.go:52 +0x42
> go.mondoo.com/cnspec/v9/policy/scan.(*LocalScanner).runMotorizedAsset(0xc00100e0f0?, 0x15d0478?)
> 	/home/christian/workspace/mondoo/github.com/cnspec/policy/scan/local_scanner.go:604 +0x75
> go.mondoo.com/cnspec/v9/policy/scan.(*LocalScanner).RunAssetJob(0xc0004c6850, 0xc000fe4c80)
> 	/home/christian/workspace/mondoo/github.com/cnspec/policy/scan/local_scanner.go:568 +0x3fb
> go.mondoo.com/cnspec/v9/policy/scan.(*LocalScanner).distributeJob.func2()
> 	/home/christian/workspace/mondoo/github.com/cnspec/policy/scan/local_scanner.go:492 +0x105
> created by go.mondoo.com/cnspec/v9/policy/scan.(*LocalScanner).distributeJob in goroutine 1
> 	/home/christian/workspace/mondoo/github.com/cnspec/policy/scan/local_scanner.go:472 +0x1749
> ```
> 
> I only removed the `action` field from the `ignored-query` in above policy.
> 
> The panic does not happen with cnspec v9.7.0. So, my guess is this PR introduced/triggered it.
> 
> The panic also happens when I only change the `action` field from 4 to 0.

Tried to reproduce this today, but it is gone.

czunker (approved):
Thanks @imilchev

PR #947, opened 2023-11-14
🐛 filter policies without removing them from the bundle

arlimus:
Removing policies from the bundle on filter calls actually caused havoc because it removed potentially dependent policies from the bundle. For example: if you bundle has policy A + B, and B depends on A, and you set `--policy` to `B`, then A gets removed from the bundle and leads to errors.

We only need to remove the policy from the list of executed policies in the resolver, not from the bundle itself.


imilchev (approved):
Tested this! Works great! Thanks a lot 🎉

PR #951, opened 2023-11-16
⭐️ replace dque with pdque

mariuskimmina:
Fixes https://github.com/mondoohq/mondoo-operator/issues/852

### DIFF
--- a/policy/scan/disk_queue_test.go
+++ b/policy/scan/disk_queue_test.go
@@ -11,13 +11,6 @@ import (
 	"go.mondoo.com/cnquery/v9/providers-sdk/v1/inventory"
 )
 
-var testQueueConfig = diskQueueConfig{
-	dir:         "testdir", // TODO: consider configurable path
-	filename:    "disk-queue",
-	segmentSize: 500,
-	sync:        false,
-}
-
 func TestDiskQueueClient_EnqueueDequeue(t *testing.T) {
 	tempDir, err := os.MkdirTemp("", "testdir")
 	if err != nil {
--- a/policy/scan/disk_queue_test.go
+++ b/policy/scan/disk_queue_test.go
@@ -1,3 +1,6 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
 package scan
 
 import (
--- a/policy/scan/disk_queue_test.go
+++ b/policy/scan/disk_queue_test.go
@@ -25,7 +25,6 @@ func TestDiskQueueClient_EnqueueDequeue(t *testing.T) {
 
 	handler := func(job *Job) {
 		completionChannel <- struct{}{} // Signal completion
-		// Perform additional checks on job if necessary
 	}
 
 	client, err := newDqueClient(testConfig, handler)
--- a/policy/scan/disk_queue_test.go
+++ b/policy/scan/disk_queue_test.go
@@ -6,7 +6,6 @@ package scan
 import (
 	"os"
 	"testing"
-	"time"
 
 	"go.mondoo.com/cnquery/v9/providers-sdk/v1/inventory"
 )
@@ -22,9 +21,10 @@ func TestDiskQueueClient_EnqueueDequeue(t *testing.T) {
 	testConfig := defaultDqueConfig
 	testConfig.dir = tempDir
 
-	handlerCalled := false
+	completionChannel := make(chan struct{}, 50) // Channel to signal job completion
+
 	handler := func(job *Job) {
-		handlerCalled = true
+		completionChannel <- struct{}{} // Signal completion
 		// Perform additional checks on job if necessary
 	}
 
@@ -57,13 +57,16 @@ func TestDiskQueueClient_EnqueueDequeue(t *testing.T) {
 			},
 		},
 	}
-	for i := 1; i < 50; i++ {
+	for i := 0; i < 50; i++ {
 		client.Channel() <- *testJob
 	}
-	// Allow some time to process
-	time.Sleep(2 * time.Second)
 
-	if !handlerCalled {
-		t.Errorf("Expected handler to be called after dequeue")
+	for i := 0; i < 50; i++ {
+		<-completionChannel
+	}
+
+	// Verify that all jobs have been processed
+	if len(completionChannel) != 0 {
+		t.Errorf("Expected handler to be called 50 times, but was called %d times", 50-len(completionChannel))
 	}
 }
--- a/policy/scan/disk_queue_test.go
+++ b/policy/scan/disk_queue_test.go
@@ -1,0 +1,90 @@
+package scan
+
+import (
+	"os"
+	"testing"
+	"time"
+
+	"go.mondoo.com/cnquery/v9/providers-sdk/v1/inventory"
+)
+
+var testQueueConfig = diskQueueConfig{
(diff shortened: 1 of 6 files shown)


policy/scan/disk_queue_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"os"
+	"testing"
+	"time"
+
+	"go.mondoo.com/cnquery/v9/providers-sdk/v1/inventory"
+)
+
+func TestDiskQueueClient_EnqueueDequeue(t *testing.T) {
```
imilchev, line 13: can you also add a test that starts 2 goroutines that are both enqueuing jobs and have 1 go routine processing? At the end of the test make sure you have processed the right amount of jobs and that there are no files remaining on the disk

policy/scan/disk_queue_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	client, err := newDqueClient(testConfig, handler)
+	if err != nil {
+		t.Fatalf("Failed to create diskQueueClient: %v", err)
+	}
+	defer client.Stop()
+
+	// Initially, the directory should be empty
+	initialFileCount, err := countFilesInDir(tempDir)
+	if err != nil {
```
imilchev: you shouldn't do such checks. Just use `require.NoError(t, err)`

policy/scan/disk_queue_test.go
```diff
@@ -0,0 +1,86 @@
...
+								},
+							},
+						},
+						ManagedBy: "mondoo-operator-123",
+					},
+				},
+			},
+		},
+	}
+	for i := 1; i < 50; i++ {
```
imilchev: somewhere you should verify that 50 jobs have been processed. Ideally, your handler should just increment a counter and have some ms of sleep. At the end of the execution you should verify that your counter is equal to the amount of jobs you have created

policy/scan/pdque/queue_test.go
```diff
@@ -0,0 +1,407 @@
...
+	testDir, err := os.MkdirTemp("", "test_enqueue_dequeue")
+	if err != nil {
+		t.Fatalf("Failed to create temp dir: %v", err)
+	}
+	defer os.RemoveAll(testDir)
+	// Setup: Initialize the queue
+	queue, err := NewOrOpen("testQueue", testDir, 10, func() interface{} {
+		return new(testObj)
+	})
+	if err != nil {
```
imilchev: rework this checks with `require.NoError(err)`. Makes the code much shorter and easier to read

policy/scan/pdque/queue_test.go
```diff
@@ -0,0 +1,407 @@
...
+		t.Fatalf("Failed to create queue: %v", err)
+	}
+	defer q.Close()
+
+	var wg sync.WaitGroup
+
+	// Concurrently enqueue jobs
+	wg.Add(1)
+	go func() {
```
imilchev, line 316: run 2 of these in parallel.

policy/scan/disk_queue_test.go
imilchev: you can do that test in the pdqueue tests

policy/scan/pdque/queue.go
```diff
@@ -0,0 +1,414 @@
...
+			}
+
+			return obj, nil
+		}
+	}
+
+	return nil, errors.New("no jobs in queue")
+}
+
+//func (q *Queue) DequeueBlock() (interface{}, error) {
```
arlimus: Are we keeping this around to enable it shortly after merging, or is this a bit of a leftover?
If it's something that comes in a quick follow-up, I'd be ok to leave it (as an exception). Otherwise, generally, it'd be great to keep the code clean :)
(I suspect it's by error since the same method name is defined below)

policy/scan/pdque/queue.go
```diff
@@ -0,0 +1,414 @@
...
+		path:    path,
+		builder: builder,
+		maxSize: maxSize,
+	}
+	que.cond = sync.NewCond(&que.mu)
+
+	return que, nil
+}
+
+func NewOrOpen(name string, path string, maxSize int, builder func() interface{}) (*Queue, error) {
```
arlimus, line 76: semantic thingy: The difference between `New` and `NewOrOpen` when the folder does not exist is that `NewOrOpen` will create the folder, while `New` will not. This feels unintuitive, since in the case where the folder doesn't exist both should behave similarly. 

As a possible suggestion if that makes sense, either just rename it to `CreateOrOpen` (making it different from `New` semantically) or implement the folder-check+create in `New` too.

policy/scan/pdque/queue.go
```diff
@@ -0,0 +1,414 @@
...
+
+			return obj, nil
+		}
+	}
+
+	return nil, ErrNoJobs // Custom error to indicate no jobs are available
+}
+
+// ErrNoJobs is a custom error to indicate that no jobs are available in the queue.
+var ErrNoJobs = errors.New("no jobs in queue")
```
arlimus: nit: move the error up to be with its friends (the other error definitions)

policy/scan/pdque/queue.go
```diff
@@ -0,0 +1,414 @@
...
+			timestamp++
+			filename = strconv.FormatInt(timestamp, 10)
+		}
+	}
+	return filename, nil
+}
+
+// check the currentSize of the queue
+// We do a lot of disk operations here, could find a more performant approach
+func (q *Queue) currentSize() (int, error) {
```
arlimus, line 333: this is just for a follow-up: 
Do we expect multiple go-processes to run jobs on the same queue-folder? If so, then yes, we'd have to check the folder contents on every call. If not, we could check the contents once at startup and then just keep track of the internal # of jobs. (my general recommendation is to avoid running parallel processes, but I'd love to learn more!)

arlimus (changes requested):
A few small things remaining, very close!

policy/scan/pdque/queue.go
arlimus: Also: there is an argument for blocking the queue while counting on disk (because jobs may be added in the meantime), but in the multi-process use-case it is almost impossible to safeguard since any other process will likely not block and wait

policy/scan/pdque/queue.go
mariuskimmina: Oh, yeah `New` should of course create the directory, not just `NewOrOpen`

policy/scan/pdque/queue.go
mariuskimmina: I am not sure if we do multiple processes in production but the queue should be able to handle it if so, there is a test for it too. The `currentSize` function does not have a lock but it's only called by `Enqueue` which does have a lock - so I think we should be safe here. Question remains if the performance of this will eventually become a bottleneck or not.

imilchev: the current library makes sure you cannot start multiple processes with the same disk queue (folder)

imilchev (approved):

policy/scan/pdque/queue.go
arlimus: Awesome, thank you!!

arlimus (approved):
Thank you for the updates + implementation @mariuskimmina  !!
kudos for the review @imilchev :grin:

PR #952, opened 2023-11-16
🐛 Try to make mquery checksums stable

jaym:
- We cannot change a value of the mquery before its stored. This causes the checksum to change. Its better we do this before we calculate checksums. I've made this happen when the bundle is compiled
- The queries variants refer to need to have their checksums before we checksum it. So we need to topologically sort the queries before compiling such that all dependent queries have been compiled and checksummed before compiling the depending query.

### DIFF
--- a/policy/bundle.go
+++ b/policy/bundle.go
@@ -657,6 +657,58 @@ func detectVariantCyclesDFS(mrn string, statusMap map[string]nodeVisitStatus, qu
 	return nil
 }
 
+func topologicalSortQueries(queries []*explorer.Mquery) ([]*explorer.Mquery, error) {
+	// Gather all top-level queries with variants
+	queriesMap := map[string]*explorer.Mquery{}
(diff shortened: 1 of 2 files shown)


policy/bundle.go
```diff
@@ -966,6 +1028,10 @@ func (c *bundleCache) precompileQuery(query *explorer.Mquery, policy *Policy) *e
 		return nil
 	}
 
+	if query.Title == "" {
```
preslavgerchev, line 1031: this happens after the compile call, is this intended?

jaym: I think this is happening before compile as the function is `precompileQuery`.

preslavgerchev: given we removed this in cnquery, can we do the same here?

jaym: Can you explain further. I still see a precompileQuery in cnquery

PR #953, opened 2023-11-17
🧹 Prevent nil panic when recording is not set

czunker:
In case recording isn't specififed during scanner initialisation, set the NullRecording


czunker:
Same as for cnquery https://github.com/mondoohq/cnquery/pull/2332

imilchev (approved):

PR #955, opened 2023-11-18
⭐️ preserve head comments

chris-rock:
Before this change the `cnspec bundle fmt` format removed head comments. Sometimes its nice to be able to attach comments and preserve it. Its quite tricky with the yaml library. We work around this by overwriting the Unmarshal and Marshal function. We do not preserve all comments yet since there is way more fine-tuning required to make this work.

The following is working now:

```yaml
# test
policies:
    - uid: sample-policy
      name: Sample Policy
      version: 1.0.0
 ```

### DIFF
--- a/internal/bundle/bundle.yac.go
+++ b/internal/bundle/bundle.yac.go
@@ -18,6 +18,12 @@ type FileContext struct {
 	Column int
 }
 
+type Comments struct {
+	HeadComment string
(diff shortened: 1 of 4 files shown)


benr (approved):

benr:
Nice, thank you @chris-rock !!!

PR #962, opened 2023-11-22
✨Add sort flag to bundle fmt. Add SortContentsV2.

preslavgerchev:
Add a `--sort` flag to the `bundle fmt` cmd. It's optional and if passed will use the sort the selected fields.

Add `SortContentsV2` which makes bundle sorting a bit more extensible + a test

### DIFF
--- a/apps/cnspec/cmd/bundle.go
+++ b/apps/cnspec/cmd/bundle.go
@@ -30,6 +30,7 @@ func init() {
 	policyBundlesCmd.AddCommand(policyLintCmd)
 
 	// fmt
+	policyFmtCmd.Flags().StringArray("sort", []string{}, "sort the selected fields in the bundle.")
 	policyBundlesCmd.AddCommand(policyFmtCmd)
 
 	// docs
@@ -146,14 +147,14 @@ var policyFmtCmd = &cobra.Command{
 	Short:   "Apply style formatting to one or more policy bundles.",
 	Args:    cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
+		sort, _ := cmd.Flags().GetStringArray("sort")
 		ensureProviders()
 		for _, path := range args {
-			err := bundle.FormatRecursive(path)
+			err := bundle.FormatRecursive(path, sort)
 			if err != nil {
 				fmt.Println(err)
 				os.Exit(1)
 			}
-
 		}
 		log.Info().Msg("completed formatting policy bundle(s)")
 	},
--- a/apps/cnspec/cmd/bundle.go
+++ b/apps/cnspec/cmd/bundle.go
@@ -30,7 +30,7 @@ func init() {
 	policyBundlesCmd.AddCommand(policyLintCmd)
 
 	// fmt
-	policyFmtCmd.Flags().StringSlice("sort", []string{}, "sort the selected fields in the bundle.")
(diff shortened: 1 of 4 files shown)


apps/cnspec/cmd/bundle.go
```diff
@@ -30,6 +30,7 @@ func init() {
 	policyBundlesCmd.AddCommand(policyLintCmd)
 
 	// fmt
+	policyFmtCmd.Flags().StringSlice("sort", []string{}, "sort the selected fields in the bundle.")
```
chris-rock: If I read this right, I cannot just sort with `--sort` but need to understand what I want to sort?

apps/cnspec/cmd/bundle.go
```diff
@@ -146,14 +147,14 @@ var policyFmtCmd = &cobra.Command{
 	Short:   "Apply style formatting to one or more policy bundles.",
 	Args:    cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
+		sortFields, _ := cmd.Flags().GetStringSlice("sort")
 		ensureProviders()
 		for _, path := range args {
-			err := bundle.FormatRecursive(path)
+			err := bundle.FormatRecursive(path, sortFields)
```
chris-rock: Do we really need more than a boolean indicator for sorting?

internal/bundle/fmt.go
```diff
@@ -68,13 +68,18 @@ func sanitizeStringForYaml(s string) string {
 	return strings.Join(lines, "\n")
 }
 
-func FormatFile(filename string) error {
+func FormatAndSortFile(filename string, sortFields []string) error {
```
chris-rock: I would keep the name simple, the method signature changes anyway. `Format(filename string, sort bool)` sounds like it could do the trick

internal/bundle/fmt.go
```diff
@@ -127,3 +132,31 @@ func FormatBundleData(data []byte) ([]byte, error) {
 
 	return Format(b)
 }
+
+func SortBundleData(data []byte, sortFields []string) ([]byte, error) {
```
chris-rock: I would just call `SortBundle` and only accept bundle as input. I would sort by all parts automatically. The loading should not be done inside of the function. It would allow us to reuse the function way more often. So if we always sort by all in SortContentV2 and do not need the config setup. Wouldn't this make the whole implementation easier?

policy/bundle.go
```diff
@@ -424,15 +424,102 @@ func (p *Bundle) PolicyMRNs() []string {
...
-		return p.Policies[i].Mrn < p.Policies[j].Mrn
-	})
+type SortConfig struct {
+	SortQueries  *SortField
+	SortPolicies *SortField
+	SortVariants *SortField
+}
+
+type SortField struct {
+	Uid bool
```
chris-rock: Can we not use do this automatic, since MRN includes the uids as well, I am not sure why we need both.

chris-rock (commented):
Super great @preslavgerchev I added some comments for improvements. Happy to discuss if you think I missed anything.

policy/bundle.go
preslavgerchev: I added a check to verify if mrn isnt empty. If we're formatting a bundle that has no mrns, we need to fallback to uids

chris-rock (approved):
Great simplification

PR #966, opened 2023-11-28
🧹 Add sortContents to yac bundle struct. Simplify formatting.

preslavgerchev:
The yac bundle holds extra properties (file context, comments). To avoid losing those we add a `SortContents` function to the yac `Bundle` struct


chris-rock (approved):
Great improvement @preslavgerchev

PR #973, opened 2023-11-30
🐛 fix rejected exceptions execution

imilchev:
Rejected exceptions shouldn't count as exceptions. That bit was missing from the original implementation, so I added it here


mariuskimmina (approved):
Great to see more tests!

PR #982, opened 2023-12-06
⭐ Add output handler to store asset reports

preslavgerchev:
Add a new `--output-target` flag that specifies a target where the scan report can be stored.
This currently only supports local files and AWS SQS urls. If no target is specified, it falls back to CLI reporting (which is the current default behaviour) 

```
# for AWS sqs
~/go/bin/cnspec scan --output-target https://sqs.us-east-1.amazonaws.com/921877552404/myqueue --output json
# for local file
~/go/bin/cnspec scan --output-target file://somedir/file
~/go/bin/cnspec scan --output-target report.json --output json
```

for AWS SQS reporting  only `--output yaml` and `--output json` are supported

### DIFF
--- a/apps/cnspec/cmd/scan.go
+++ b/apps/cnspec/cmd/scan.go
@@ -143,12 +143,12 @@ var scanCmdRun = func(cmd *cobra.Command, runtime *providers.Runtime, cliRes *pl
 
 	logger.DebugDumpJSON("report", report)
 
-	cliConf := reporter.ReportConfig{
+	handlerConf := reporter.HandlerConfig{
 		Format:       conf.OutputFormat,
 		OutputTarget: conf.OutputTarget,
 		Incognito:    conf.IsIncognito,
 	}
-	outputHandler, err := reporter.NewOutputHandler(cliConf)
+	outputHandler, err := reporter.NewOutputHandler(handlerConf)
 	if err != nil {
 		log.Fatal().Err(err).Msg("failed to create an output handler")
 	}
--- a/apps/cnspec/cmd/scan.go
+++ b/apps/cnspec/cmd/scan.go
@@ -63,6 +63,7 @@ func init() {
 	scanCmd.Flags().MarkHidden("category")
 	scanCmd.Flags().Int("score-threshold", 0, "If any score falls below the threshold, exit 1.")
 	scanCmd.Flags().Bool("share", false, "create a web-based private reports when cnspec is unauthenticated. Defaults to false.")
+	scanCmd.Flags().String("output-target", "", "Set output target to which the asset report will be sent. Currently only supports AWS SQS topic URLs and local files")
 }
 
 var scanCmd = &cobra.Command{
@@ -79,7 +80,7 @@ To manually configure a policy, use this:
 		$ cnspec scan local -f bundle.mql.yaml --incognito
 
 `,
-	PreRun: func(cmd *cobra.Command, args []string) {
+	PreRunE: func(cmd *cobra.Command, args []string) error {
 		// Special handling for users that want to see what output options are
 		// available. We have to do this before printing the help because we
 		// don't have a target connection or provider.
@@ -109,6 +110,11 @@ To manually configure a policy, use this:
 		viper.BindPFlag("record", cmd.Flags().Lookup("record"))
 
 		viper.BindPFlag("output", cmd.Flags().Lookup("output"))
+		if err := viper.BindPFlag("output-target", cmd.Flags().Lookup("output-target")); err != nil {
+			return err
+		}
+
+		return nil
 	},
 	ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		if len(args) == 0 {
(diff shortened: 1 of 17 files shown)


apps/cnspec/cmd/scan.go
```diff
@@ -63,6 +63,7 @@ func init() {
 	scanCmd.Flags().MarkHidden("category")
 	scanCmd.Flags().Int("score-threshold", 0, "If any score falls below the threshold, exit 1.")
 	scanCmd.Flags().Bool("share", false, "create a web-based private reports when cnspec is unauthenticated. Defaults to false.")
+	scanCmd.Flags().String("report-dest-url", "", "Set the URL to send the asset report to. Currently only supports AWS SQS topic URLs.")
```
chris-rock: I think we should stick with `output-file` as we use it for liniting. Output file can have multiple formats:

```
--output-file /path/local
--output-file file:///path/local
--output-file awssqs://sqs.us-east-2.amazonaws.com/123456789012/myqueue?region=us-east-2
```

chris-rock: If we do not support local files out of the box, we should mark the flag hidden until we support at least local.

preslavgerchev: What about `output-target`? I find it weird to pass an SQS url to something called `output-file`

Also, a normal SQS url looks like this:
```
https://sqs.us-east-1.amazonaws.com/177715257436/MyQueue
```
I think we should stick that, instead of asking users to provide it with the `awssqs://` prefix (that seems to be specific to the gocloud dev lib we use). If we want that, we should make it clear in the doc with an example

cli/reporter/output_handler_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	for i, sqsUrl := range invalidSqsUrls {
+		_, err := NewOutputHandler(sqsUrl, "JSON")
+		require.Error(t, err, i)
+	}
+}
+
+func TestOutputHandlerLocal(t *testing.T) {
+	validFileTarget := []string{
+		"file:///root/test",
```
chris-rock, line 31: Do you think ` --output json --output-target report.json` should not work?

chris-rock: I even used `--output json --output-target file://report.json` and it has not stored the report in the current directory.

chris-rock (commented):
Very nice structured implementation. I am just having trouble to get the local file reporter running.

cli/reporter/output_handler_test.go
preslavgerchev: I just tested it, work fine for me
```
~/go/bin/cnspec scan --output json --output-target file://report.json
....
→ wrote report to file file=report.json
```

Are you seeing any errors?

preslavgerchev: > Do you think ` --output json --output-target report.json` should not work?

It needs to be prefixed with `file://` to be explicit. 

We could also remove the prefix requirement for now as we only have 2 targets and the SQS url is easily recognizable. 

But in the future it might be a problem with determining the target if we pass something like `mygcpproject/project/queue`, the handler might not be able to determine if this is a file or a GCP queue (I am just imagining here, don't know if this is actually a valid GCP queue URL)

apps/cnspec/cmd/scan.go
```diff
@@ -63,6 +63,7 @@ func init() {
 	scanCmd.Flags().MarkHidden("category")
 	scanCmd.Flags().Int("score-threshold", 0, "If any score falls below the threshold, exit 1.")
 	scanCmd.Flags().Bool("share", false, "create a web-based private reports when cnspec is unauthenticated. Defaults to false.")
+	scanCmd.Flags().String("output-target", "", "Set output target to which the asset report will be sent. Currently only supports AWS SQS topic URLs and local files")
```
chris-rock, line 66: Until we support more native os paths file `--output-file report.json` we should mark this flag as hidden. Its not common to require a `file://` prefix for a local file.

cli/reporter/output_handler_test.go
chris-rock: No error, it just prints the json output on cli

go.mod
```diff
@@ -6,7 +6,10 @@ go 1.21
 
 toolchain go1.21.3
 
-require go.mondoo.com/cnquery/v9 v9.10.0
+require (
+	go.mondoo.com/cnquery/v9 v9.10.0
+	gocloud.dev v0.34.0
```
chris-rock: nit, I think we want to keep the first require just for cnquery

test/cli/testdata/cnspec_scan.ct
```diff
@@ -26,6 +26,7 @@ Flags:
       --inventory-file string       Set the path to the inventory file.
   -j, --json                        Run the query and return the object in a JSON structure.
   -o, --output string               Set output format: compact, csv, full, json, junit, report, summary, yaml (default "compact")
+      --output-target string        Set output target to which the asset report will be sent. Currently only supports AWS SQS topic URLs and local files
```
chris-rock, line 29: Can we add a cli test that runs a single-check policy, and stores the output on a local disk?

apps/cnspec/cmd/scan.go
```diff
@@ -135,7 +142,21 @@ var scanCmdRun = func(cmd *cobra.Command, runtime *providers.Runtime, cliRes *pl
 	}
 
 	logger.DebugDumpJSON("report", report)
-	printReports(report, conf, cmd)
+	outputHandler, err := reporter.NewOutputHandler(conf.OutputTarget, conf.OutputFormat)
```
chris-rock: I am a bit confused with the output handling here:

1. I would expect that cli output is a proper output handler as well with target STDOUT and format cli
2. This would harmonize the code and also allow us to report to multiple outputs in the future
3. I like that we keep cli out and allow additional output but it would be more explicit if it would be two instances of the output handler.

test/cli/testdata/cnspec_scan.ct
chris-rock: We do that in follow up PRs.

chris-rock (approved):
Thank you @preslavgerchev

PR #983, opened 2023-12-06
🧹 make bundle fmt more robust and sanitize policy docs content as well as properties

chris-rock:
- replace tabs and /r with /n
- ensure x00 is not used

### DIFF
--- a/internal/bundle/fmt.go
+++ b/internal/bundle/fmt.go
@@ -63,7 +63,23 @@ func ParseYaml(data []byte) (*Bundle, error) {
 func sanitizeStringForYaml(s string) string {
 	lines := strings.Split(s, "\n")
 	for j := range lines {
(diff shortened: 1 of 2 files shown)


internal/bundle/fmt.go
```diff
@@ -94,10 +110,24 @@ func FormatFile(filename string, sort bool) error {
 // Format formats the Bundle
 func FormatBundle(b *Bundle, sort bool) ([]byte, error) {
 	// to improve the formatting we need to remove the whitespace at the end of the lines
+
+	for i := range b.Policies {
```
preslavgerchev, line 114: can we add a test for the new sanitization?

chris-rock: Completed

preslavgerchev (approved):

PR #985, opened 2023-12-06
⭐️ Add reporting jobs for data queries.

preslavgerchev:
Until now we had always deleted the data queries' reporting jobs and attached the datapoints to their parent RJs. With this change we instead attach the datapoints to the data queries' RJs and we do not delete them. 

This means that we now have scores coming in for data queries, even though those will always be of the unscored type.

This brings scoring and data queries a lot closer in terms of their behaviour in the resolver. For instance, we no longer need to do special handling for framework controls and data queries, those work just like the checks.

I need to write more tests for this but the code changes should be there

- [x] tests

### DIFF
--- a/policy/resolver.go
+++ b/policy/resolver.go
@@ -1138,8 +1138,6 @@ func (s *LocalServices) jobsToQueries(ctx context.Context, policyMrn string, cac
 		Queries: map[string]*ExecutionQuery{},
 	}
 
-	// fill in all reporting jobs. we will remove the data query jobs and replace
-	// them with direct collections into their parent job later
 	for _, rj := range cache.reportingJobsByUUID {
 		collectorJob.ReportingJobs[rj.Uuid] = rj
 	}
--- a/policy/resolver.go
+++ b/policy/resolver.go
@@ -1034,41 +1034,39 @@ func (cache *policyResolverCache) addCheckJobVariants(ctx context.Context, query
 }
 
 func (cache *policyResolverCache) addDataQueryJob(ctx context.Context, query *explorer.Mquery, ownerJob *ReportingJob) {
-	if len(query.Variants) != 0 {
-		err := cache.addDataQueryVariants(ctx, query, ownerJob)
-		if err != nil {
-			log.Error().Err(err).Str("queryMrn", query.Mrn).Msg("failed to add data query variants")
-		}
-		return
-	}
-
-	uuid := cache.global.relativeChecksum(query.Checksum)
+	uuid := cache.global.relativeChecksum(query.Mrn)
 	queryJob := cache.global.reportingJobsByUUID[uuid]
 
-	// note: the ReportingJob is only a placeholder and is replaced by individual query LLX checksum ReportingJobs
 	if queryJob == nil {
 		queryJob = &ReportingJob{
-			Uuid:       cache.global.relativeChecksum(query.Checksum),
+			Uuid:       uuid,
 			QrId:       query.Mrn,
 			ChildJobs:  map[string]*explorer.Impact{},
 			Datapoints: map[string]bool{},
 			Type:       ReportingJob_DATA_QUERY,
-			// FIXME: DEPRECATED, remove in v10.0 vv
-			DeprecatedV8IsData: true,
 		}
-		cache.global.reportingJobsByUUID[queryJob.Uuid] = queryJob
+		cache.global.codeIdToMrn[query.CodeId] = append(cache.global.codeIdToMrn[query.CodeId], query.Mrn)
+		cache.global.reportingJobsByUUID[uuid] = queryJob
 		cache.global.reportingJobsByMsum[query.Checksum] = append(cache.global.reportingJobsByMsum[query.Checksum], queryJob)
 		cache.childJobsByMrn[query.Mrn] = append(cache.childJobsByMrn[query.Mrn], queryJob)
 	}
 
 	// local aspects for the resolved policy
 	queryJob.Notify = append(queryJob.Notify, ownerJob.Uuid)
-
-	ownerJob.Datapoints[queryJob.Uuid] = true
-	// we set a placeholder for the execution query, just to indicate it will be added
-	cache.global.executionQueries[query.Checksum] = nil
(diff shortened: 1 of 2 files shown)


policy/resolver.go
```diff
@@ -1271,36 +1262,19 @@ func (s *LocalServices) jobsToQueries(ctx context.Context, policyMrn string, cac
...
-
-				if err = connectDatapointsToReportingJob(executionQuery, parentJob, collectorJob.Datapoints); err != nil {
-					return nil, nil, err
-				}
-
-				// we don't need this any longer, since every datapoint now reports into
-				// the parent job (i.e. reports into the policy instead of the data query)
-				delete(collectorJob.ReportingJobs, rj.Uuid)
-				delete(parentJob.Datapoints, rj.Uuid)
+			// We connect the datapoints to the data query reporting job. Previously we had connected the datapoints
```
preslavgerchev, line 1273: This is the biggest change. We don't remove the data query's RJ. Instead if we connect the datapoint to it.

policy/resolver.go
```diff
@@ -1582,24 +1555,20 @@ func (s *LocalServices) jobsToControls(cache *frameworkResolverCache, framework
...
 			uuid := cache.relativeChecksum(mquery.Mrn)
-			queryJob := &ReportingJob{
-				Uuid:      uuid,
-				QrId:      mquery.Mrn,
-				ChildJobs: map[string]*explorer.Impact{},
-				Type:      ReportingJob_DATA_QUERY,
-			}
-			nuJobs[uuid] = queryJob
-			for controlMrn := range targets {
-				controlJob := ensureControlJob(cache, nuJobs, controlMrn, framework, frameworkGroupByControlMrn)
```
preslavgerchev, line 1592: We no longer need this, there is code below that ties framework controls to its links (both queries and checks)

jaym:
I tested it out and it behaves as described. It might be good to have tests for this specific case

preslavgerchev:
> I tested it out and it behaves as described. It might be good to have tests for this specific case

I extended the simple policy test case to verify that we also have RJs for data queries when not tied to a framework

policy/resolver.go
```diff
@@ -1034,41 +1034,39 @@ func (cache *policyResolverCache) addCheckJobVariants(ctx context.Context, query
...
 	if queryJob == nil {
 		queryJob = &ReportingJob{
-			Uuid:       cache.global.relativeChecksum(query.Checksum),
+			Uuid:       uuid,
 			QrId:       query.Mrn,
 			ChildJobs:  map[string]*explorer.Impact{},
 			Datapoints: map[string]bool{},
 			Type:       ReportingJob_DATA_QUERY,
-			// FIXME: DEPRECATED, remove in v10.0 vv
-			DeprecatedV8IsData: true,
```
czunker, line 1057: Shouldn't we keep this until v10?

policy/resolver.go
```diff
@@ -1034,41 +1034,39 @@ func (cache *policyResolverCache) addCheckJobVariants(ctx context.Context, query
...
 
 	// local aspects for the resolved policy
 	queryJob.Notify = append(queryJob.Notify, ownerJob.Uuid)
-
-	ownerJob.Datapoints[queryJob.Uuid] = true
-	// we set a placeholder for the execution query, just to indicate it will be added
-	cache.global.executionQueries[query.Checksum] = nil
-	cache.global.dataQueries[query.Checksum] = struct{}{}
-	cache.global.queriesByMsum[query.Checksum] = query
+	if ownerJob.ChildJobs[queryJob.Uuid] == nil {
```
czunker, line 1058: nit: Isn't `queryJob.Uuid` also `uuid` here?

policy/resolver.go
```diff
@@ -1034,41 +1034,39 @@ func (cache *policyResolverCache) addCheckJobVariants(ctx context.Context, query
...
+		ownerJob.ChildJobs[queryJob.Uuid] = query.Impact
```
czunker, line 1059: From what I understood, we do not set `Impact`. So, this could be `nil`. At other places in the code, we have these checks:
```
if impact == nil {
	impact = &explorer.Impact{}
}
```
Should we add this here, too?

policy/resolver_test.go
```diff
@@ -132,19 +132,28 @@ policies:
...
+		require.Len(t, rp.Filters, 1)
+		require.Len(t, rp.CollectorJob.ReportingJobs, 3)
+
+		qrIdToRj := map[string]*policy.ReportingJob{}
+		for _, rj := range rp.CollectorJob.ReportingJobs {
+			qrIdToRj[rj.QrId] = rj
+		}
+		// scoring queries report by code id
+		require.NotNil(t, qrIdToRj[b.Queries[1].CodeId])
+		// data queries report by mrn
```
czunker, line 152: Is `codeId` vs. `mrn` by intention, or should we clean this up in a follow-up?

policy/resolver.go
preslavgerchev: yep, but i prefer `queryJob.Uuid` as it provides slightly more context as to *what* uuid we're talking about

policy/resolver_test.go
preslavgerchev: currently by intention. scoring queries go by code id, data queries now go by mrn (this is new, we havent had reporting jobs for data queries before)

in the future, i think we want everything to report by mrn

policy/resolver.go
preslavgerchev: I have copied this over from the `addCheckJob` func, I haven't seen that elsewhere

policy/resolver.go
preslavgerchev: Good point, I will add back. I don't think we use this anywhere but it does not hurt to keep it for the time being

policy/resolver.go
czunker: I saw it here: https://github.com/mondoohq/cnspec/blob/main/policy/resolver.go#L906-L908

But when it works for `addCheckJob`, I assume it is safe.

czunker (approved):
Thanks @preslavgerchev

PR #987, opened 2023-12-07
✨ add GetBundleV2 to PolicyHub

imilchev:
The idea is that this function will let us return a minimal version of the bundle for a given MRN. This should reduce the amount of network traffic we incur for each scan and should make scans faster.

The actual usage of this new endpoint will be done in another PR. For now I want to merge this, so I can add the implementation to the backend


policy/hub.go
```diff
@@ -542,14 +552,21 @@ func (s *LocalServices) computePolicyBundle(ctx context.Context, bundleMap *Poli
...
 	if s.Upstream == nil {
 		return nil, errors.New("failed to retrieve upstream policy " + mrn + " since upstream is not defined")
 	}
 
 	logCtx.Debug().Str("policy", mrn).Msg("marketplace> fetch policy bundle from upstream")
-	bundle, err := s.Upstream.GetBundle(ctx, &Mrn{Mrn: mrn})
+	var bundle *Bundle
+	var err error
+
+	if minimal {
```
chris-rock: can we make this a feature flag instead that we set into context? See https://github.com/mondoohq/cnquery/blob/main/featureflags.go and https://github.com/mondoohq/cnquery/blob/main/featureflags.go#L120-L122

policy/hub.go
```diff
@@ -542,14 +552,21 @@ func (s *LocalServices) computePolicyBundle(ctx context.Context, bundleMap *Poli
...
+		bundle, err = s.Upstream.GetMinimalBundle(ctx, &Mrn{Mrn: mrn})
+	} else {
+		bundle, err = s.Upstream.GetBundle(ctx, &Mrn{Mrn: mrn})
```
chris-rock: Feature flags are translated to http header, then the server could return the right thing based on a feature flag. Then we do not need two methods. See https://github.com/mondoohq/cnquery/blob/main/apps/cnquery/cmd/root.go#L148-L171

policy/hub.go
imilchev: But the goal is not to globally remove the docs. I believe there are cases where those docs may be needed. I want to only remove them if explicitly asked for that. The only way to do this is to have 2 different endpoints or 1 endpoint with an input parameter that can toggle this behaviour.

policy/cnspec_policy.proto
```diff
@@ -518,6 +518,7 @@ service PolicyHub {
   rpc SetBundle(Bundle) returns (Empty) {}
   rpc ValidateBundle(Bundle) returns (Empty) {}
   rpc GetBundle(Mrn) returns (Bundle) {}
+  rpc GetBundleForMrn(BundleReq) returns (Bundle) {}
```
preslavgerchev: nit: can we call this `GetBundleV2`? no real preference here, just sounds better to me 🤷

policy/cnspec_policy.proto
```diff
@@ -518,6 +518,7 @@ service PolicyHub {
   rpc SetBundle(Bundle) returns (Empty) {}
   rpc ValidateBundle(Bundle) returns (Empty) {}
   rpc GetBundle(Mrn) returns (Bundle) {}
```
preslavgerchev: can we mark this as deprecated so that we start phasing it out slowly?

imilchev:
not needed anymore

PR #988, opened 2023-12-07
⭐️ Allow loading policies from an s3 bucket.

preslavgerchev:
Introduce a `BundleResolver` interface and 2 concrete implementations:
 - A local (file) resolver
 - An s3 resolver

The `BundleFromPaths` will iterate through the possible resolvers and find the right one for the target we're specifying

This allows for pulling policies from an s3 bucket:
```
~/go/bin/cnspec scan -f s3://mysupernotexistingbucket1234567
```
Or local (like before, unchanged)
```
~/go/bin/cnspec scan -f packs.yaml
```

It also supports pulling a specific s3 item:
```
~/go/bin/cnspec scan -f s3://mysupernotexistingbucket1234567/packs.mql.yaml
```
It also works with both at the same time:
```
// s3 bucket + local
~/go/bin/cnspec scan -f s3://mysupernotexistingbucket1234567 -f packs.yaml
```
```
// s3 key + local
~/go/bin/cnspec scan -f s3://mysupernotexistingbucket1234567/pol.mql.yaml -f packs.yaml
```


policy/bundle_s3_resolver.go
```diff
@@ -0,0 +1,59 @@
...
+
+package policy
+
+import (
+	"context"
+	"io"
+	"strings"
+
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+
```
chris-rock: nit whitespace

policy/bundle_s3_resolver.go
```diff
@@ -0,0 +1,59 @@
...
+type s3Resolver struct{}
+
+func (l *s3Resolver) Load(ctx context.Context, path string) (*Bundle, error) {
+	cfg, err := awsconfigv2.LoadDefaultConfig(ctx)
+	if err != nil {
+		return nil, err
+	}
+	bucket := strings.TrimPrefix(path, "s3://")
+	s3Client := s3.NewFromConfig(cfg)
+	files, err := s3Client.ListObjects(ctx, &s3.ListObjectsInput{Bucket: &bucket})
```
chris-rock: do we need pagination here?

policy/bundle_s3_resolver.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return nil, err
+	}
+	mergedBundle := &Bundle{}
+	resolvedNames := []string{}
```
chris-rock, line 54: How can we pull just one key? like `https://mybucket.s3.amazonaws.com/policy.mql.yaml`

policy/bundle_s3_resolver.go
preslavgerchev: the API specifies this returns up to a 1000 items. i'd say we can optimize this later if there's a use case for that

policy/bundle_s3_resolver.go
preslavgerchev: Great point, that was not implemented (yet). I added this in with the new commit.

go.mod
```diff
@@ -6,7 +6,10 @@ go 1.21
 
 toolchain go1.21.3
 
-require go.mondoo.com/cnquery/v9 v9.10.0
+require (
+	github.com/aws/aws-sdk-go-v2/service/s3 v1.47.3
```
chris-rock: I think we should keep the dep in the second group, makes it easier to understand

policy/bundle_file_resolver.go
```diff
@@ -0,0 +1,118 @@
...
+
+import (
+	"context"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+
```
chris-rock: nit: whitespace

chris-rock (approved):
Thank you @preslavgerchev Great work.

Can you please fix the nits and add some tests for the bundle resolver?

PR #994, opened 2023-12-11
🧹 do not call SynchronizeAssets twice

imilchev:
don't call `SynchronizeAssets` once when the assets are discovered and then another time before scanning each asset. That seems unnecessary. I ran a few tests with MacOS and Kubernetes and it seems like this changes nothing. Would be nice if other people can test it too

### DIFF
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -466,6 +466,10 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 				platformMrn := batch[i].asset.PlatformIds[0]
 				batch[i].asset.Mrn = platformAssetMapping[platformMrn].AssetMrn
 				batch[i].asset.Url = platformAssetMapping[platformMrn].Url
+				if batch[i].asset.Labels == nil {
+					batch[i].asset.Labels = map[string]string{}
+				}
+				batch[i].asset.Labels["mondoo.com/project-id"] = platformAssetMapping[platformMrn].ProjectId
 			}
 		} else {
 			// ensure we have non-empty asset MRNs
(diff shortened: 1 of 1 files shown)


czunker:
I gave it a try with
```
cnspec scan container registry gcr.io/mondoo-dev-XXX/ubuntu
```
LGTM. First, assets appear unscored and later scored.

policy/scan/local_scanner.go
```diff
@@ -523,54 +528,9 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
...
 	log.Debug().Msgf("connecting to asset %s", job.Asset.HumanName())
 
-	upstream := s.upstreamServices(job.UpstreamConfig)
-	if upstream != nil {
-		resp, err := upstream.SynchronizeAssets(job.Ctx, &policy.SynchronizeAssetsReq{
-			SpaceMrn: job.UpstreamConfig.SpaceMrn,
-			List:     []*inventory.Asset{job.Asset},
-		})
-		if err != nil {
-			log.Error().Err(err).Msgf("failed to synchronize asset to Mondoo Platform %s", job.Asset.Mrn)
```
czunker, line 579: That's the only difference I can spot.

When a single asset cannot be synced, we do not add an error to the progress bar, but ...
Actually not sure, what happens. I'd like to create a testcase for this.

czunker: I took the registry scan to create an inventory and changed one asset to not be reachable.

The behavior between the current cnspec and this PR is the same. Three assets get synchronized upstream.

preslavgerchev:
This may be a v8 remnant. Afaik we had 2 sync calls as some of the asset properties (platform info, etc) were not set during the first call (discovery) but were available during the scan. I will test it out with Azure

vjeffrey:
ill give it a whirl with aws

vjeffrey:
i actually did see a difference - without this change, running `scan aws --discover instances` gives me two instances with names. with this change, those instances don't have names

imilchev:
I tested what @vjeffrey described. This isn't working with v9.11.0 either. That problem isn't introduced by the change in this PR

preslavgerchev:
Azure and ms365 also lgtm to me with those changes.

czunker:
I'm fine merging this.

@vjeffrey / @preslavgerchev Thumbs up?

I'd suggest putting the `asset.name` with aws discovery into a separate issue.

preslavgerchev (approved):

PR #995, opened 2023-12-12
🧹 batch sync assets

imilchev:
Instead of calling `SynchronizeAssets` once for all discovered assets, we create batches. This makes sure that if we have discovered a huge amount of assets, the call would still succeed. 

I intentionally put a low number for the batch size because with this implementation, batches are scanned in parallel, which also decreases the total scan time.

Downside of this approach is that the progress bar results ordering is no longer deterministic. However, since this is a progress bar, I don't think ordering should matter

### DIFF
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -410,7 +410,9 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 	scanGroups.Add(1)
 	go func() {
 		defer scanGroups.Done()
-		multiprogress.Open()
+		if err := multiprogress.Open(); err != nil {
+			log.Error().Err(err).Msg("failed to open progress bar")
+		}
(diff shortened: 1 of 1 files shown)


jaym:
cnquery/cnspec and its providers are in general not thread safe and doing this will trigger data races.  Not sure how to get this information for the providers, but you can at least see these data races fairly quickly:
```
go run -race ./apps/cnspec/cnspec.go scan k8s
```

imilchev:
it looks like those are related to the `BufferedCollector` and ranger somehow. I am not entirely sure how to read those. I will not execute anything in parallel here. Will make the whole PR synchronous but still working with batches so we can unblock the fix for the original issue

imilchev:
The scans are now synchronous, batch size is 100

czunker:
I gave this a try with a container registry scan. Works as expected.

PR #1008, opened 2023-12-20
🐛 do not report for disabled controls

imilchev:
We don't want to report scores for disabled/out-of-scope controls. This was never intended and seems to be a bug

### DIFF
--- a/policy/framework.go
+++ b/policy/framework.go
@@ -548,11 +548,33 @@ func ResolveFramework(mrn string, frameworks map[string]*Framework) *ResolvedFra
 		Nodes:         map[string]ResolvedFrameworkNode{},
 	}
 
+	// note: this code will disable a control if it is disabled in any framework in the hirearchy
+	// (asset or space-level). This should be fine for now since we don't have a use-case of enabling
+	// an already disabled control on a lower level in the hierarchy.
+	disabledControls := map[string]struct{}{}
+	for _, framework := range frameworks {
+		for _, group := range framework.Groups {
+			// If the group is not an exception group or has been rejected, skip over it
+			if (group.Type != GroupType_DISABLE && group.Type != GroupType_OUT_OF_SCOPE) ||
+				group.ReviewStatus == ReviewStatus_REJECTED {
+				continue
+			}
+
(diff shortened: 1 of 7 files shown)


policy/framework.go
```diff
@@ -548,11 +548,33 @@ func ResolveFramework(mrn string, frameworks map[string]*Framework) *ResolvedFra
...
 	}
 
+	// note: this code will disable a control if it is disabled in any framework in the hirearchy
+	// (asset or space-level). This should be fine for now since we don't have a use-case of enabling
+	// an already disabled control on a lower level in the hierarchy.
+	disabledControls := map[string]struct{}{}
+	for _, framework := range frameworks {
+		for _, group := range framework.Groups {
+			// If the group is not an exception group or has been rejected, skip over it
+			if (group.Type != GroupType_DISABLE && group.Type != GroupType_OUT_OF_SCOPE) ||
```
czunker: We have similar code in the policy resolver: https://github.com/mondoohq/cnspec/blob/main/policy/resolver.go#L1504-L1523

I'm not sure whether that's actually needed here.

policy/scan/local_scanner_test.go
```diff
@@ -138,10 +138,10 @@ func (s *LocalScannerSuite) BeforeTest(suiteName, testName string) {
 							{
 								Type: "k8s",
 								Options: map[string]string{
-									"path": "./testdata/2pods.yaml",
+									"path": "./testdata/1pod.yaml",
 								},
 								Discover: &inventory.Discovery{
-									Targets: []string{"auto"},
+									Targets: []string{"pods"},
```
czunker, line 144: Why needs this test to change?

imilchev: I wanted to have 1 report instead of 3. Easier to write the tests

policy/framework.go
imilchev: I am also not entirely sure when that other code is executed. I know for sure though that this bit here determines whether we report scores for controls

czunker (approved):
Nice!. Thanks for the fix and all the tests.

PR #1012, opened 2023-12-22
🐛 add review status to framework execution checksum

imilchev:
The review status actually changes the execution for a framework so we need to include it in the execution checksum

### DIFF
--- a/policy/framework.go
+++ b/policy/framework.go
@@ -374,6 +374,7 @@ func (f *Framework) updateAllChecksums(ctx context.Context,
 		}
 
 		executionChecksum = executionChecksum.AddUint(uint64(group.Type))
+		executionChecksum = executionChecksum.AddUint(uint64(group.ReviewStatus))
 
 		if group.Docs != nil {
 			contentChecksum = contentChecksum.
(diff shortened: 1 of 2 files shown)


policy/framework.go
```diff
@@ -374,6 +374,7 @@ func (f *Framework) updateAllChecksums(ctx context.Context,
 		}
 
 		executionChecksum = executionChecksum.AddUint(uint64(group.Type))
+		executionChecksum = executionChecksum.AddUint(uint64(group.ReviewStatus))
```
preslavgerchev, line 377: do we have tests for checksums? would be nice to add that there

imilchev: we don't but I can try to add one that checks that the checksum is different when the review status changes

preslavgerchev: +1, we should build up the list of test cases here, a minimal (just review status change) start is great

preslavgerchev (approved):

policy/framework.go
preslavgerchev: tests 🎉

PR #1023, opened 2024-01-03
🏃🏻‍♂️ retrieve space bundle instead of asset bundle for upstream scans

imilchev:
We don't need the asset bundle when we run upstream scans. All we need is the space bundle, which already contains all queries, controls, policies and frameworks. This speeds up scans significantly. 

The more policies/frameworks are enabled in a space, the more this change will speed up the scans when compared to the latest cnspec release

I tested this with a sample space and a fresh minikube instance (11 assets). Total scan duration:
  - Old: 24s
  - New: 16s


imilchev:
not needed anymore

PR #1026, opened 2024-01-07
⭐️ experimental update sub-command

chris-rock:
This PR adds the experimental update command to cnspec

```
cnspec update
```

It download the install script and runs it locally.

### DIFF
--- a/apps/cnspec/cmd/update.go
+++ b/apps/cnspec/cmd/update.go
@@ -1,0 +1,134 @@
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"os/exec"
+	"runtime"
+
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+	"github.com/spf13/cobra"
+	"go.mondoo.com/cnquery/v9/cli/config"
+)
+
+func init() {
+	rootCmd.AddCommand(updateCmd)
+}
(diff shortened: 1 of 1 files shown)


apps/cnspec/cmd/update.go
```diff
@@ -0,0 +1,134 @@
...
+
+	_, err = file.Write(scriptData)
+	if err != nil {
+		return errors.Wrap(err, "could not write downloaded script")
+	}
+
+	log.Info().Str("script", file.Name()).Msg("script downloaded successfully")
+
+	log.Info().Msg("run update script")
+	cmd := exec.Command("bash", file.Name())
```
czunker, line 83: Not every Linux has `bash` installed. Can we changed this? Or should we check that bash is present?

czunker:
I do not understand what this command will be used for. Why not let the package manager do the update?

chris-rock:
> I do not understand what this command will be used for. Why not let the package manager do the update?

That only works well on Linux systems and is more difficult on windows and macOS. To ensure we just use the package manager, we rely on the install script.

Essentially this command makes it easier for users, since the do not need to manually download the script and run the script. It came up quite a lot as feedback from Windows users.

apps/cnspec/cmd/update.go
chris-rock: Right now the script requires bash, therefore we should check for its existence. Great point.

chris-rock:
@czunker PR is updated and checks for bash now

czunker (approved):
Worked with my tests on a Linux box.

Thanks @chris-rock

PR #1030, opened 2024-01-08
🐛 fix merging of `for`-referenced props + add example

arlimus:
Note: Please update cnquery to latest post https://github.com/mondoohq/cnquery/pull/2973 and this should work

1. The `for` field is used to specify what a prop is meant to set. Unfortunately it was not used during resolution, so it was effectively ignored so far.

2. Properties have been hard to use because they lack documentation. This PR adds a full example on how properties work and how they can be used across properties.

Given the new example, we can now run:

```bash
cnspec scan -f examples/props.mql.yaml
```

```
Data queries:
home.+: "ex on 1"
home.+: "ex on 2"
homeDir.+: "ex on 3"
```

It also now supports overrides:

```bash
cnspec scan -f examples/props.mql.yaml --props userHome="'nya'"
```

```
Data queries:
home.+: "nya on 1"
home.+: "nya on 2"
homeDir.+: "nya on 3"
```

### DIFF
--- a/examples/props.mql.yaml
+++ b/examples/props.mql.yaml
@@ -1,0 +1,51 @@
+# Copyright (c) Mondoo, Inc.
+# SPDX-License-Identifier: BUSL-1.1
+
+# This bundle contains two policies
+policies:
+  - uid: example1
+    name: Example policy 1
+    version: "1.0.0"
+    authors:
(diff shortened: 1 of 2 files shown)


chris-rock (approved):
Very nice feature!

PR #1033, opened 2024-01-08
⭐️ add policy subcommand

imilchev:
- [x] policy list
- [x] policy upload
- [x] policy delete
- [x] policy info
- [x] policy download
- [x] policy enable
- [x] policy disable
- [x] policy init
- [x] policy format
- [x] policy lint
- [x] policy upload


vjeffrey:
looking nice!

Screenshot 2024-01-10 at 13 45 27

as a followup improvement, could we make the policy list pageable? the list can be pretty long

is there a way to list not-enabled policies? we have an enable command, but when i went to try it i couldn't find how i would know the mrn/uid of the policy i want to enable

how do u provide the output file for the download? i couldn't figure it out on cli (i mean, i know i can go look at the code now, but maybe we need something in the description to help or something?)

apps/cnspec/cmd/bundle.go
```diff
@@ -57,114 +74,33 @@ func ensureProviders() error {
...
-				fmt.Println(err)
-				os.Exit(1)
-			}
-		}
-		log.Info().Msg("completed formatting policy bundle(s)")
-	},
+	Run: runPolicyLint,
 }
 
 var policyPublishCmd = &cobra.Command{
```
chris-rock, line 97: Is there a reason why we are not using the implementation from policy upload here?

policy/scan/scan.pb.go
```diff
@@ -3,7 +3,7 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.32.0
+// 	protoc-gen-go v1.31.0
```
chris-rock: Please update the protoc-gen-go and remove those changes form PR

upstream/gql.go
```diff
@@ -0,0 +1,89 @@
...
+) ([]*policy.Policy, error) {
+	var q struct {
+		Content struct {
+			TotalCount int `json:"totalCount"`
+			Edges      []struct {
+				Cursor string `json:"cursor"`
+				Node   struct {
+					Policy struct {
+						Mrn  string
+						Name string
```
chris-rock: We should fetch the `action` here as well, so that we know the policy is enabled, preview not disabled. In addition, `trustLevel` and `access` help us to determine if the policy is private and uploaded to the space.

apps/cnspec/cmd/policy.go
```diff
@@ -0,0 +1,747 @@
...
+			}
+
+			policies, err = cnspec_upstream.SearchPolicy(
+				context.Background(), mondooClient, opts.GetParentMrn(), ptr.To(true), ptr.To(true), ptr.To(true))
+			if err != nil {
+				return err
+			}
+		}
+		for _, policy := range policies {
+			fmt.Println(policy.Name + " " + policy.Version)
```
chris-rock:
```
My Policy Title v1.0.0 (enabled, private)
  //policy.api.mondoo.app/policies/mondoo-kubernetes-security
  
My Policy Title v1.0.0 (preview)
  //policy.api.mondoo.app/policies/mondoo-kubernetes-security
```

There are three different scenarios that we want to handle here:

1. Policy is enabled or in preview, then we show it in brackets
2. If a policy is not enabled we do not show anything
3. If we policy is private, we display private, otherwise we do not add text

chris-rock (commented):
Looks really good and works like a charm. Added a few smaller things that would make this PR even more awesome.

apps/cnspec/cmd/policy.go
imilchev: Do we have preview policies?

apps/cnspec/cmd/bundle.go
imilchev: I already had implemented a new one when I saw this. I just kept the new one, because I verified it works. No other reason

chris-rock (approved):
Thank you @imilchev That is a great addition.

PR #1036, opened 2024-01-09
🐛 Fix issue with policies matching against incorrect platforms

jaym:
The issues here is the way policies are selected to be included. It checks this field called ComputedFilters of the policy. I'm not 100% sure what this is, but from the comment, it sounds like this has any filter that this policy uses. However, I don't think that is useful to answer the question of whether the policy should be included. For example, I run arch linux. If I include the alma linux 8 cis benchmark, that policy is included. That is because the following query is defined:

```
  - uid: cis-almalinux-8--4.2.11
    title: Ensure sshd KexAlgorithms is configured
    impact: 100
    filters: |
      file("/etc/crypto-policies/config").exists == false
      command('update-crypto-policies --show').stdout.trim == ""
...
```

This change fixes the problem by pruning out any reporting job nodes that are policies that have no children or datapoints

### DIFF
--- a/policy/resolver_test.go
+++ b/policy/resolver_test.go
@@ -265,12 +265,7 @@ policies:
 	})
 	require.NoError(t, err)
 	require.NotNil(t, rp)
-	require.Len(t, rp.CollectorJob.ReportingJobs, 2)
-	for _, rj := range rp.CollectorJob.ReportingJobs {
-		if rj.Type == policy.ReportingJob_CHECK {
-			require.Fail(t, "expected no check reporting job")
-		}
-	}
+	require.Len(t, rp.CollectorJob.ReportingJobs, 0)
 }
 
 func TestResolve_IgnoredQuery(t *testing.T) {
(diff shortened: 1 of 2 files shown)


policy/resolver_test.go
```diff
@@ -265,12 +265,7 @@ policies:
 	})
 	require.NoError(t, err)
 	require.NotNil(t, rp)
-	require.Len(t, rp.CollectorJob.ReportingJobs, 2)
-	for _, rj := range rp.CollectorJob.ReportingJobs {
-		if rj.Type == policy.ReportingJob_CHECK {
-			require.Fail(t, "expected no check reporting job")
-		}
-	}
+	require.Len(t, rp.CollectorJob.ReportingJobs, 0)
```
jaym, line 268: This seems to be an unintended consequence of this change. If you disable all the checks for a policy, it looks no different than disabling the policy

arlimus: I think it's fine to prune a policy from execution if no checks+queries are run, this is good behavior

arlimus (approved):
> The issues here is the way policies are selected to be included. It checks this field called ComputedFilters of the policy. I'm not 100% sure what this is, but from the comment, it sounds like this has any filter that this policy uses. However, I don't think that is useful to answer the question of whether the policy should be included.

Correct, it covers all checks in the policy, no matter where they are. I had decided to pre-run as many filters as possible to reduce the amount of client-server round-trips that are necessary to resolve. The idea was to cascade filters down the line (ie a cascaded filter collection+resolution), so that clients would be able to run fewer things, but we haven't gotten to that part yet.

The overall computed filters inform the selection of policies, but cannot determine it all on their own, as you have pointed out. That was the issue I had introduced during the update. Pruning policies in order based on the returned filters is the correct approach. I'd love to revisit this entire algo and optimize it down the line, including the resolution and pruning.

As for this PR: Thank you so much for fixing this oversight!

PR #1037, opened 2024-01-10
🧹 Add bytes reporter to support the packer plugin

czunker:
This adds a reporter which allows to access the rendered report as bytes.Buffer. This is needed, so we can output it inside the packer plugin.

For reference: https://github.com/mondoohq/packer-plugin-cnspec/pull/170

### DIFF
--- a/cli/reporter/cli_reporter.go
+++ b/cli/reporter/cli_reporter.go
@@ -101,6 +101,13 @@ func NewReporter(format Format, incognito bool) *Reporter {
 	}
 }
 
+// This allows the packer-plugin-cnspec to set the output writer directly
+// The plugin needs this to work the the packer output
+func (r *Reporter) WithOutput(out io.Writer) *Reporter {
+	r.out = out
+	return r
+}
+
 func (r *Reporter) WriteReport(ctx context.Context, data *policy.ReportCollection) error {
 	switch r.Format {
 	case Compact:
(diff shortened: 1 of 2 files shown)


cli/reporter/cli_reporter.go
```diff
@@ -101,6 +101,13 @@ func NewReporter(format Format, incognito bool) *Reporter {
 	}
 }
 
+// This allows the packer-plugin-cnspec to set the output writer directly
```
preslavgerchev: nit: can we update the comment to just describe what this func does? we shouldnt reference which component needs that as the comment might become outdated over time

cli/reporter/file_handler.go
```diff
@@ -27,9 +27,6 @@ func (h *localFileHandler) WriteReport(ctx context.Context, report *policy.Repor
 	}
 	defer f.Close() //nolint: errcheck
 	reporter := NewReporter(h.format, false)
-	if err != nil {
```
preslavgerchev, line 30: nice

cli/reporter/cli_reporter.go
preslavgerchev: also consider adding this to the `NewReporter` func instead, would be nice to get out the hardcoded `os.Stdout` there

preslavgerchev (approved):

cli/reporter/cli_reporter.go
czunker: But, isn't `os.Stdout` the correct default for the CLI reporter?

preslavgerchev (approved):

PR #1040, opened 2024-01-11
🏃🏻‍♂️ improve scan speed by not retrieving bundles for assets

imilchev:
After #1023 and some thoughts and discussions on this topic, I tried out something else. It turns out we don't need the bundle at all for execution and reporting results upstream. We only use it to print the CLI report. 

This PR makes sure we aren't pulling the bundle for every asset. It needs a small backend change that will allow agents to call `GetPolicyFilters` but apart from that it works exactly the same as the old code

Another benefit of this approach is that we will use significantly less network traffic. This has become a problem for big k8s cluster that result in many assets being scanned

### DIFF
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -1020,6 +1020,7 @@ func (s *localAssetScanner) runPolicy() (*policy.ResolvedPolicy, error) {
 	}
 	log.Debug().Str("asset", s.job.Asset.Mrn).Msg("client> got policy filters")
 	logger.TraceJSON(rawFilters)
+	logger.DebugDumpYAML("policyFilters", rawFilters)
 
 	filters, err := s.UpdateFilters(&explorer.Mqueries{Items: rawFilters.Items}, 5*time.Second)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


policy/scan/local_scanner.go
```diff
@@ -979,29 +998,32 @@ func (s *localAssetScanner) ensureBundle() error {
...
+		}
+		log.Debug().Msg("client> got policy bundle")
+		logger.TraceJSON(assetBundle)
+		logger.DebugDumpYAML("assetBundle", assetBundle)
 	}
-	log.Debug().Msg("client> got policy bundle")
-	logger.TraceJSON(assetBundle)
-	logger.DebugDumpJSON("assetBundle", assetBundle)
 
 	rawFilters, err := hub.GetPolicyFilters(s.job.Ctx, &policy.Mrn{Mrn: s.job.Asset.Mrn})
```
jaym, line 1017: we should probably dump `rawFilters` to a file since that is what actually gets used

jaym (approved):

policy/scan/local_scanner.go
imilchev: added

imilchev:
Here is a graph of a k8s scan that shows the network bandwidth usage for during the scan. 
- The scan took 9mins
- The download speed peaked at about 1MiB/s
image

Here is the same graph with the changes from this PR
- The scan took 5mins
- The download speed peaked at about 360KiB/s but quickly dropped to about 170KiB/s
image

PR #1046, opened 2024-01-15
⭐️ panic handler to report panics

chris-rock:
applies the cnquery panic handler to cnspec as well, see https://github.com/mondoohq/cnquery/pull/2963


arlimus (approved):
those are great!

PR #1052, opened 2024-01-17
✨ Add support for sending scan reports to an Azure Service bus.

preslavgerchev:
Allows sending reports to an azure service bus queue or a topic.

```
~/go/bin/cnspec scan --output-target https://test-bus-azure.servicebus.windows.net/msg-queue --output json
```

```
~/go/bin/cnspec scan --output-target https://test-bus-azure.servicebus.windows.net/msg-topic --output json
```

Currently, this only supports default azure auth currently (which is done via the AZ cli). Future extension to support other type of auth (via secret/cert) will require more work


chris-rock (approved):
Thank you @preslavgerchev

PR #1060, opened 2024-01-22
🧹use custom provider directory for pipeline

chris-rock:
This makes sure different runs use different provider install directories. It shields PRs better and makes sure we test properly.

### DIFF
--- a/.github/workflows/pr-test-lint.yml
+++ b/.github/workflows/pr-test-lint.yml
@@ -71,7 +71,7 @@ jobs:
 
       - name: Set provider env
         run: echo "PROVIDERS_PATH=${PWD}/.providers" >> $GITHUB_ENV
-      - name: Display Provider PAth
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-test-lint.yml
```diff
@@ -67,6 +68,12 @@ jobs:
           key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go-
+
+      - name: Set provider env
+        run: echo "PROVIDERS_PATH=${PWD}/.providers" >> $GITHUB_ENV
+      - name: Display Provider PAth
```
imilchev: nit: capitalization is off

chris-rock:
```suggestion
      - name: Display Provider Path
```

PR #1067, opened 2024-01-24
✨ add benchmark test for scans

imilchev:
add benchmark testing that compares every commit in a PR with latest `main`. If the diff is above 150%, it errors out and it comments on the violating commit. Looks like this https://github.com/mondoohq/cnspec/commit/6e83272f1a7de7f8c95eb0f57e7f0767409f695a#comments

we also get a job summary like this https://github.com/mondoohq/cnspec/actions/runs/7643019168#summary-20824279571

### DIFF
--- a/.github/workflows/main-benchmark.yml
+++ b/.github/workflows/main-benchmark.yml
@@ -30,8 +30,9 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: ">=${{ env.golang-version }}"
+          cache: false
       - name: Run benchmark
-        run: make benchmark/go > benchmark.txt
+        run: make benchmark/go | tee benchmark.txt
 
         # Remove log statements and leave just the benchmark results
       - name: Cleanup benchmark file
--- a/.github/workflows/main-benchmark.yml
+++ b/.github/workflows/main-benchmark.yml
@@ -40,7 +40,7 @@ jobs:
 
       # Download previous benchmark result from cache (if exists)
       - name: Download previous benchmark data
-        uses: actions/cache@v1
+        uses: actions/cache/restore@v4
         with:
           path: ./cache
           key: ${{ runner.os }}-benchmark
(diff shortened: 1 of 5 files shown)


.github/workflows/main-benchmark.yml
```diff
@@ -0,0 +1,63 @@
+name: Benchmark main
```
chris-rock, line 1: Can we simply call this benchmark>

.github/workflows/pr-test-lint.yml
```diff
@@ -84,6 +84,49 @@ jobs:
         name: Event File
         path: ${{ github.event_path }}
 
+  go-bench:
```
chris-rock, line 94: Why do we implement this twice?

policy/scan/benchmark/benchmark_test.go
```diff
@@ -0,0 +1,121 @@
...
+	runtime := testutils.Local()
+	conf := mqlc.NewConfig(runtime.Schema(), cnquery.DefaultFeatures)
+	job := &scan.Job{
+		Inventory: &inventory.Inventory{
+			Spec: &inventory.InventorySpec{
+				Assets: []*inventory.Asset{
+					{
+						Connections: []*inventory.Config{
+							{
+								Type: "k8s",
```
chris-rock, line 38: Normally we do not have any provider installed by default. Do we not need to ensure the k8s provider is installed. If so we also would need to clean this up afterwards to not mess with other tests.

chris-rock (commented):
This is great to see the benchmark tests coming to cnspec.

policy/scan/benchmark/benchmark_test.go
chris-rock: Yeah this will fail after we merged https://github.com/mondoohq/cnspec/pull/1060

imilchev: I don't think it will fail. When we run a scan, the required provider is auto-installed if it's not present. That should be fine. Since the goal of this benchmark is not to test provider installation or uninstall, I don't think it's necessary to cleanup providers. Here we only care about the CPU and memory usage during a scan

.github/workflows/pr-test-lint.yml
imilchev: in this file we run the benchmark for a PR. It compares the results with the results of the latest commit on `main`. The implementation for the `main` branch is different because it runs the benchmark and then uploads the benchmark results in the cache, such that PRs start comparing results with that commit.

It's a slightly different flow but I couldn't combine the 2 into a single job

.github/workflows/main-benchmark.yml
imilchev: this one runs only on the `main` branch, that's why I called it `Benchmark main`

chris-rock (approved):
Great work @imilchev Would love to see this for cnquery as well.

PR #1072, opened 2024-01-26
🐛 update llx recordings to cnquery

arlimus:


imilchev (approved):

apps/cnspec/cmd/scan.go
```diff
@@ -352,7 +352,7 @@ func RunScan(config *scanConfig, scannerOpts ...scan.ScannerOption) (*policy.Rep
 		opts = append(opts, scan.WithUpstream(config.runtime.UpstreamConfig))
 	}
 	if config.runtime.Recording != nil {
```
preslavgerchev: nit: we can remove the entire `if` check, `config.runtime.Recording != nil` can never be false

arlimus: +1, :man_facepalming: overlooked that

PR #1074, opened 2024-01-26
🧹 Use azservicebus sdk instead of the go-cloud one for sending a message

preslavgerchev:
See https://github.com/google/go-cloud/issues/3378 for more details as to why.


chris-rock (approved):
Thank you @preslavgerchev

PR #1096, opened 2024-02-01
🐛 Fix resolved policy root pruning bug

jaym:
In #1036, code was added to prune nodes that were being used. However, this happens before the frameworks get a say in what happens. This meant it was possible to prune the root reporting job casuing the resolved policy to be broken

### DIFF
--- a/policy/resolver.go
+++ b/policy/resolver.go
@@ -585,6 +585,11 @@ func (s *LocalServices) tryResolve(ctx context.Context, bundleMrn string, assetF
 	reportingJobUUIDs := topologicalSortReportingJobs(collectorJob.ReportingJobs)
 	for _, rjUUID := range reportingJobUUIDs {
 		rj := collectorJob.ReportingJobs[rjUUID]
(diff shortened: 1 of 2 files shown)


policy/resolver.go
```diff
@@ -585,6 +585,11 @@ func (s *LocalServices) tryResolve(ctx context.Context, bundleMrn string, assetF
 	reportingJobUUIDs := topologicalSortReportingJobs(collectorJob.ReportingJobs)
 	for _, rjUUID := range reportingJobUUIDs {
 		rj := collectorJob.ReportingJobs[rjUUID]
+		if rj.QrId == "root" {
```
preslavgerchev, line 588: can this also be fixed by moving the pruning to after the frameworks also add their RJs?

jaym: i'm not sure. We always report frameworks. I don't know if that is an accident or explicitly wanted. If we move it, those get pruned

preslavgerchev (approved):

PR #1104, opened 2024-02-06
🧹 coordinator v2

imilchev:
requires mondoohq/cnquery#3218


imilchev:
not needed anymore. Will be solved differently

PR #1105, opened 2024-02-07
🧹 Handle unimplemented vuln report

czunker:


policy/scan/rpc_status.go
```diff
@@ -0,0 +1,36 @@
...
+	"regexp"
+	"strings"
+
+	"go.mondoo.com/ranger-rpc/codes"
+	"go.mondoo.com/ranger-rpc/status"
+)
+
+// rpcStatus tries to parse an error as a status.Status. If it fails, return a generic
+// This can help when we get an rpc error mangled through GraphQL
+func rpcStatus(err error) status.Status {
```
chris-rock, line 16: Can we use the standard method? https://github.com/mondoohq/ranger-rpc/blob/main/status/status_test.go#L120-L127

czunker: Not here. The standard method assumes it get's a status wrapped in an error. But what we get here went through GraphQL and is not a wrapped status. Because of GraphQL we loose some information. This is the only way I could find to get the information back into structured data.

chris-rock: Ok, another reason to go away from qql for this ...

czunker:
As discussed, we are not going to implement this. But will switch to proto soon.

PR #1115, opened 2024-02-14
Expand terms we flag in the spellcheck

tas50:
Flag more tech names + whitelist/blacklist


chris-rock (approved):
Great addition @tas50 Should this be part of the common spell check dict?

tas50:
You can't have a shared pattern file so we have to sync this a bit

PR #1119, opened 2024-02-15
🧹 use error reporter when running serve command

imilchev:
this will make sure we aren't calling `GetBundle` when running in `cnspec serve` mode. We don't really need it, since we only care about the errors

### DIFF
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -52,6 +52,7 @@ type LocalScanner struct {
 	// allows setting the upstream credentials from a job
 	allowJobCredentials bool
 	disableProgressBar  bool
+	reportType          ReportType
 	scanMutex           sync.Mutex
 }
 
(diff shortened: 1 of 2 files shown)


policy/scan/local_scanner.go
```diff
@@ -52,6 +52,7 @@ type LocalScanner struct {
 	// allows setting the upstream credentials from a job
 	allowJobCredentials bool
 	disableProgressBar  bool
+	reportType          ReportType
```
chris-rock, line 55: We should consider adjusting the enum https://github.com/mondoohq/cnspec/blob/5ea86601d8542258804f3d8c19493f653b4feb19/policy/scan/scan.proto#L34-L38 and make 0 Unspecified. This would ensure that full needs to be explicitly set.

policy/scan/local_scanner.go
```diff
@@ -153,6 +160,11 @@ func (s *LocalScanner) Run(ctx context.Context, job *Job) (*ScanResult, error) {
 		return nil, err
 	}
 
+	// The job report type has precedence over the global report type. The default is FULL
+	if job.ReportType == ReportType_FULL {
```
chris-rock, line 164: If we introduce 0 unspecified, only overwrite if it is not unspecified

chris-rock (commented):
Great improvement for the serve mode. I suggest we adjust the enum to make it more clear. 0 in enums should always be unspecified.

chris-rock (approved):
We handle the rename of the enum in a followup PR. Thank you @imilchev

PR #1133, opened 2024-02-22
🐛 Set asset annotations during cnspec serve.

preslavgerchev:
Requires https://github.com/mondoohq/cnquery/pull/3406 to turn green


chris-rock (approved):
Thank you @preslavgerchev

PR #1135, opened 2024-02-23
✨ Add reporting jobs for scoring queries MRNs. 

preslavgerchev:
Add reporting jobs for queries by MRN. Keep the existing ones by code id so that we have both until we transition over.

Here is a small sample policy you can verify the resolved policy with:
```
policies:
- name: testpolicy1
  uid: testpolicy1
  groups:
  - filters: return true
    checks:
    - uid: query-1
queries:
- uid: query-1
  mql: return 2 > 1
  ```

### DIFF
--- a/policy/resolver.go
+++ b/policy/resolver.go
@@ -1003,39 +1003,56 @@ func (s *LocalServices) policyGroupToJobs(ctx context.Context, group *PolicyGrou
 }
 
 func (cache *policyResolverCache) addCheckJob(ctx context.Context, check *explorer.Mquery, impact *explorer.Impact, ownerJob *ReportingJob) {
-	uuid := cache.global.relativeChecksum(check.Checksum)
-	queryJob := cache.global.reportingJobsByUUID[uuid]
+	// FIXME SOON: we want to add a RJ that goes by the check's mrn and check's codeid
+	// we use the relative checksums as uuids and the RJ's ids are either the check's mrn or codeid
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: policy/resolver.go [hunk 1]
```diff
@@ -1281,8 +1305,6 @@ func (s *LocalServices) jobsToQueries(ctx context.Context, policyMrn string, cac
 			rj := reportingjobs[i]
 			// (2) Scoring Queries handling
 			if !isDataQuery {
-				rj.QrId = codeID
```
preslavgerchev, line 1284: we now assign the code id directly as a QrId in `addCheckJob` so this is no longer needed

policy/resolver.go
```diff
@@ -1264,7 +1281,9 @@ func (s *LocalServices) jobsToQueries(ctx context.Context, policyMrn string, cac
 
 		cache.executionQueries[checksum] = executionQuery
 		executionJob.Queries[codeID] = executionQuery
-
+		// TODO: unsure if this needs to go in here.. some code uses the executionJob.Queries for different assertions
+		// so maybe we want it in here as well?
+		executionJob.Queries[query.Mrn] = executionQuery
```
preslavgerchev, line 1294: unsure if we want this here. it seems to work just fine with it and i've seen some code that iterates over that map of queries.

czunker:
We will give this a try: https://github.com/mondoohq/cnspec/pull/1344

PR #1143, opened 2024-02-28
⚙️ Add timeout when sending message to a service bus. Use a custom cred chain to workaround managed identity cred issues.

preslavgerchev:
* Set a timeout on sending the message to 30s.
 * Set the right content type if the report is JSON
 * Use a custom cred resolver. The default implementation sets the timeout to a second. I have observed random, inconsistent failures related to that when working with the default creds. Use a custom chain resolver that sets the timeout to 5 seconds instead (this should be more than enough). This resolver also works with CLI. env and workload creds, covering almost all cases

### DIFF
--- a/cli/reporter/azure_service_bus_handler.go
+++ b/cli/reporter/azure_service_bus_handler.go
@@ -5,9 +5,14 @@ package reporter
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"regexp"
 	"strings"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	pol "github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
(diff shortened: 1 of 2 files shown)


cli/reporter/azure_service_bus_handler.go
```diff
@@ -49,10 +54,17 @@ func (h *azureSbusHandler) WriteReport(ctx context.Context, report *policy.Repor
 	if err != nil {
 		return err
 	}
+
 	msg := &azservicebus.Message{
 		Body: data,
 	}
-	err = sender.SendMessage(ctx, msg, &azservicebus.SendMessageOptions{})
+	if h.format == JSON {
+		typ := "application/json"
```
chris-rock, line 62: should we set the right content type for protobuf as well?

chris-rock (approved):
Nice

cli/reporter/azure_service_bus_handler.go
preslavgerchev: i dont think we support protobuf here? the format enum doesnt define that

PR #1155, opened 2024-03-06
✨ add isLastBatch to StoreResults

imilchev:
add a flag that notifies when the last batch of results for an asset has been sent


policy/executor/internal/collector.go
```diff
@@ -86,12 +86,12 @@ func (c *BufferedCollector) run() {
 			}
 			c.lock.Unlock()
 
-			if len(results) > 0 {
-				c.collector.SinkData(results)
+			if len(results) > 0 || done {
```
czunker, line 89: More for my understanding of this logic:
This will send an empty StoreResults to the server after we saved all the scores/results?

imilchev: yes, we always need to send a `StoreResults` with `IsLastBatch = true`. If it happens to be that we have no data to send when we called `FlushAndStop` we should still make the call to notify the server we are done

czunker (approved):

PR #1170, opened 2024-03-12
🧹 only pull vuln report for AggregateReporter

imilchev:
When we do not use the `AggregateReporter` we don't care about vulnerabilities at all. This change will make sure we aren't calling `GetVulnCompactReport` when running `cnspec serve` and `cnspec serve-api`

### DIFF
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -514,7 +514,8 @@ func (s *LocalScanner) RunAssetJob(job *AssetJob) {
 	job.Reporter.AddReport(job.Asset, results)
 
 	upstream := s.upstreamServices(job.Ctx, job.UpstreamConfig)
-	if upstream != nil {
+	// The vuln report is relevant only when we have an aggregate reporter
+	if vulnReporter, isAggregateReporter := job.Reporter.(VulnReporter); upstream != nil && isAggregateReporter {
 		// get new gql client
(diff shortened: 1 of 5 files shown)


policy/scan/local_scanner.go
```diff
@@ -514,7 +514,8 @@ func (s *LocalScanner) RunAssetJob(job *AssetJob) {
 	job.Reporter.AddReport(job.Asset, results)
 
 	upstream := s.upstreamServices(job.Ctx, job.UpstreamConfig)
-	if upstream != nil {
+	// The vuln report is relevant only when we have an aggregate reporter
+	if _, isAggregateReporter := job.Reporter.(*AggregateReporter); upstream != nil && isAggregateReporter {
```
jaym: it might be cleaner to remove `AddVulnReport` from the Reporter interface (and the implementations that do not need it) and have it be part of an extension interface. That way, is someone decides to make another reported and wants it, things keep working

imilchev: good call! Updated the PR

jaym: I think you would want to remove the implementations from `reporter_error.go` and `reporter_noop.go`

imilchev: ooops... missed that. It's fixed now

jaym (approved):

PR #1220, opened 2024-04-02
⚙️ Log current time when running cnspec serve.

preslavgerchev:
Gain more insights when looking at `cnspec serve` logs on when certain actions have occurred.


jaym (approved):
you could probably also setup the logger to include the timestamp

PR #1221, opened 2024-04-03
✨ Perform check-in for the client when running a scan via cnspec serve.

preslavgerchev:
Whenever running a scan via `cnspec serve`, ensure to also check in the managed client for consistency. To achieve this:
 * Split the `checkinpinger` into `checkinhandler` and `checkinpinger`. The first one knows only how to perform a check in upstream. The latter one knows only how to run a timer-based function and to call `checkinhandler`
 * Re-use the `checkinhandler` when running the scan


arlimus (approved):
awesome to have this updated, nice find!

PR #1224, opened 2024-04-04
🐛 fix risk factors missing in inmemory w/ upstream

arlimus:
When running against upstream policies, we don't get risk factors from the policy bundle, since we generally don't need the bundle to run the scan. However, we do require basic risk factor info to help score everything before we send data up.

In this change we pull risk info from the resolved policy and inject it into the inmemory datastore before the scan is started (unless the risk factor exists for any reason, like prior policy bundles). It has enough information for scoring.

For any later reporting step we may still inject all risk factor metadata into the inmemory store before printing the output.

### DIFF
--- a/internal/datalakes/inmemory/policyresolver.go
+++ b/internal/datalakes/inmemory/policyresolver.go
@@ -615,6 +615,19 @@ func (db *Db) SetResolvedPolicy(ctx context.Context, mrn string, resolvedPolicy 
 		return errors.New("failed to save resolved policy '" + mrn + "'")
 	}
 
+	// We need the risk factors for initial reporting, but don't require all
+	// their metadata. The risk factors in the resolved policy provides everything
+	// we need for scoring. If we fetch a full risk factor we can replace it.
+	for mrn, rf := range resolvedPolicy.CollectorJob.RiskFactors {
+		rf.Mrn = mrn
+		if _, err := db.getRiskFactor(ctx, rf.Mrn); err == nil {
+			continue
+		}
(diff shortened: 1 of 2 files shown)


internal/datalakes/inmemory/policyresolver.go
```diff
@@ -615,6 +615,19 @@ func (db *Db) SetResolvedPolicy(ctx context.Context, mrn string, resolvedPolicy
 		return errors.New("failed to save resolved policy '" + mrn + "'")
 	}
 
+	// We need the risk factors for initial reporting, but don't require all
+	// their metadata. The risk factors in the resolved policy provides everything
+	// we need for scoring. If we fetch a full risk factor we can replace it.
+	for mrn, rf := range resolvedPolicy.CollectorJob.RiskFactors {
```
imilchev, line 621: doesn't that mean we can remove https://github.com/mondoohq/cnspec/blob/main/policy/hub.go#L203-L207 ?

arlimus: "For any later reporting step we may still inject all risk factor metadata into the inmemory store before printing the output."
i.e. since we only pull from the resolved policy here it only has super limited metadata. Whenever we can we want to pull from the bundle. Just in this case, we don't get a bundle until we start printing - or not at all if we are in serve mode.

arlimus: Thanks for considering it!! could have easily been a miss on my part

imilchev (approved):

PR #1246, opened 2024-04-15
⭐️ Add evidence to compliance controls.

preslavgerchev:
Adds evidence to compliance controls. The evidence struct closely resembles a policy and allows for running scans with self-contained framework files.

Internally, we unpack as following:
 * Each `Evidence` object gets its own policy with the respective checks, queries,  title and description.
 * If the uid is not present, we generate one by using the framework uid and the evidence's index in the list (to ensure no collisions) 
 * For each framework, we generate  only one framework map from its evidences. We iterate through all the controls, generating a control map for the list of evidences.


chris-rock (approved):
Thank you @preslavgerchev

PR #1248, opened 2024-04-15
🛑  report scores in numbers

chris-rock:
Initially we introduced a rating by letter that where close to US school system grades. We learned over time that users wanted to see in more details which score they have. Therefore already exposed the underlying score in v9 to display the number along with the rating. The next version is not outputting the letter grade anymore. 

**Before**

Screenshot 2024-04-15 at 21 05 49

**After**

Screenshot 2024-04-15 at 21 05 12


arlimus (approved):
Looks great! We may want to consider adding the words (critical / high / medium / low) later, but no rush right now

tas50:
@chris-rock can you take a look at the alignment now with the errors. It looks like we went from aligning the names always to only if they're scored.

chris-rock:
Have you tested it or is that from the screenshot? The screenshot is outdated because the tests cover this case.

PR #1249, opened 2024-04-15
🛑 switch to json v2 output format

chris-rock:
after https://github.com/mondoohq/cnspec/pull/1244

This PR enables the v2 json format as standard. The new format is accessible via:

```
cnspec scan local --json
```

or 

```
cnspec scan local --output json
```

For backwards-compatibility, the old format is available via:

```
cnspec scan local --output json-v1
```


arlimus (approved):
awesome!

PR #1252, opened 2024-04-16
⭐️ Improve compliance framework evidence generation.

preslavgerchev:
This is a follow-up to #1246 
 * Generate one policy per framework. Evidences are now pulled into policy groups rather than separate policies.  This means  we do not have to deal with multiple policies, specially for larger frameworks.
 * Add `GenerateEvidenceObjects` onto a framework which will pull out the evidence and return both a policy and a framework map.
 * Pull out generation code entirely into `evidence.go`. This makes it fully testable, the bundle now only calls `GenerateEvidenceObjects`.
 * Improve tests


chris-rock (approved):
Thank you @preslavgerchev

PR #1261, opened 2024-04-17
🔨 provide a way to debug memory usage per query

imilchev:
Provide a way to debug the memory usage per executed query. This is intentionally not a debug message, because we need to read extra info to be able to print it. It is also not tied to the existing `DEBUG` flag because that one is storing bundles locally and switching the log output to debug. Storing the bundles would not work in a k8s setup because we have the entire filesystem as read-only.


arlimus (approved):
helpful to find heavy allocations

PR #1268, opened 2024-04-18
add installer release workflow

lexicoder:


.github/workflows/release-installer.yml
```diff
@@ -0,0 +1,14 @@
+name: Trigger installer release
+
+on:
```
chris-rock: I think this triggers too early. It needs to trigger after all binaries have been uploaded.

lexicoder: I've modified this to run after the goreleaser job.

chris-rock (approved):

PR #1273, opened 2024-04-24
adjust the asset filter to the new style

atomic111:
adjust the filter to the new style and add some desc to the example policy


chris-rock (approved):
Thank you @atomic111

Issue #1272, opened 2024-04-23
installer trigger does not work


czunker:
**Describe the bug**
![image](https://github.com/mondoohq/cnspec/assets/827818/dbee00e3-5306-4fe4-baf8-8daa5eba8266)

https://github.com/mondoohq/cnspec/actions/runs/8799595922/job/24149395311

**Expected behavior**
The trigger should work.

**Additional context**
For other triggers, e.g. cnquery version bump in cnspec, we split this up.
The source repo triggers the target the repo:
https://github.com/mondoohq/cnquery/blob/main/.github/workflows/bump-cnspec.yml
The target repo creates the PR on event:
https://github.com/mondoohq/cnspec/blob/main/.github/workflows/cnquery-update.yml

lexicoder:
@czunker This should now have been fixed by #1275.

czunker:
Thanks, @lexicoder. Let's close it for now. We can re-open it when it fails during the next release.

PR #1275, opened 2024-04-25
change workflow being triggered for installer release

lexicoder:
Fixes mondoohq/server#7776.


.github/workflows/goreleaser.yml
```diff
@@ -171,4 +171,4 @@ jobs:
       - name: Run installer release workfow
```
benr, line 172: We shouldn't run this if 'skip_publish' is set.

See https://github.com/mondoohq/cnspec/blob/2202d3cd36d6235527b3c24b69911ba5d6a1cba1/.github/workflows/goreleaser.yml#L142

benr (approved):

lexicoder:
Fixes #1272.

PR #1277, opened 2024-04-30
🐛 fix junit reporter panic

imilchev:
The bundle can be empty if cnspec encountered an error during scanning. In that case, we need to make sure we don't exit with a panic


czunker (approved):
Thanks @imilchev

PR #1343, opened 2024-06-25
📄  Update readme with new messaging

misterpantz:


tas50 (approved):
Put those providers in spellcheck and ship ig

PR #1344, opened 2024-06-26
✨ Store query MRNs with reporting jobs

czunker:
Add all MRNs which resolv to the same code id with their reporting job.


czunker:
This would supersede https://github.com/mondoohq/cnspec/pull/1135

policy/resolver.go
```diff
@@ -1163,31 +1178,44 @@ func (cache *policyResolverCache) addCheckJobVariants(ctx context.Context, query
...
 		cache.global.codeIdToMrn[query.CodeId] = append(cache.global.codeIdToMrn[query.CodeId], query.Mrn)
-		cache.global.reportingJobsByUUID[uuid] = queryJob
-		cache.global.reportingJobsByMsum[query.Checksum] = append(cache.global.reportingJobsByMsum[query.Checksum], queryJob)
-		cache.childJobsByMrn[query.Mrn] = append(cache.childJobsByMrn[query.Mrn], queryJob)
+		cache.global.reportingJobsByUUID[uuid] = rj
+		cache.global.reportingJobsByMsum[query.Checksum] = append(cache.global.reportingJobsByMsum[query.Checksum], rj)
+		cache.childJobsByMrn[query.Mrn] = append(cache.childJobsByMrn[query.Mrn], rj)
+	}
+
+	if _, ok := cache.global.reportingJobsByCodeId[query.CodeId]; !ok {
```
czunker, line 1186: I'm not sure we also want to add it here.
Current focus might be only on checks.

policy/resolver_test.go
```diff
@@ -1451,3 +1453,54 @@ queries:
...
 	})
 }
+
+func TestResolve_TwoMrns(t *testing.T) {
+	b := parseBundle(t, `
+owner_mrn: //test.sth
+policies:
+- uid: policy1
+  groups:
+  - type: chapter
```
jaym: Can we add one more check where the code id matches but the filter does not match the asset

policy/resolver_test.go
```diff
@@ -1451,3 +1453,54 @@ queries:
...
+      mql: asset.name == props.name
+      props:
+      - uid: name
+        mql: return "definitely not the asset name"
+    - uid: check2
+      mql: asset.name == props.name
+      props:
+      - uid: name
+        mql: return "definitely not the asset name"
+`)
```
jaym, line 1536: it might be good to also have a queries example as the data queries have their reporting jobs built differently

jaym:
Do we need special handling for variants? I think I would expect that the specific query that gets executed for a variant would end up with the variant's mrn. Right now, I think these have their mrn reported in their code id field since they have not code

internal/bundle/bundle.yac.go
```diff
@@ -368,14 +368,15 @@ func (d ControlRef) MarshalYAML() (interface{}, error) {
...
-	Checks      []*Mquery   `protobuf:"bytes,21,rep,name=checks,proto3" json:"checks,omitempty" yaml:"checks,omitempty"`
-	Queries     []*Mquery   `protobuf:"bytes,22,rep,name=queries,proto3" json:"queries,omitempty" yaml:"queries,omitempty"`
-	FileContext FileContext `json:"-" yaml:"-"`
-	Comments    Comments    `json:"-" yaml:"-"`
+	Uid         string        `protobuf:"bytes,3,opt,name=uid,proto3" json:"uid,omitempty" yaml:"uid,omitempty"`
+	Mrn         string        `protobuf:"bytes,4,opt,name=mrn,proto3" json:"mrn,omitempty" yaml:"mrn,omitempty"`
+	Desc        string        `protobuf:"bytes,2,opt,name=desc,proto3" json:"desc,omitempty" yaml:"desc,omitempty"`
+	Title       string        `protobuf:"bytes,1,opt,name=title,proto3" json:"title,omitempty" yaml:"title,omitempty"`
+	Checks      []*Mquery     `protobuf:"bytes,21,rep,name=checks,proto3" json:"checks,omitempty" yaml:"checks,omitempty"`
+	Controls    []*ControlRef `protobuf:"bytes,23,rep,name=controls,proto3" json:"controls,omitempty" yaml:"controls,omitempty"`
```
czunker, line 376: I'm not sure why this is part of the PR.
I didn't change that. Perhaps `policy/generate` was missing on a previous PR.

policy/resolver_test.go
czunker: Added.

policy/resolver_test.go
czunker: Added.

czunker:
> Do we need special handling for variants? I think I would expect that the specific query that gets executed for a variant would end up with the variant's mrn. Right now, I think these have their mrn reported in their code id field since they have no code

I added a test for variants:
https://github.com/mondoohq/cnspec/pull/1344/files#diff-7ee0cbfbc4c0c6d335de85d19eb6f7e77aa8e24b223323392c43d4e7943dd696R1608

I'm not sure we need special handöing for this. This PR does not change the value of qr_id. It just adds the MRNs in any case.

policy/cnspec_policy.proto
```diff
@@ -525,6 +525,9 @@ message ReportingJob {
   map<string,bool> datapoints = 7;
   map<string,cnquery.explorer.Impact> child_jobs = 9;
   Type type = 10;
+  // We will only store a ReportingJob by codeID
+  // But we add all the MRNs to it, so we can easily retrieve the MRN for a check
```
arlimus: Love that you are adding comments! let's make sure they are more targeted to the field, eg:
"MRNs of all checks / queries that this reporting job represents"

policy/resolver.go
```diff
@@ -376,11 +376,12 @@ type resolverCache struct {
...
 	propsCache       explorer.PropsCache
 
-	reportingJobsByUUID map[string]*ReportingJob
-	reportingJobsByMsum map[string][]*ReportingJob // Msum == Mquery.Checksum, i.e. only reporting jobs for mqueries
-	reportingJobsActive map[string]bool
-	errors              []*policyResolutionError
-	bundleMap           *PolicyBundleMap
+	reportingJobsByUUID   map[string]*ReportingJob
+	reportingJobsByMsum   map[string][]*ReportingJob // Msum == Mquery.Checksum, i.e. only reporting jobs for mqueries
+	reportingJobsByCodeId map[string][]*ReportingJob // CodeId == check.CodeId
```
arlimus: replace `check` in the comment with either `Mquery` (to be consistent with the above) or query. ReportingJobs don't mind if it's a check or a data query

policy/resolver.go
arlimus: I think it should be fine, but please keep an eye open for side-effects (and if necessary fix this)

arlimus (approved):
I added a few small things to fix, feel free to get those in and then merge. The rest looks great :+1:  Thank you @czunker :)

policy/resolver.go
czunker: Thanks for the feedback, changed this.

policy/cnspec_policy.proto
czunker: Thanks for the feedback, changed this.

PR #1345, opened 2024-06-26
⭐️ Allow override of scoring system through PolicyRef

jaym:
scoring system can be set in the policy ref. If it is set in the policy ref, it overrides whatever the policy itself specifies as the scoring system


policy/resolver_test.go
```diff
@@ -185,7 +185,7 @@ policies:
 	})
 }
 
-func TestResolve_PolicyActionIgnore(t *testing.T) {
+func TestResolve_PolicyActionScoringSystem(t *testing.T) {
```
czunker, line 243: Perhaps a nit pick: L229 has the old name.

The other thing I'm not sure about:
To me this reads like mixing two cases. I don't know whether this makes sense at this point or whether this should be separate cases.

czunker (approved):

PR #1346, opened 2024-06-27
🤖 Auto approve go dep PRs by dependabot

czunker:


czunker:
The official docs suggest to put this into a separate trigger:
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#approve-a-pull-request

This is in the first commit of this PR.
But because of the nature how our tests are triggered, I had to add a sleep to ensure all tests are present. This felt wrong, so I moved the approval into the workflow. This will also help us in case we want to auto-approve other PRs later.

imilchev (approved):
LGTM! Can we also include the provider cnquery update pr that we generate on cnquery release?

PR #1350, opened 2024-07-02
🐛 fix policy scoring with exceptions

imilchev:
We ignore all scores that are out of scope, disabled or ignored from the policy score calculation


czunker (approved):
Thanks for all the tests

policy/score_calculator.go
```diff
@@ -113,13 +113,17 @@ func AddDataScore(calculator ScoreCalculator, totalDeps int, finishedDeps int) {
 
 func (c *averageScoreCalculator) Add(score *Score, impact *explorer.Impact) {
 	switch score.Type {
-	case ScoreType_Skip:
+	case ScoreType_Skip, ScoreType_Disabled, ScoreType_OutOfScope:
```
preslavgerchev, line 116: if the query is disabled, why would it even be executed and end up with a score?

PR #1352, opened 2024-07-02
Make risk magnitude a struct

jaym:


internal/bundle/yacit/main.go
```diff
@@ -17,7 +17,7 @@ import (
 func main() {
 	res := yacit.New(yacit.YacItConfig{
 		SkipUnmarshal: []string{
-			"Impact", "ImpactValue", "Filters", "Remediation",
+			"Impact", "ImpactValue", "Filters", "Remediation", "RiskMagnitude",
```
preslavgerchev, line 20: just curious: does having the field here means that it will be unmarshalled using our own `UnmarshalJSON` func?

preslavgerchev (approved):

internal/bundle/yacit/main.go
jaym: It means don't generate the unmarshal function because we will write it manually

PR #1353, opened 2024-07-02
Allow overriding risk magnitudes

jaym:
Allows overriding the risk magnitude for a risk and completely disabling a risk.

After #1352


jaym:
this is not ready yet. We need to pick the correct checks for the risk factor

policy/cnspec_policy.proto
```diff
@@ -192,8 +197,9 @@ message RiskFactor {
   repeated cnquery.explorer.Mquery checks = 2;
 
   ScopeType scope = 70;
-  float magnitude = 71;
-  bool is_absolute = 72;
+  float deprecated_v11_magnitude = 71;
+  bool deprecated_v11_is_absolute = 72;
```
czunker: nit: Could you please add a comment what to use instead?

policy/cnspec_policy.proto
```diff
@@ -645,7 +653,7 @@ message ScoreDelta {
 message ScoredRiskFactor {
   string mrn = 1;
   float risk = 2;
-  bool is_absolute = 3;
```
czunker: I know, we save a risk factor struct in the backend. Do you know whether it is this one? Would this break the unmarshal of already saved risk factors?

policy/resolver.go
```diff
@@ -710,6 +719,42 @@ func (s *LocalServices) refreshChecksums(executionJob *ExecutionJob, collectorJo
...
+}
+
+func (s *LocalServices) mergeRisk(cache *resolverCache, riskFactor *RiskFactor) {
+	if existing, ok := cache.riskFactors[riskFactor.Mrn]; ok {
+		if riskFactor.Magnitude != nil {
+			existing.Magnitude = riskFactor.Magnitude
+		}
+
+		if riskFactor.Action != explorer.Action_UNSPECIFIED {
+			existing.Action = riskFactor.Action
```
czunker, line 750: Perhaps an edge case:
Could the same risk factor be defined in multiple policies?
This line would then mean that the last non-unspecified action wins. Is this as it should be?

jaym: ya, i think that problem already exists. But if its defined in the space policy it will win

czunker (approved):

PR #1354, opened 2024-07-03
🧹 Downgrade libc to 1.53.3.

preslavgerchev:


go.mod
```diff
@@ -33,7 +33,7 @@ require (
 	github.com/spf13/pflag v1.0.6-0.20201009195203-85dd5c8bc61c
 	github.com/spf13/viper v1.19.0
 	github.com/stretchr/testify v1.9.0
-	go.mondoo.com/cnquery/v11 v11.11.0
+	go.mondoo.com/cnquery/v11 v11.10.0
```
vjeffrey: this says we also downgrade cnquery?

vjeffrey: should we release cnquery to avoid the downgrade?

preslavgerchev: oh good point, i will adjust (to a specific commit)

vjeffrey (approved):

PR #1359, opened 2024-07-08
🐛 attempt to fix banded scoring with exceptions

imilchev:
Second attempt at fixing banded scoring, after #1350 didn't fully work

### DIFF
--- a/policy/score_calculator.go
+++ b/policy/score_calculator.go
@@ -67,28 +67,6 @@ func AddSpecdScore(calculator ScoreCalculator, s *Score, found bool, impact *exp
 		return
 	}
 
-	// everything else is modify or activate
-
-	if impact.Scoring == explorer.ScoringSystem_IGNORE_SCORE {
-		calculator.Add(&Score{
-			// We override the type because:
-			// 1. If it is set to Result, its value will be added to the total
-			// calculation in most calculators despite its weight.
-			// 2. We don't want to set it to unscored, because technically we
-			// just ignore the score.
-			// Thus we set the score to unknown for the sake of the calculator,
-			// thus it knows it is handling a scored result, but also knows not
-			// to count it.
-			Type:            ScoreType_Unknown,
-			Value:           score.Value,
-			Weight:          0,
-			ScoreCompletion: score.ScoreCompletion,
-			DataCompletion:  score.DataCompletion,
-			DataTotal:       score.DataTotal,
-		}, nil)
-		return
-	}
-
 	if impact.Weight > 0 {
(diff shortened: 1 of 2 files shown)


policy/score_calculator.go
```diff
@@ -490,22 +468,22 @@ func (c *bandedScoreCalculator) Add(score *Score, impact *explorer.Impact) {
 
 			if category <= 10 {
 				c.critMax += score.Weight
-				if score.Value == 100 {
+				if score.Value < 100 {
```
imilchev, line 471: based on my understanding the counter should only increase if the score is not `Pass`

policy/score_calculator_test.go
```diff
@@ -239,7 +239,24 @@ func TestBandedScores(t *testing.T) {
...
+				{Value: 100, ScoreCompletion: 100, DataCompletion: 33, DataTotal: 3, Weight: 10, Type: ScoreType_Result},
+			},
+			impacts: []*explorer.Impact{
+				// 10 critical checks
+				{Value: &explorer.ImpactValue{Value: 100}},
+				{Value: &explorer.ImpactValue{Value: 100}},
+				// 10 high checks
+				{Value: &explorer.ImpactValue{Value: 80}},
+			},
+			out: &Score{Value: 100, ScoreCompletion: 100, DataCompletion: 66, Weight: 20, Type: ScoreType_Result},
```
imilchev, line 259: I added a case with all passing checks. I would expect if everything passed, then the policy score should also be 100. In the original code that is not the case. It works now, however, the values for the other cases changed too...

policy/score_calculator.go
```diff
@@ -67,28 +67,6 @@ func AddSpecdScore(calculator ScoreCalculator, s *Score, found bool, impact *exp
 		return
 	}
 
-	// everything else is modify or activate
-
-	if impact.Scoring == explorer.ScoringSystem_IGNORE_SCORE {
```
imilchev, line 72: in #1350 I already made sure all the calculators handle exceptions properly, so this workaround shouldn't be needed anymore

policy/score_calculator.go
jaym: based on the `Calculate` function, I think `c.crit` and the rest are supposed to count the number of failures:
```
		if c.crit != 0 {
			res.Value = uint32(math.Floor(float64(50) * pcrCrit))
		} else if c.high != 0 {
			res.Value = uint32(math.Floor(float64(50)*pcrHigh)) + 10
```
If there failures, we need to fall into the first category with failures.

So this all seems right to me

jaym (approved):

PR #1376, opened 2024-07-18
🧹 update mql syntax

chris-rock:
Updates MQL queries to use the new syntax

### DIFF
--- a/internal/bundle/testdata/failing_lint.mql.yaml
+++ b/internal/bundle/testdata/failing_lint.mql.yaml
@@ -10,7 +10,7 @@ policies:
       - checks:
           - uid: sshd-01
             title: Ensure the port is set to 22
-            mql: sshd.config.params["Port"] == 22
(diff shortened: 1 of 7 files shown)


internal/bundle/testdata/failing_lint.mql.yaml
```diff
@@ -10,7 +10,7 @@ policies:
       - checks:
           - uid: sshd-01
             title: Ensure the port is set to 22
-            mql: sshd.config.params["Port"] == 22
```
tas50, line 13: Do we want to keep the old syntax in a test somewhere to make sure it continues to function?

chris-rock: Those files are not used for testing. Tests are in https://github.com/mondoohq/cnquery/blob/main/mqlc/mqlc_test.go#L1607-L1639

arlimus (approved):
nice, new syntax!

PR #1377, opened 2024-07-19
🧹 support MONDOO_OUTPUT as env variable

chris-rock:
Add support to set the output as env variable:

```bash
export MONDOO_OUTPUT=nodata,nocontrols
cnspec scan local  
```


scottford-io (approved):
Looks good @chris-rock 

Thank you!

PR #1380, opened 2024-07-22
✨ lint error when query is used both as check and data query

imilchev:


internal/bundle/lint.go
```diff
@@ -414,6 +417,20 @@ func lintFile(file string) (*Results, error) {
 				uid := check.Uid
 				updateAssignedQueries(check, assignedQueries, globalQueriesByUid)
 
+				checks[check.CodeId] = struct{}{}
+				if _, ok := dataQueries[check.CodeId]; ok {
```
imilchev: at this point in the code, `CodeId` isn't set. If we move this logic to outside of this function, the file context is missing...

If we want to implement this error, we will need to use this in combination with the ocmpiled bundle map. The issue is that the bundlemap indexes queries by MRN, but this struct has only UIDs. There is no easy way to match which query gets which mrn/codeid

internal/bundle/lint.go
```diff
@@ -414,6 +417,20 @@ func lintFile(file string) (*Results, error) {
 				uid := check.Uid
 				updateAssignedQueries(check, assignedQueries, globalQueriesByUid)
 
+				checks[check.Uid] = struct{}{}
```
imilchev, line 425: we only check by uid since this is under user control. It is still possible to cause weirdness when the MQL is identical for a query and a check but that's a bug in the resolved policy. It's not something to be caught by the linter

internal/bundle/lint.go
```diff
@@ -34,6 +34,7 @@ const (
 	queryTitle                 = "query-name"
 	queryUidUnique             = "query-uid-unique"
 	queryUnassigned            = "query-unassigned"
+	queryUsedAsDifferentTypes  = "query-used-as-different-types"
```
preslavgerchev, line 37: i think you need to update the `LinterRules` var with this entry too so that it shows in the sarif report

imilchev: good point! I missed that

preslavgerchev (approved):

PR #1385, opened 2024-07-25
Critical: Fix banded scoring within individual bands

arlimus:
Variations within bands were inverted and led to incorrect results on the edges of those variations i.e. if all scores within a band were failing or only one was failing. The tests reflect have been added to test out this behavior of variation within bands and will be further expanded in the future.

### DIFF
--- a/policy/score_calculator_test.go
+++ b/policy/score_calculator_test.go
@@ -4,6 +4,7 @@
 package policy
 
 import (
+	"fmt"
 	"strconv"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


policy/score_calculator_test.go
```diff
@@ -404,3 +405,129 @@ func TestImpact(t *testing.T) {
...
+		assert.Equal(t, 55, int(res[1].Value))
+		res = testEachScoreCalculator(0, 10, 5, 10, 0, 10, 0, 10, t)
+		assert.Equal(t, 35, int(res[1].Value))
+		res = testEachScoreCalculator(0, 10, 10, 10, 0, 10, 0, 10, t)
+		assert.Equal(t, 10, int(res[1].Value))
+	})
+}
+
+// This is a toolchain that can be used to see what the scoring mechanism would produce under different scenarios
+func TestScoreMechanisms(t *testing.T) {
```
arlimus, line 513: This has been added to have a playground for experimenting with scoring mechanisms. Uncomment the last line to force the failure to be printed

jaym (approved):

PR #1392, opened 2024-07-31
🧹use slice filter to list frameworks

slntopp:


upstream/framework.gql.go
```diff
@@ -20,7 +20,7 @@ type UpstreamFramework struct {
 	State mondoogql.ComplianceFrameworkState
 }
 
-func ListFrameworks(ctx context.Context, c *gql.MondooClient, scopeMrn string, state *mondoogql.ComplianceFrameworkState) ([]*UpstreamFramework, error) {
+func ListFrameworks(ctx context.Context, c *gql.MondooClient, scopeMrn string, states *[]mondoogql.ComplianceFrameworkState) ([]*UpstreamFramework, error) {
```
preslavgerchev: this signature looks a bit weird, why not just pass a slice instead of a ptr to a slice?

slntopp: Just what mondoogql expects. afaik if we make slice argument mandatory, then you'd have to send empty slice from the frontend to meet the schema

preslavgerchev: Can we make this an implementation detail then? We can still pass the slice to the actual GQL call, but i think the arg in this function should be a `states []mondoogql...`.

slntopp: Updated the implementation

preslavgerchev (approved):

PR #1398, opened 2024-08-08
🐛 make sure reporting job mrns work correctly for variants

imilchev:
The issue with the current code is that all variant queries would get all MRN of other variant queries. That is because we index the mrn by codeid and variant queries all have the same codeid since they have no mql block. This PR makes sure a variant query gets the mrn of itself only


arlimus (approved):
Nice catch @imilchev , Kudos!

PR #1408, opened 2024-08-23
🐛 Find variant checks by parent MRN

czunker:
Compliance only knows about parent MRNs and not variant MRNs. To allow a mapping from compliance control to executed check, we need to store the check with it's parent MRN.

### DIFF
--- a/policy/resolver.go
+++ b/policy/resolver.go
@@ -8,6 +8,7 @@ import (
 	"encoding/base64"
 	"encoding/binary"
 	"math/rand"
+	"slices"
 	"sort"
 	"time"
 
@@ -1160,10 +1161,21 @@ func (cache *policyResolverCache) addCheckJob(ctx context.Context, check *explor
 			Type:       ReportingJob_CHECK,
 			Mrns:       []string{},
 		}
-		// We don't track the MRNs for variant queries through the cachem, since variant queries
+		// We don't track the MRNs for variant queries through the cache, since variant queries
 		// have no MQL, meaning they get the same code ID
 		if len(check.Variants) == 0 {
 			cache.global.codeIdToMrn[check.CodeId] = append(cache.global.codeIdToMrn[check.CodeId], check.Mrn)
+			// Because we have no variants, we might have to add the MRN of the parent job
+			if ownerJob != nil && ownerJob.Type == ReportingJob_CHECK && len(ownerJob.Mrns) > 0 {
+				// We might have variants, where multiple variant queries apply
+				// Don't save the same MRNs multiple times
+				for _, mrn := range ownerJob.Mrns {
+					if slices.Contains(cache.global.codeIdToMrn[check.CodeId], mrn) {
+						continue
+					}
+					cache.global.codeIdToMrn[check.CodeId] = append(cache.global.codeIdToMrn[check.CodeId], mrn)
+				}
+			}
 		}
 		cache.global.reportingJobsByUUID[uuid] = rj
 		cache.global.reportingJobsByMsum[check.Checksum] = append(cache.global.reportingJobsByMsum[check.Checksum], rj)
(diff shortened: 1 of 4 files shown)


policy/resolver.go
```diff
@@ -649,21 +650,30 @@ func (s *LocalServices) tryResolve(ctx context.Context, bundleMrn string, assetF
...
+
 	resolvedPolicy := ResolvedPolicy{
-		GraphExecutionChecksum: resolvedPolicyExecutionChecksum,
-		Filters:                matchingFilters,
-		FiltersChecksum:        assetFiltersChecksum,
-		ExecutionJob:           executionJob,
-		CollectorJob:           collectorJob,
-		ReportingJobUuid:       reportingJob.Uuid,
+		GraphExecutionChecksum: rpChecksumInclJobs.String(),
+		// GraphExecutionChecksum: resolvedPolicyExecutionChecksum,
```
czunker: This one was only influenced by the policy or a framework:
https://github.com/mondoohq/cnspec/blob/main/policy/resolver.go#L479

And was then used as is for the resolved policy.

policy/resolver.go
```diff
@@ -649,21 +650,30 @@ func (s *LocalServices) tryResolve(ctx context.Context, bundleMrn string, assetF
...
 	}
 
+	// phase 6: refresh all checksums
+	// This uses the ReportingJobs checksums, so calculate them first.
+	s.refreshChecksums(executionJob, collectorJob)
+
+	// resolvedPolicyExecutionChecksum is the GraphExceutionChecksum of the policy and the framework
+	// it does not change if any of the jobs changes, only if the policy or the framework changes
+	rpChecksumInclJobs := checksums.New
+	rpChecksumInclJobs.Add(resolvedPolicyExecutionChecksum) // nolint: staticcheck
```
jaym: something here seems wrong to me. checksums.add returns the new value but you never use it

czunker: Fixed in the new calculation

policy/resolver.go
```diff
@@ -657,8 +662,15 @@ func (s *LocalServices) tryResolve(ctx context.Context, bundleMrn string, assetF
 		rj.RefreshChecksum()
 	}
 
+	// resolvedPolicyExecutionChecksum is the GraphExceutionChecksum of the policy and the framework
```
jaym: This doesn't feel right. There is a function called `BundleExecutionChecksum`. I think this manipulation needs to be done there. The graph execution checksum is passed into this function so we can see if we have something for it or it needs to be recomputed. To me, that means all graph execution checksums need to be calculated the same way

czunker: `BundleExecutionChecksum` is called on line L484. But it does not know about reporting jobs or any other job.

Everything about jobs is kept out of any checksum up to this point. Even the checksums of the reporting jobs get refreshed after the jobs are included in a checksum refresh. L661-L663 vs. L658.

The reporting jobs' checksums are included in `refreshChecksums` on L710, but at this point they are empty.

I don't know whether this is a bug or intent.

czunker: As just discussed, I moved this into the `BundleExecutionChecksum`:
https://github.com/mondoohq/cnspec/pull/1408/files#diff-01d77c6835a56d0f0539b547e1b48fe04335555a2e6809f62ad629dff934bb21R102-R106

policy/resolver_test.go
```diff
@@ -234,9 +234,9 @@ policies:
 		require.NoError(t, err)
 		require.NotNil(t, rp)
 		require.Len(t, rp.CollectorJob.ReportingJobs, 5)
-		ignoreJob := rp.CollectorJob.ReportingJobs["8Sis0SvMbtI="]
+		ignoreJob := rp.CollectorJob.ReportingJobs["q7gxFtwx4zg="]
```
czunker, line 237: These UUIDs changed, because the new BundleExecutionChecksum is included in the UUIDs: https://github.com/mondoohq/cnspec/blob/main/policy/resolver.go#L536

jaym (approved):

PR #1431, opened 2024-09-13
✨ Add a source to scores

czunker:


policy/cnspec_policy.proto
```diff
@@ -640,6 +640,9 @@ message Score {
   int64 failure_time = 11;
   // risk factors that were applied to this score
   ScoredRiskFactors risk_factors = 12;
+  // Source of the score
+  // Currently, this only applies to vulnerability scores
+  string source = 13;
```
chris-rock: I have questions:

1. One score can have multiple sources. Eg. MVD can find a CVE and MS defender can find a CVE. Then what lands in the source
2. I am not sure a simple string is enough. On GraphQL layer we need to return: source id, source title, source icon. How do we achieve that without storing all the data all the time. eg. we track a list of data sources per space and then refer ids?

chris-rock: I recommend we do the following:

```
// Source is used to identify the source.
message Source {
  // Required. Name of the source
  string name = 1;
  // Optional. URL of the source
  string url = 2;
}
```

czunker: Thanks for the suggestion. I changed it accordingly.

chris-rock (approved):
Thank you @czunker

PR #1440, opened 2024-09-25
🐛 Chunk sending results by size

jaym:
There is an upper limit to what is accepted by the platform per request. Make sure we do not go past it when sending data.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	github.com/spf13/pflag v1.0.6-0.20201009195203-85dd5c8bc61c
 	github.com/spf13/viper v1.19.0
 	github.com/stretchr/testify v1.9.0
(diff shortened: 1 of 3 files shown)


imilchev (commented):
LGTM! I think we have similar code in cnquery though. It may be a better idea to define the chunking there so we can re-use it in both cnquery and cnspec

jaym:
Do you have specific code in mind?

imilchev:
Here https://github.com/mondoohq/cnquery/blob/045aa0ee10614422610a855314de93aca6841cda/mql/internal/collector.go#L31

jaym:
https://github.com/mondoohq/cnquery/pull/4711

imilchev (approved):
LGTM!

PR #1461, opened 2024-10-28
✨ experimental: Add policy migrations to bundles

arlimus:
This contains their metadata only, not surrounding functionality at this point. It gives us a container to store migrations between versions of policies and collections, to retain information about what was added, or removed, or just modified (eg UIDs changing between versions).


arlimus:
@preslavgerchev This change sets up initial migration structure and enforces a few changes from my earlier design. It is necessary for some of the other work we are doing behind the scenes.

policy/cnspec_policy.proto
```diff
@@ -147,6 +147,41 @@ message Bundle {
...
+// to new policies.
+message Migration {
+  enum Action {
+    UNSPECIFIED = 0;
+    CREATE = 1;
+    REMOVE = 2;
+    MODIFY = 3;
+  }
+
+  MigrationMatch match = 1;
```
preslavgerchev, line 174: whats the match for?

policy/cnspec_policy.proto
```diff
@@ -147,6 +147,41 @@ message Bundle {
...
+  // from older versions or other collections.
+  // Experimental as of v11.x
+  repeated MigrationGroup migration_groups = 11;
+}
+
+// MigrationGroup contains a set of migrations for a policy collection.
+// The policy should generally contain the UID and affected version.
+message MigrationGroup {
+  Policy policy = 1; // only informational fields will be used (version)
+  repeated Migration migrations = 2;
```
preslavgerchev, line 160: is the intend here to descrieb migrations within the policy, i.e. queries being removed?

policy/cnspec_policy.proto
arlimus: These are the conditions on which a check is matched for a migration. Imagine throwing all checks at a migration, when this condition is true (i.e. when it matches) then the appropriate action is run.

policy/cnspec_policy.proto
arlimus: Yes, and the 3 actions are: removal, creation, and modification (as seen in the enum)

preslavgerchev (approved):

PR #1465, opened 2024-10-31
Rewrite policy resolver

jaym:
Things that are known to be different from the old implementation:
- all queries are reported by code id. and these reporting jobs report to the mrn
- Only the first matching variant is selected
- A reporting job is created for a query by code id and mrn. The code id reporting job notifies the mrn based one
  - controls have the code id one as their child job so as not to be affected by impact. We should add a scoring system for controls so this hackery is not needed
- Data queries in a policy have their impact set to ignore score
- There is one global impact for each query. The worst one is chosen and used everywhere
- Anything that is not connected to an execution query is not added to the resolved policy. This will get rid of cases where we have policies / controls  with no queries, frameworks with no controls


policy/resolved_policy_builder.go
```diff
@@ -0,0 +1,1132 @@
...
+	if !canRun(action) {
+		return false
+	}
+
+	// Create a node for the framework, but only if its a valid framework mrn
+	// Otherwise, we have the asset / space policies which we will connect
+	// to. We need to do this because we cannot have a space frame and space
+	// policy reporting job because they would have the same qr id.
+	// TODO: we should create a new reporting job type for asset and space
+	// reporting jobs so its cleare that we can connect both frameworks and
```
imilchev:
```suggestion
	// reporting jobs so its clear that we can connect both frameworks and
```

imilchev (commented):
Tested it and seems to work fine. Only added 1 typo comment

imilchev (approved):

policy/resolver_v2_test.go
czunker, line 1: Wow, a lot of tests. Thanks!

policy/resolver_v2_test.go
```diff
@@ -0,0 +1,1900 @@
...
+	rpTester.ReportingJobByMrn(queryMrn("query-2")).Notifies(policyMrn("testpolicy1"))
+
+	rpTester.ReportingJobByMrn(policyMrn("testpolicy1")).Notifies("root")
+
+	rpTester.doTest(t, rp)
+
+	require.Equal(t, float32(0.9), rp.CollectorJob.RiskFactors[riskFactorMrn("sshd-service")].Magnitude.GetValue())
+}
+
+func TestResolveV2_FrameworkExceptions(t *testing.T) {
```
czunker, line 1684: Would it make sense to also add a test case for exceptions for a policy/check?

policy/resolver_v2_test.go
```diff
@@ -0,0 +1,1900 @@
...
+	rpTester.DoesNotExecutesQuery(queryMrn("unix-uname"))
+	rpTester.DoesNotExecutesQuery(queryMrn("check-os-unix"))
+
+	rpTester.CodeIdReportingJobForMrn(queryMrn("windows-uname")).Notifies(queryMrn("windows-uname"))
+	rpTester.ReportingJobByMrn(queryMrn("windows-uname")).Notifies(queryMrn("uname"))
+	rpTester.ReportingJobByMrn(queryMrn("uname")).Notifies("root")
+
+	rpTester.CodeIdReportingJobForMrn(queryMrn("check-os-windows")).Notifies(queryMrn("check-os-windows"))
+	rpTester.ReportingJobByMrn(queryMrn("check-os-windows")).Notifies(queryMrn("check-os"))
+	rpTester.ReportingJobByMrn(queryMrn("check-os")).Notifies("root")
```
czunker, line 1578: Does this mean, we can later query the result of a variant check by the actually executed check (`check-os-windows`) and the variant MRN (`check-os`)?

policy/resolver_v2_test.go
```diff
@@ -0,0 +1,1900 @@
...
+	b := parseBundle(t, `
+owner_mrn: //test.sth
+policies:
+- owner_mrn: //test.sth
+  mrn: //test.sth
+  groups:
+  - policies:
+    - uid: policy-active
+    - uid: policy-ignored
+      action: 4
```
czunker, line 592: Would a test with a preview policy help besides all the other tests you already wrote?

policy/resolver.go
```diff
@@ -491,6 +503,20 @@ func (s *LocalServices) tryResolve(ctx context.Context, bundleMrn string, assetF
...
+		if err != nil {
+			return nil, err
+		}
+
+		err = s.DataLake.SetResolvedPolicy(ctx, bundleMrn, resolvedPolicy, V2Code, false)
+		if err != nil {
+			return nil, err
+		}
+
+		return resolvedPolicy, nil
```
czunker, line 517: Definitely not a blocker, more like personal preference:

On first read, I skipped the `return` statement and was expecting an `else`.
I'm not sure whether it is actually possible, but would it be possible to choose between current and ng resolver outside of the `tryResolve`?
As said, more like a personal preference, feel free to ignore it.

policy/resolved_policy_builder.go
```diff
@@ -0,0 +1,1132 @@
...
+
+// isGroupMatching checks if the policy group is matching. A policy group is matching if it is not rejected,
+// and it is not expired. If it has filters, it must have at least one filter that matches the asset filters
+func (b *resolvedPolicyBuilder) isGroupMatching(group group) bool {
+	if group.GetReviewStatus() == ReviewStatus_REJECTED {
+		return false
+	}
+
+	if group.GetEndDate() != 0 {
+		// TODO: we also need to check if the group is accepted or rejected
```
czunker: Is this something that needs to happen now, before this is merged?

policy/cnspec_policy.proto
```diff
@@ -554,12 +554,16 @@ message ReportingJob {
...
   enum Type {
     UNSPECIFIED = 0;
-    CHECK = 1;
-    DATA_QUERY = 2;
     CONTROL = 3;
     POLICY = 4;
     FRAMEWORK = 5;
     RISK_FACTOR = 6;
+
+    // DO NOT USE CHECK OR DATA_QUERY, THEY ARE DEPRECATED
```
czunker, line 564: What should be used?
Or what is the equivalent to `CONTROL` for a policy?

policy/resolver_v2_test.go
jaym: yes, ill add those

policy/resolver_v2_test.go
jaym: we report a result for both and their results are the same. Do you think this covers the usecase you're thinking of? if you search for `check-os`, you will get the result we store for `check-os`. If you search for `check-os-windows`, you will get the result we store for `check-os-windows`

policy/resolver_v2_test.go
jaym: correct me if i'm mistaken, but I think preview and ignored are the same

policy/resolved_policy_builder.go
jaym: comment is wrong, it happens in the lines above. will remove

policy/cnspec_policy.proto
jaym: i've left it as UNSPECIFIED. it's not used by anything

policy/resolver_v2_test.go
czunker: Sounds good to me. That should cover the use case where we didn't find the result when querying for `check-os`. Thanks.

policy/resolver_v2_test.go
czunker: You are correct.

jaym:
A bug i've found so far:
- risk factors check scores are sent. This ends up causing errors like `Couldn't find any queries for incoming value for BP5NURsd9Pw`. There is some code in the cnspec execution that tries to determine if a non risk tried to run the query and only then sends it up. However that algorithm is no longer working

PR #1469, opened 2024-11-04
⭐️ new `cnspec integrate` experimental command

afiune:
This is an **EXPERIMENTAL** command.

Our first iteration of an automated way to create integrations in the Mondoo platform. ✨

The first integration we are enabling is Azure, to do so, the user can run:
```
cnspec integrate azure --space <the_space_to_create_the_integration>
```

The flow to create an integration is:

* Verify we have access to the Mondoo platform
* Verify we can access Azure
* Detect subscriptions to integrate
* More validation like, integration name, etc.
* Generation of automation code
* Write code to disk
* Run automation code with confirmation from user


vjeffrey:
not a blocker at all for this pr, but just curious, why not read the space from the config if it exists/if the user doesnt include the space in the command

vjeffrey (approved):
🧁 this is awesome!

i was able to run the command and see all the resources tf was gonna generate. 
i later failed on an error,
```
Error: retrieving directory role for template ID "88d8e3e3-8f55-4a1e-953a-9b9898b8876b": result was nil
```
not sure if you know what that means? i'm assuming it's a perms thing i dont have access to 

in any case, this seems to be working well!

afiune:
@vjeffrey nice suggestion! I'll add loading the space from the config with a TODO check if it is a service account, which is not really critical since the provider will fail already with:
```
x exit status 1

Error: Failed to create Mondoo client

  with provider["registry.terraform.io/mondoohq/mondoo"],
  on main.tf line 10, in provider "mondoo":
  10: provider "mondoo" {

valid service account needs to be provided
```

But I'd like to fail faster, not when we are executing the automation already.

PR #1472, opened 2024-11-05
✨ policy last applied field

slntopp:


policy/cnspec_policy.proto
```diff
@@ -114,6 +116,8 @@ message Policy {
   string graph_content_checksum = 38;
   string local_execution_checksum = 39;
   string graph_execution_checksum = 40;
+  // timestamp when the policy was last applied
+  int64 last_applied = 48;
```
jaym: I don't think this one makes sense. I would only expect this information in `PolicyRef`. What's the thought for how this one is used?

jaym (approved):

PR #1477, opened 2024-11-12
⭐️ allow and deny subscriptions + scan VMs options

afiune:
This PR adds three more flags to the experimental command `cnspec integrate azure`.

`--allow`, `--deny` used to control the integration of subscriptions. By default, all subscriptions will be discovered and integrated for continuous scanning.

To choose the subscriptions to scan, pass the list of subscriptions using the `--allow` flag.

	cnspec integrate azure --allow <subscription_id_1> --allow <subscription_id_2>

To scan all subscriptions expect those you specify, pass the list of subscriptions you don't
want Mondoo to scan using the `--deny` flag.

	cnspec integrate azure --deny "<subscription_id_1>,<subscription_id_2>"

The new `--scan-vms` enables scanning VMs on Azure subscriptions via RunCommand.

**NOTE** that --allow and --deny are mutually exclusive and can't be use together.

Finally, I moved the `onboarding` code that is generic to its own package so that new integration commands like `gcp` or `aws` can leverage it.

### DIFF
--- a/internal/onboarding/exec.go
+++ b/internal/onboarding/exec.go
@@ -83,58 +83,47 @@ func getOutputDirPath(location string, identifier string) (string, error) {
 	)), nil
 }
 
-// Execute a terraform plan & apply
-func TerraformPlanAndExecute(workingDir string) error {
+// Execute a terraform plan & apply.
+//
+// It returns a true which indicates that terraform apply was executed, this is useful
+// for callers to display a success message or not
+func TerraformPlanAndExecute(workingDir string) (applied bool, err error) {
 	// Ensure Terraform is installed
 	tf, err := LocateOrInstallTerraform(false, workingDir)
 	if err != nil {
-		return err
+		return
 	}
-
-	vw := terminal.NewVerboseWriter(10)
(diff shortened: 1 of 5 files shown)


internal/onboarding/exec.go
```diff
@@ -0,0 +1,480 @@
...
+
+	vw = terminal.NewVerboseWriter(10)
+	tf.SetStdout(vw)
+	tf.SetStderr(vw)
+
+	// Apply plan
+	if err := TerraformExecApply(tf); err != nil {
+		return errors.New(provideGuidanceAfterFailure(err, tf.WorkingDir(), tf.ExecPath()))
+	}
+	vw.Close()
```
jaym: is it an issue if these are not closed on error?

jaym (approved):

internal/onboarding/exec.go
afiune: Good question.

There is no issue, but I think it is nice to clear the terminal since we are extracting the error and presenting it to the user, this way it won't be duplicate.

PR #1480, opened 2024-11-14
✨ Support multiple sources for a score

czunker:


policy/cnspec_policy.proto
```diff
@@ -966,10 +969,36 @@ message PurgeAssetsConfirmation {
   map<string, string> errors = 2;
 }
 
+// Sources is a list of sources.
+// Used to get past marshal/unmarshal limitations of repeated Source fields.
```
afiune, line 979: Approaching this very naive; what limitation is that?

czunker: You cannot call `proto.Marshal(...)` or unmarshal on a `repeated ...` field directly.

afiune (approved):

PR #1487, opened 2024-11-20
🧹 decouple checkin from service startup

imilchev:
Make sure the check-in is not blocking service startup. It now happens in the background

### DIFF
--- a/apps/cnspec/cmd/backgroundjob/checkin_pinger.go
+++ b/apps/cnspec/cmd/backgroundjob/checkin_pinger.go
@@ -43,7 +43,9 @@ func (c *checkinPinger) Start() {
 	}
 
 	// run check-in once on startup
-	runCheckIn()
+	go func() {
+		runCheckIn()
+	}()
 
 	jitter := time.Duration(rand.Int63n(int64(c.interval)))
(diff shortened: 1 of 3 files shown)


apps/cnspec/cmd/backgroundjob/serve_windows.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 func Serve(timer time.Duration, splay time.Duration, handler JobRunner) {
-	isIntSess, err := svc.IsAnInteractiveSession()
+	isIntSess, err := svc.IsWindowsService()
```
imilchev, line 20: this call is deprecated so I replaced it with its successor

apps/cnspec/cmd/backgroundjob/checkin_pinger.go
```diff
@@ -43,7 +43,9 @@ func (c *checkinPinger) Start() {
 	}
 
 	// run check-in once on startup
-	runCheckIn()
+	go func() {
+		runCheckIn()
```
chris-rock, line 47: do we need to protect us from running multiple of those go routines in parallel? This should always only run once, we do not need to queue

Issue #1485, opened 2024-11-20
`cnspec integrate` preflight check


afiune:
When running `cnspec integrate azure` we should verify that the user has the right permissions
to create the integration.

https://github.com/mondoohq/cnspec/blob/main/apps/cnspec/cmd/integrate.go#L151-L152

The user should have either `Privileged Role Administrator` or optionally `Global Administrator`

https://learn.microsoft.com/en-us/entra/identity/role-based-access-control/permissions-reference#privileged-role-administrator

PR #1494, opened 2024-11-22
⭐️ integrate preflight check: role assignments

afiune:
This change adds a minimalistic preflight check where we verify that the user has the right role assignments to onboard an Azure environment.

We need one of "**Privileged Role Administrator**" or "**Global Administrator**"

https://learn.microsoft.com/en-us/entra/identity/role-based-access-control/permissions-reference#privileged-role-administrator

Note that we run Azure CLI commands directly so that `cnspec` doesn't have to import the entire Azure SDK which will make the binary much bigger, which today is unnecessary for the amount of checks we are running.

## Example Success

Screenshot 2024-11-22 at 2 58 09 PM

## Example Failure

Screenshot 2024-11-22 at 3 02 27 PM

### DIFF
--- a/apps/cnspec/cmd/integrate.go
+++ b/apps/cnspec/cmd/integrate.go
@@ -148,8 +148,11 @@ NOTE that --allow and --deny are mutually exclusive and can't be use together.`,
 				os.Exit(1)
 			}
 
-			// TODO ideally, we should verify that the user has the "Privileged Role Administrator" or "Global Administrator"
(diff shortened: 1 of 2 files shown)


jaym (approved):

afiune:
Closes https://github.com/mondoohq/cnspec/issues/1485

chris-rock:
Woot. Thank you @afiune great addition!

PR #1496, opened 2024-11-25
🧹 Make resolved policy execution use reporting queries

jaym:
The execution was previously implicitly tying together reporting queries and reporting jobs. This now uses the ReportingQueries section of the CollectorJob to do this.

Also, part of this change, I've introduced an additional reporting job types. One represents the executed query(reporting job by code id). The other represents a thing that is both a check and a query. This was helpful in fixing an issue where we don't properly forward a score and intead rescore it. This would turn skipped scores into unscored.

This last point is one of those non-backwards compatible changes. Once the new resolved policy is used everywhere, this bug will be seen in old clients. I don't see a way to fix it in a backwards compatible way, and its a small enough issue that I think its fine.

Note: While this should work with the old resolved policy, its probably best to wait to release it with the resolved policy rewrite.


policy/executor/internal/builder.go
```diff
@@ -21,6 +21,7 @@ type query struct {
 	codeBundle         *llx.CodeBundle
 	requiredProps      map[string]string
 	resolvedProperties map[string]*llx.Primitive
+	notifies           []string
```
jaym, line 24: this gets filled in with the information from the reporting queries section

policy/executor/internal/builder.go
```diff
@@ -192,14 +194,21 @@ func (b *GraphBuilder) Build(runtime llx.Runtime, assetMrn string) (*GraphExecut
...
 		if canRun {
 			ge.addExecutionQueryNode(queryID, q, q.resolvedProperties, b.datapointType)
 		} else {
 			unrunnableQueries = append(unrunnableQueries, q)
 		}
-		ge.addReportingQueryNode(queryID, q)
+		n := ge.addReportingQueryNode(queryID, q)
+		if len(q.notifies) > 0 {
+			for _, notify := range q.notifies {
+				ge.addEdge(n.id, NodeID(notify))
```
jaym: creates the edge from the node we use to represent a reporting query to the node we use to represent the reporting job. This will be the reporting job for query code id

jaym: Also note that the reporting query node doesn't fully match whats in the resolved policy. We create one of these for every query, scoring or data. Only scoring queries are in the reporting queries section of the resolved policy.

policy/executor/internal/nodes.go
```diff
@@ -522,21 +522,32 @@ func (nodeData *ReportingJobNodeData) score() (*policy.Score, error) {
...
-
-		if nodeData.childScores[nodeData.queryID] == nil {
-			panic("invalid reporting job")
-		}
+	if nodeData.forwardScore {
 		var s *policy.Score
-		if v := nodeData.childScores[nodeData.queryID].score; v == nil {
+
+		if len(nodeData.childScores) == 0 {
 			s = &policy.Score{
```
jaym, line 529: This is needed because reporting query nodes in our graph don't have children if they're data queries only.

policy/executor/internal/nodes.go
```diff
@@ -522,21 +522,32 @@ func (nodeData *ReportingJobNodeData) score() (*policy.Score, error) {
...
+			for k := range nodeData.childScores {
+				child = k
+			}
+			if v := nodeData.childScores[child].score; v == nil {
+				s = &policy.Score{
+					QrId: nodeData.queryID,
+					Type: policy.ScoreType_Result,
+				}
+			} else {
+				s = proto.Clone(v).(*policy.Score)
```
imilchev: not sure how often this gets called but we had some issues with memory/cpu consumption a while back. A lot of those were caused by `proto.Clone`. It is doing a lot of allocations and reflection magic which is super inefficient. If we need to clone this structure often, it's best to manually write a function that does that

imilchev (approved):

policy/executor/internal/nodes.go
jaym: i can change this in a follow up pr. Would like to avoid it being part of this one

PR #1514, opened 2024-12-11
🐛 Use 100 for the default value of impact

jaym:
It's possible for impact to be be nil, in which case we should use the default value which is 100. Using the score value doesn't make sense. It would mean that the category would change based on if it passed or failed in cases where impact was not defined.

### DIFF
--- a/policy/score_calculator_test.go
+++ b/policy/score_calculator_test.go
@@ -276,6 +276,23 @@ func TestBandedScores(t *testing.T) {
 			},
 			out: &Score{Value: 9, ScoreCompletion: 100, DataCompletion: 66, Weight: 20, Type: ScoreType_Result},
 		},
+		{
+			in: []*Score{
+				// 10 critical checks (9ok, 1not)
+				{Value: 0, ScoreCompletion: 100, DataCompletion: 80, DataTotal: 5, Weight: 1, Type: ScoreType_Result},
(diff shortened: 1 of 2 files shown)


policy/score_calculator_test.go
```diff
@@ -276,6 +276,23 @@ func TestBandedScores(t *testing.T) {
...
+				{Value: 100, ScoreCompletion: 100, DataCompletion: 33, DataTotal: 3, Weight: 10, Type: ScoreType_Result},
+			},
+			impacts: []*explorer.Impact{
+				// 10 critical checks
+				{Value: &explorer.ImpactValue{Value: 100}},
+				{Value: nil},
+				// 10 high checks
+				{Value: &explorer.ImpactValue{Value: 80}},
+			},
+			out: &Score{Value: 45, ScoreCompletion: 100, DataCompletion: 66, Weight: 20, Type: ScoreType_Result},
```
czunker, line 294: I don't understand this part of the test.

The previous test (L273) uses impact 100. The PR description states that we use 100 for impact, when it is nil.
Shouldn't the scores be the same in the two tests? L277 vs. L294

jaym: i copied the test on 243. This one is slightly different than the one that is directly before. score values 0, 100, 0 vs 0, 100, 100

czunker: Ah, I see. Thanks.

czunker (approved):
Thanks @jaym

PR #1520, opened 2024-12-19
🤖 Automatically create GH release after cnquery bump

czunker:


.github/workflows/gh-release.yaml
```diff
@@ -0,0 +1,93 @@
...
+        uses: actions/checkout@v4
+      - name: Set release version
+        run: echo "RELEASE_VERSION=$(cat VERSION)" >> $GITHUB_ENV
+      - name: Release
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: ${{ env.RELEASE_VERSION }}
+          generate_release_notes: true
+          make_latest: true
+          token: ${{ secrets.GH_BUILDER_TOKEN }}
```
czunker: I'm not sure about this token...

czunker:
This is based on the workflow we already use for the packer-plugin-release: https://github.com/mondoohq/packer-plugin-cnspec/blob/main/.github/workflows/gh-release.yaml

.github/workflows/gh-release.yaml
```diff
@@ -0,0 +1,93 @@
...
+on:
+  push:
+    paths:
+      - "VERSION"
+    branches:
+      - main
+  workflow_dispatch:
+
+env:
+  # C07QZDJFF89 == #release-coordination
```
czunker, line 13: Has the slack token privilege to write to this channel?

.github/workflows/gh-release.yaml
czunker: As discussed:
I switched to the app token, as we already use it cnspec:
https://github.com/mondoohq/cnspec/blob/2b0dd2aa8e343409ca5c33a309da4105521fea6b/.github/workflows/pr-test-lint.yml#L138

.github/workflows/gh-release.yaml
philipbalinov: yes, slack token has privileges for the whole org

philipbalinov (approved):
LGTM, thank you @czunker!

PR #1525, opened 2024-12-27
🚨 recover from panic's inside goroutines

afiune:
The main cnspec goroutine has already a `recover()` statement, though Go doesn't
share that between goroutines and since the calculation of the policies happen on
another thread, we were not reporting upstream.


czunker (approved):
LGTM
Thanks @afiune

PR #1527, opened 2024-12-27
🐛 fix CI test

afiune:
This changed since the `github` provider changed.

We still want to test that flags are loaded, so I am updating the test and not removing it.


czunker (approved):
Thanks @afiune

PR #1545, opened 2025-01-14
🧹 Don't write report collection for debug output

jaym:
I added this because our tests loaded these from a file. However, the more I think about it, its probably not a good idea to do this because it can get really large


jaym:
> > ... our tests loaded these from a file.
> 
> Do we need a new way to generate these for test cases?

There was no documented way and this is what I did with this line of code

czunker (approved):

PR #1547, opened 2025-01-15
✨ Trigger installer release

czunker:


.github/workflows/goreleaser.yml
```diff
@@ -150,6 +150,30 @@ jobs:
...
+          # error on HTTP code different to 200
+          command: |
+            vSEMVER=${{  github.ref_name }}
+            SEMVER="${vSEMVER//v}"
+            curl -o /dev/null -s -w "%{http_code}\n" "https://releases.mondoo.com/cnspec/${SEMVER}/cnspec_${vSEMVER}_SHA256SUMS" | grep 200          
+
+      - name: Trigger installer release
+        uses: peter-evans/repository-dispatch@v3
+        with:
+          token: ${{ secrets.RELEASR_ACTION_TOKEN }}
```
czunker: I doubt this is correct.
Which token should we use here?

philipbalinov: I suggest you use the `actions/create-github-app-token` step as in https://github.com/mondoohq/installer/pull/508/files ; i've set the necessary secrets/apps permissions already

czunker: Thanks. I changed it and moved it into a separate job, so that the other depended job does not fail in case of an error.

PR #1551, opened 2025-01-21
📝 report when query takes longer than 5 min to execute

vjeffrey:
requires https://github.com/mondoohq/cnquery/pull/5107

### DIFF
--- a/policy/executor/internal/execution_manager.go
+++ b/policy/executor/internal/execution_manager.go
@@ -173,6 +173,10 @@ func (em *executionManager) executeCodeBundle(codeBundle *llx.CodeBundle, props 
 			Str("qrid", codeID).
 			Dur("duration", time.Since(startTime)).
 			Msg("finished query execution")
+		if time.Since(startTime) > 5*time.Minute {
+			// if the query duration was more than 5 minutes, send an alert to platform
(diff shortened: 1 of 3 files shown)


policy/executor/internal/execution_manager.go
```diff
@@ -173,6 +173,10 @@ func (em *executionManager) executeCodeBundle(codeBundle *llx.CodeBundle, props
 			Str("qrid", codeID).
 			Dur("duration", time.Since(startTime)).
 			Msg("finished query execution")
+		if time.Since(startTime) > 5*time.Minute {
+			// if the query duration was more than 5 minutes, send an alert to platform
+			health.ReportSlowQuery("cnspec", cnspec.Version, cnspec.Build, health.SlowQueryInfo{Query: codeID, Duration: time.Since(startTime)})
```
jaym, line 178: if we have the mql, it might make life easier

jaym (approved):

policy/executor/internal/execution_manager.go
vjeffrey: i couldnt find where to access the mql string from this context, but ya that would be awesome if we can

PR #1558, opened 2025-01-28
✨ add specific score type for snoozed checks

imilchev:
Report scores for snoozed checks with score type `Snoozed`


policy/score.go
```diff
@@ -16,6 +16,7 @@ const (
 	ScoreType_Unscored
 	ScoreType_OutOfScope
 	ScoreType_Disabled
+	ScoreType_Snoozed
```
jaym, line 19: do you need to modify the function below?

imilchev: whoops... I probably should

PR #1561, opened 2025-01-29
🐛 Fix broken links

czunker:
https://github.com/mondoohq/cnspec/actions/runs/13004679114/job/36269400650

Looks like the link checker cannot follow redirects.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -6,20 +6,20 @@
 
 Before building from source, be sure to install:
 
-- [Go 1.19.0+](https://golang.org/dl/)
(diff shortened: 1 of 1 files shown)


preslavgerchev (approved):

docs/development.md
```diff
@@ -6,7 +6,7 @@
 
 Before building from source, be sure to install:
 
-- [Go 1.19.0+](https://golang.org/dl/)
+- [Go 1.19.0+](https://go.dev/dl/)
```
preslavgerchev: nit: do we need to bump the golang ver here?

czunker: Good call. IIRC, at some point, we starting depending on 1.22.

PR #1587, opened 2025-02-25
🐛 fix panics

imilchev:
Make sure we return a proper error when delayed discovery detected no platform. This is not expected but currently not sure how to reproduce it. We should at least make sure we don't panic.
Also fix another panic when framework map is `nil`


claytoncasey01 (approved):
LGTM

PR #1589, opened 2025-02-28
🧹 Report asset data on panic

czunker:


policy/scan/local_scanner.go
```diff
@@ -418,7 +421,47 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
...
+				}
+
+				serviceAccount := opts.GetServiceCredential()
+				if serviceAccount == nil {
+					log.Error().Msg("no service account configured")
+					return
+				}
+
+				tags := map[string]string{
+					"assetMrn":             batch[0].Asset.Mrn,
```
vjeffrey: do we have a guarantee that batch will be of length more than 0 ?

czunker: Good call.
I'm quite sure this is present. The batch is created from the discovered assets, and we exit early when didn't discover anything. But better be safe than sorry.

PR #1598, opened 2025-03-11
🐛 set viper output flag

chris-rock:
Fixes the `cnspec vuln --output json`

### DIFF
--- a/apps/cnspec/cmd/vuln.go
+++ b/apps/cnspec/cmd/vuln.go
@@ -20,14 +20,15 @@ import (
 func init() {
 	rootCmd.AddCommand(vulnCmd)
 	vulnCmd.Flags().StringP("output", "o", "full", "Set output format: "+reporter.AllFormats())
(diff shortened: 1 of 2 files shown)


apps/cnspec/cmd/vuln.go
```diff
@@ -20,25 +20,27 @@ import (
 func init() {
 	rootCmd.AddCommand(vulnCmd)
 	vulnCmd.Flags().StringP("output", "o", "full", "Set output format: "+reporter.AllFormats())
-	vulnCmd.Flags().BoolP("json", "j", false, "Run the query and return the object in a JSON structure.")
```
afiune, line 23: I think it is ok to remove this flag since it never worked, the only think I can think of is that, even though it was not working, someone is using it in CI. If so and we remove it, jobs will fail, but it is an edge case so, let's move faster and remove it!

afiune (approved):

apps/cnspec/cmd/vuln.go
chris-rock: Agreed, would be a different discussion for the `scan` command.

PR #1604, opened 2025-03-18
🧹 Fix deprecated/removed/moved parameters for golangci

czunker:
This fixes https://github.com/mondoohq/cnspec/actions/runs/13894237360/job/38871321104\?pr\=1603

These are the same fixes as for cnquery.


czunker:
https://github.com/mondoohq/cnquery/blob/main/.github/.golangci.yml#L5-L12

PR #1608, opened 2025-03-20
:broom: Fix Linter, missing testcase added

mm-weber:


internal/bundle/lint.go
```diff
@@ -387,31 +387,44 @@ func lintFile(file string) (*Results, error) {
...
-				// if the filters are set within the group
-				len(group.Policies) == 0 && !checksHaveFiltersOrVariants(group, globalQueriesByUid) {
-				location := Location{
-					File:   file,
-					Line:   group.FileContext.Line,
-					Column: group.FileContext.Column,
-				}
+			// issue warning if no filters are assigned but do not show the warning if
+			// the group has policies
+			if (group.Filters == nil || len(group.Filters.Items) == 0) && len(group.Policies) == 0 {
```
jaym: I dont think the check on `len(group.Polilcies)` is needed here

afiune: fixed!

PR #1620, opened 2025-04-07
✨ extend policy linter to enforce no unknown fields in yaml

imilchev:
We found issues where remediations are not correctly set in policies. Seems like our linter does not enforce no unknown fileds to be present in the bundle. This PR changes that so now we can get:
```
cnspec policy lint mondoo-aws-security.mql.yaml
→ lint policy bundle file=mondoo-aws-security.mql.yaml
  RULE ID         LEVEL  FILE                          LINE  MESSAGE                         
  bundle-invalid  error  mondoo-aws-security.mql.yaml  1     cannot parse the yaml file      
                                                             mondoo-aws-security.mql.yaml:   
                                                             error unmarshaling JSON: while  
                                                             decoding JSON: json: unknown    
                                                             field "remediation"             
FTL invalid policy bundle
```

One limitation of this approach is that we do not get the exact line of the issue. This is better than nothing. A better solution that would yield the correct line numbers would be to use yamlv3. However, for that we need to re-implement all the custom unmarshallers we have for existing types to work with yamlv3.

### DIFF
--- a/internal/bundle/fmt.go
+++ b/internal/bundle/fmt.go
@@ -13,6 +13,7 @@ import (
 	"github.com/rs/zerolog/log"
 	"go.mondoo.com/cnspec/v11/policy"
 	"gopkg.in/yaml.v3"
+	k8sYaml "sigs.k8s.io/yaml"
 )
 
 // Formats the given bundle to a yaml string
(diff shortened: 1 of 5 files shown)


internal/bundle/fmt.go
```diff
@@ -54,7 +54,10 @@ func FormatRecursive(mqlBundlePath string, sort bool) error {
 func ParseYaml(data []byte) (*Bundle, error) {
 	baseline := Bundle{}
 
-	err := yaml.Unmarshal([]byte(data), &baseline)
+	err := yaml.Unmarshal(data, &baseline)
+	if err != nil {
+		return nil, err
+	}
 	return &baseline, err
```
glower: we could return `&baseline, nil` here to be more clear

imilchev: oh, I actually didn't mean to change this at all. I will make sure there is no change to this file and we should be good to go. Good catch!

PR #1644, opened 2025-05-01
🐛 onboarding: honor allowed subs when scan VMs in enabled

afiune:
When the user selects both, a list of subscriptions to integrate via the `--allow` flag, and scan VMs on those subscriptions via the `--scan-vms` flag, the generated code didn't respect the list of subscriptions and tried to grant access to all available subs.

This change fixes that!

### DIFF
--- a/internal/onboarding/azure.go
+++ b/internal/onboarding/azure.go
@@ -57,22 +57,6 @@ func GenerateAzureHCL(integration AzureIntegration) (string, error) {
 	if err != nil {
 		return "", errors.Wrap(err, "failed to generate self signed cert subject block")
 	}
-	customRolePermissionsBlock, err := tfgen.HclCreateGenericBlock("permissions", nil,
-		tfgen.Attributes{
-			"actions ": []string{
-				"Microsoft.Compute/virtualMachines/runCommands/read",
-				"Microsoft.Compute/virtualMachines/runCommands/write",
-				"Microsoft.Compute/virtualMachines/runCommands/delete",
-				"Microsoft.Compute/virtualMachines/runCommand/action",
-			},
-			// not_actions = []
-			// data_actions = []
-			// not_data_actions = []
-		},
-	)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/onboarding/azure.go
```diff
@@ -245,41 +206,151 @@ func GenerateAzureHCL(integration AzureIntegration) (string, error) {
...
+	var resourceRMCustomRoleDefinition *tfgen.HclResource
+	if integration.ScanVMs {
+		customRolePermissionsBlock, err := tfgen.HclCreateGenericBlock("permissions", nil,
+			tfgen.Attributes{
+				"actions ": []string{
+					"Microsoft.Compute/virtualMachines/runCommands/read",
+					"Microsoft.Compute/virtualMachines/runCommands/write",
+					"Microsoft.Compute/virtualMachines/runCommands/delete",
+					"Microsoft.Compute/virtualMachines/runCommand/action",
+				},
```
czunker: We have this block of permissions twice now. Here and at L239.

I understand why you moved it, but perhaps keep at least the list of permissions in one place.

internal/onboarding/azure_test.go
```diff
@@ -192,41 +189,150 @@ resource "time_sleep" "wait_time" {
...
+`
+	assert.Equal(t, expected, code)
+}
 
-resource "azurerm_role_assignment" "mondoo_security" {
-  count              = length(data.azurerm_subscriptions.available.subscriptions)
-  principal_id       = azuread_service_principal.mondoo.object_id
-  role_definition_id = azurerm_role_definition.mondoo_security.role_definition_resource_id
-  scope              = data.azurerm_subscriptions.available.subscriptions[count.index].id
+func TestGenerateAzureHCLScanVMsForAllSubscriptions(t *testing.T) {
```
czunker, line 369: Would be nice to also have a test case like "TestGenerateAzureHCLScanVMsForAllowedSubscriptions" with `integration.Allow` being set.

PR #1652, opened 2025-05-09
Notify when a policy checksum must be recalculated

jaym:
The execution checksum needs to change as time passes a groups start and end. This change returns to the caller when the checksum should be recomputed


imilchev (approved):
LGTM

PR #1653, opened 2025-05-12
Improve the linter

syrull:
I've rewrote the linter, still a draft

### DIFF
--- a/internal/bundle/checks_core.go
+++ b/internal/bundle/checks_core.go
@@ -1,0 +1,165 @@
+// internal/bundle/checks_core.go
+// Copyright (c) Mondoo, Inc.
+// SPDX-License-Identifier: BUSL-1.1
+
+package bundle
+
+import (
+	"regexp"
+)
+
+const (
+	LevelError   = "error"
+	LevelWarning = "warning"
+)
+
+// reResourceID: lowercase letters, digits, dots or hyphens, fewer than 200 chars, more than 5 chars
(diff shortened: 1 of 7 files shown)


internal/bundle/testdata/fail_PolicyMissingChecksRuleID.yaml
```diff
@@ -0,0 +1,209 @@
+
+policies:
```
chris-rock: Names should not be use different styles. Here we use snake casing with camel casing

internal/bundle/testdata/fail_PolicyMissingChecksRuleID.yaml
```diff
@@ -0,0 +1,209 @@
+
```
chris-rock: yaml files should be formatted

chris-rock: all files should be ending with `.mql.yaml` and then formatted with `cnspec policy fmt`

internal/bundle/testdata/fail-policy-missing-checks.mql.yaml
```diff
@@ -0,0 +1,189 @@
+policies:
```
chris-rock, line 1: Most of the test data is not used, we should strip the test files to the specific test we are doing

chris-rock (approved):

PR #1655, opened 2025-05-12
🐛 Policy list -a limit fix

claytoncasey01:
This fixes the issue where `cnspec policy list -a` only returns 200 items, it will now return all of them. 

#### Current Master
Screenshot 2025-05-12 at 11 57 03 AM

#### This PR
Screenshot 2025-05-12 at 11 57 21 AM

### DIFF
--- a/upstream/gql.go
+++ b/upstream/gql.go
@@ -36,26 +36,26 @@ func SearchPolicy(
 	var policies []*UpstreamPolicy
 	var after *mondoogql.String = nil
 
-	for {
-		var q struct {
-			Content struct {
-				TotalCount int `json:"totalCount"`
-				Edges      []struct {
-					Cursor string `json:"cursor"`
-					Node   struct {
-						Policy struct {
-							Mrn        string
-							Name       string
-							TrustLevel mondoogql.TrustLevel
-							Action     mondoogql.PolicyAction
-							Assigned   bool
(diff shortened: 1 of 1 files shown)


upstream/gql.go
```diff
@@ -33,57 +33,76 @@ func SearchPolicy(
...
+							Action     mondoogql.PolicyAction
+							Assigned   bool
+						} `graphql:"... on Policy"`
+					} `json:"node"`
+				} `json:"edges"`
+				PageInfo PageInfo `json:"pageInfo"`
+			} `graphql:"content(input: $input)"`
+		}
+
+		input := mondoogql.ContentSearchInput{
```
preslavgerchev: can we build this outside the for loop? it now gets built for each iteration

we only need to adjust the cursor here afaik

claytoncasey01: Good catch, I will get that updated.

preslavgerchev: can we also build the actual input object outside? the only thing that changes inside is the after so we'd only need this in the loop:
```
// Check if there are more pages
if !q.Content.PageInfo.HasNextPage {
	break
}

// Set cursor for next page
afterCursor := mondoogql.String(q.Content.PageInfo.EndCursor)
if afterCursor != nil {
   input.After = afterCursor
```

preslavgerchev (approved):

PR #1685, opened 2025-06-05
Enable runtime auto-update configuration for LocalScanner

glower:
This PR introduces a new ScannerOption, scan.WithRuntimeAutoUpdate, that allows consumers to pass a providers.UpdateProvidersConfig struct. This configuration is then applied to the LocalScanner's runtime instance after its default initialization.

### DIFF
--- a/apps/cnspec/cmd/serve_api.go
+++ b/apps/cnspec/cmd/serve_api.go
@@ -70,7 +70,13 @@ var serveApiCmd = &cobra.Command{
 			Creds:       serviceAccount,
 		}
 
+		enabledAutoUpdate := true
+		if viper.IsSet("auto_update") {
+			enabledAutoUpdate = viper.GetBool("auto_update")
+		}
+
 		scanner := scan.NewLocalScanner(
+			scan.WithAutoUpdate(enabledAutoUpdate),
 			scan.WithUpstream(&upstreamConfig),
 			scan.DisableProgressBar(),
 			scan.WithRecording(recording.Null{}),
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -45,6 +45,10 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+const (
+	defaultRefreshInterval = 3600
+)
+
 type LocalScanner struct {
 	resolvedPolicyCache *inmemory.ResolvedPolicyCache
 	queue               *diskQueueClient
@@ -59,6 +63,8 @@ type LocalScanner struct {
 	allowJobCredentials bool
 	disableProgressBar  bool
 	reportType          ReportType
+	autoUpdate          *bool
+	refreshInterval     int
 }
 
 type ScannerOption func(*LocalScanner)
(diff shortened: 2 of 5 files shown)


policy/scan/local_scanner.go
```diff
@@ -114,6 +114,17 @@ func NewLocalScanner(opts ...ScannerOption) *LocalScanner {
 	return ls
 }
 
+// WithRuntimeAutoUpdate configures the AutoUpdate settings for the runtime.
+func WithRuntimeAutoUpdate(cfg providers.UpdateProvidersConfig) ScannerOption {
+	return func(s *LocalScanner) {
+		if r, ok := s.runtime.(*providers.Runtime); ok {
```
imilchev: I think we should not do this type conversion. Instead, we should extend the `Runtime` interface, so we can call a function here.

test/providers/scan_flags_test.go
```diff
@@ -73,4 +74,17 @@ func TestScanFlags(t *testing.T) {
 			"could not read private key", // expected! it means we loaded the flags
 		)
 	})
+
+	t.Run("output contains no log messages", func(t *testing.T) {
```
glower, line 78: I'm unsure how to create a unit test to verify that auto_update is correctly set for the provider. The log entry for auto-updating the provider's version is generated only when a real provider exists, a scan is running, and the provider's version is older than the latest version.
It should be part of integration test I guess

apps/cnspec/cmd/serve_api.go
```diff
@@ -70,7 +70,13 @@ var serveApiCmd = &cobra.Command{
 			Creds:       serviceAccount,
 		}
 
+		enabledAutoUpdate := true
+		if viper.IsSet("auto_update") {
+			enabledAutoUpdate = viper.GetBool("auto_update")
+		}
+
 		scanner := scan.NewLocalScanner(
```
imilchev, line 78: why don't you just pass the auto-update config as input to `NewLocalScanner`. I can se that inside of it we create the runtime and you can specify the value in there directly.

imilchev: it seems like cnquery does things differently than cnspec. That's also what we discussed with @afiune . Why not work on making these 2 behave in the same way instead of making them differ even more

imilchev: Even if you decide to go for this approach, it seems like we use the `DefaultRuntime` here. You would be editing the default runtime which is a global variable and that's probably not what we want

glower: > why don't you just pass the auto-update config as input to NewLocalScanner. I can se that inside of it we create the runtime and you can specify the value in there directly.

Is this what you mean? Than we don't need to have a new function for the Runtime, only to use the correct setting here:
```
func (c *coordinator) newRuntime() *Runtime {
	res := &Runtime{
		coordinator:     c,
		providers:       map[string]*ConnectedProvider{},
		recording:       recording.Null{},
		shutdownTimeout: defaultShutdownTimeout,
		AutoUpdate: UpdateProvidersConfig{
			Enabled: true,
		},
	}
```

glower: Actually we don't need to do here any changes, `AutoUpdate` will be propagated correctly 
```
func (c *coordinator) NewRuntimeFrom(parent *Runtime) *Runtime {
	res := c.NewRuntime()
	res.UpstreamConfig = parent.UpstreamConfig
	res.AutoUpdate = parent.AutoUpdate
	res.recording = parent.Recording()
	for k, v := range parent.providers {
		res.providers[k] = v
	}
	return res
}
```

policy/scan/local_scanner.go
```diff
@@ -93,8 +93,9 @@ func WithReportType(reportType ReportType) ScannerOption {
 	}
 }
 
-func NewLocalScanner(opts ...ScannerOption) *LocalScanner {
+func NewLocalScanner(autoUpdateConfig providers.UpdateProvidersConfig, opts ...ScannerOption) *LocalScanner {
```
afiune: The main problem we had when I did this change https://github.com/mondoohq/cnquery/pull/5510 was that I wasn't aware that `cnspec` was using the local scanner in a custom way, I think instead of modifying this function with more options, we should go back to `cnquery` and add a `ScannerOption` that can be passed to set the runtime to auto-update. (without knowing that it is possible 😂 we will tackle that in cnquery)

afiune:
Closing in favor of https://github.com/mondoohq/cnspec/pull/1730

PR #1692, opened 2025-06-10
🧹 Set job specific provider path

czunker:
That should fix: https://github.com/mondoohq/cnspec/actions/runs/15560718430/job/43813252563\?pr\=1691


czunker:
Seems, this also does not help:
https://github.com/mondoohq/cnspec/actions/runs/15563269941/job/43820841936?pr=1692

PR #1706, opened 2025-06-18
Remove more references to cnspec-policies

tas50:
I noticed this when writing opps for policy management


mm-weber (approved):
LGTM, thank you @tas50 !

PR #1707, opened 2025-06-19
✨ add exception as part of score stats

kkereziev:


policy/cnspec_policy.proto
```diff
@@ -777,6 +777,7 @@ message ScoreStats {
   int64 oldest_scan_time = 11;
   // Newest scan in the set of assets observed
   int64 newest_scan_time = 12;
+  uint64 exceptions = 16;
```
imilchev, line 783: add comment to this to explain what it is. Also add comment to snoozed to explain what it is. It is no longer snoozed findings, instead it is counting all the ignored scores (which are multiple types now)

imilchev (approved):

PR #1712, opened 2025-06-24
✨ Return asset annotations in SynchronizeAssets

jaym:
This allows annotations to be set on the server side and be synced with the client so that mql `asset.annotations` includes those values

### DIFF
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -384,6 +384,16 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 						asset.Mrn = details.AssetMrn
 						asset.Url = details.Url
 						asset.Labels["mondoo.com/project-id"] = details.ProjectId
+
+						if asset.Annotations == nil {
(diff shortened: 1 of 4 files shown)


policy/scan/local_scanner.go
```diff
@@ -384,6 +384,16 @@ func (s *LocalScanner) distributeJob(job *Job, ctx context.Context, upstream *up
 						asset.Mrn = details.AssetMrn
 						asset.Url = details.Url
 						asset.Labels["mondoo.com/project-id"] = details.ProjectId
+
+						if asset.Annotations == nil {
+							asset.Annotations = make(map[string]string)
+						}
+						for k, v := range details.Annotations {
+							if _, ok := asset.Annotations[k]; !ok {
```
czunker, line 392: I couldn't find details about this.
It is intended that client-side annotations take precedence?

czunker (approved):
LGTM

policy/scan/local_scanner.go
jaym: Yes, if its set on the client it takes precedence

PR #1719, opened 2025-07-02
🧹 use CloneVT instead of proto.Clone

imilchev:
The `CloneVT` function is much more performant than the `proto.Clone`. I replaced a few calls I found around the codebase. Nothing changes functionally here


chris-rock (approved):
Thank you @imilchev

PR #1722, opened 2025-07-07
🐛 onboarding: ms365 does not depend on a subscription

afiune:
Microsoft 365 integrations only require creating resources in Entra ID, therefore
we do not have a dependency on any subscription to create resources. They all
get created at the Tenant level.

### DIFF
--- a/internal/onboarding/ms365_test.go
+++ b/internal/onboarding/ms365_test.go
@@ -12,9 +12,8 @@ import (
 
 func TestGenerateMs365HCL_Basic(t *testing.T) {
 	code, err := subject.GenerateMs365HCL(subject.Ms365Integration{
-		Name:    "test-ms365-integration",
-		Space:   "space-xyz",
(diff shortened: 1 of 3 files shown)


internal/onboarding/ms365_test.go
```diff
@@ -207,7 +196,7 @@ resource "mondoo_integration_ms365" "this" {
     pem_file = join("\n", [tls_self_signed_cert.credential.cert_pem, tls_private_key.credential.private_key_pem])
   }
   depends_on = [azuread_service_principal.mondoo, azuread_application_certificate.mondoo, azuread_directory_role_assignment.global_reader]
-  name       = "subscription-"
+  name       = "subscription-88afcea5fb91"
```
czunker, line 199: Do we have a different prefix for this?
Now that this is no longer linked to a subscription, we might want to avoid the work here.

afiune: Good call, I will update it in a diff PR. Thanks @czunker

PR #1725, opened 2025-07-08
Correct benchmark workflow to compare PRs against main

glower:
The previous benchmark workflow was incorrectly using the GitHub Actions
cache to store and retrieve historical data. This led to unreliable
comparisons, as pull requests were being compared against the most
recent cache entry, not necessarily the baseline from the `main` branch.

https://github.com/benchmark-action/github-action-benchmark/tree/master


imilchev (approved):
LGTM! Let's give this a try

PR #1730, opened 2025-07-09
Enable runtime auto-update configuration for LocalScanner

glower:
This PR introduces a new ScannerOption, scan.WithRuntimeAutoUpdate, that allows consumers to pass a providers.UpdateProvidersConfig struct. This configuration is then applied to the LocalScanner's runtime instance after its default initialization.

This is clean copy from this PR: https://github.com/mondoohq/cnspec/pull/1685

### DIFF
--- a/policy/scan/benchmark/benchmark_test.go
+++ b/policy/scan/benchmark/benchmark_test.go
@@ -64,7 +64,10 @@ func BenchmarkScan_SingleAsset(b *testing.B) {
 
 	job.Bundle = bundle
 
-	scanner := scan.NewLocalScanner(scan.DisableProgressBar())
+	scanner := scan.NewLocalScanner(
+		scan.DisableProgressBar(),
+		scan.WithRuntime(runtime.(*providers.Runtime)),
+	)
 	b.ResetTimer()
 
 	for i := 0; i < b.N; i++ {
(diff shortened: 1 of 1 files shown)


policy/scan/local_scanner.go
```diff
@@ -59,6 +63,8 @@ type LocalScanner struct {
 	allowJobCredentials bool
 	disableProgressBar  bool
 	reportType          ReportType
+	autoUpdate          *bool
```
preslavgerchev: why not just set it automatically to true, you do not need a ptr then

policy/scan/local_scanner.go
```diff
@@ -111,6 +132,27 @@ func NewLocalScanner(opts ...ScannerOption) *LocalScanner {
...
+	if ls.runtime == nil {
+		runtime := providers.DefaultRuntime()
+		refreshInterval := defaultRefreshInterval
+		if ls.refreshInterval > 0 {
+			refreshInterval = ls.refreshInterval
+		}
+
+		// By default, auto-update is enabled. It can be explicitly disabled
+		// by passing WithAutoUpdate(false)
+		autoUpdate := true
```
preslavgerchev: if you assign this default value when creating the scanner struct, you will not need this logic

glower: right :) fixed

afiune (approved):

PR #1744, opened 2025-07-25
Update remediation content

tas50:
- Fix formatting
- Expand remediations

### DIFF
--- a/content/mondoo-linux-workstation-security.mql.yaml
+++ b/content/mondoo-linux-workstation-security.mql.yaml
@@ -171,36 +171,46 @@ queries:
         The grub files contain information on boot settings and passwords for unlocking boot options.
 
         If the system uses UEFI, /boot/efi is a vfat filesystem. The vfat filesystem itself doesn't have the concept of permissions but can be mounted under Linux with whatever permissions desired.
-      remediation: |-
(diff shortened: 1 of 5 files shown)


content/mondoo-linux-workstation-security.mql.yaml
```diff
@@ -171,36 +171,46 @@ queries:
...
-        chown root:root /boot/grub2/grubenv
-        chmod og-rwx /boot/grub2/grubenv
+            ```bash
+            chown root:root /boot/grub/grub.cfg
+            chmod og-rwx /boot/grub/grub.cfg
 
-        chown root:root /boot/grub2/user.cfg
-        chmod og-rwx /boot/grub2/user.cfg
-        ```
+            chown root:root /boot/grub/menu.lst
```
philipbalinov, line 185: the `/boot/grub/` vs `/boot/grub2/` differentiation is more about non-redhat vs redhat-based distros. We could potentially combine both into a single command e.g.


`chown root:root /boot/{grub,grub2}/grub.cfg`

tas50: That fails on Debian:

```
root@timnas:/boot# chown root:root /boot/{grub,grub2}/grub.cfg
chown: cannot access '/boot/grub2/grub.cfg': No such file or directory
```

PR #1753, opened 2025-07-30
Linux policy: More remediations + remove 32bit support

tas50:
Add more remediations + simplify checks by removing support for legacy 32 bit systems


mm-weber (approved):
LGTM @tas50 !

PR #1761, opened 2025-08-02
Improve logic in sysctl remediations + improve auditd remediation

tas50:
I noticed a lot of our steps didn't work for my Debian system because the config was already set in the /etc/sysctl.conf file. If it's there let's make sure to edit it. If not then we can create the .d file

I also confirmed that a reload of auditd does not load the main config file so if you change that you need to do a full restart.

Finally I nuked some sudo usage in our remediations so things are consistent


mm-weber (approved):
LGTM thank you @tas50 !

PR #1770, opened 2025-08-06
Improve remediations in the Linux policy

tas50:
Cleanup the remediations to follow a more standard pattern


mm-weber (approved):
LGTM, thanks @tas50 !

PR #1786, opened 2025-08-15
Document the different sshd service name on Debian/Ubuntu

tas50:
Still needs to make it into the automated scripts, but this gets us closer


mm-weber (approved):
LGTM

PR #1793, opened 2025-08-18
⭐ Simplify auditd checks in Mondoo Linux Security

tas50:
Use the resource to remove the need for properties and complex text processing

### DIFF
--- a/content/mondoo-linux-security.mql.yaml
+++ b/content/mondoo-linux-security.mql.yaml
@@ -4980,7 +4980,7 @@ queries:
         action == "always" &&
         list == "exit" &&
         fields.contains(key == "arch" && op == "=" && value == "b64") &&
-        fields.contains(key == "auid" && op == ">" && value == "=1000") &&
+        fields.contains(key == "auid" && op == ">=" && value == "1000") &&
         fields.contains(key == "auid" && op == "!=" && value == "4294967295") &&
         syscalls.containsAll(["setxattr", "lsetxattr", "fsetxattr", "removexattr", "lremovexattr", "fremovexattr"]) &&
         keyname == "perm_mod"
@@ -4989,7 +4989,7 @@ queries:
         action == "always" &&
         list == "exit" &&
         fields.contains(key == "arch" && op == "=" && value == "b32") &&
-        fields.contains(key == "auid" && op == ">" && value == "=1000") &&
+        fields.contains(key == "auid" && op == ">=" && value == "1000") &&
         fields.contains(key == "auid" && op == "!=" && value == "4294967295") &&
         syscalls.containsAll(["setxattr", "lsetxattr", "fsetxattr", "removexattr", "lremovexattr", "fremovexattr"]) &&
         keyname == "perm_mod"
@@ -4998,7 +4998,7 @@ queries:
         action == "always" &&
         list == "exit" &&
         fields.contains(key == "arch" && op == "=" && value == "b64") &&
-        fields.contains(key == "auid" && op == ">" && value == "=1000") &&
+        fields.contains(key == "auid" && op == ">=" && value == "1000") &&
         fields.contains(key == "auid" && op == "!=" && value == "4294967295") &&
         syscalls.containsAll(["chmod", "fchmod", "fchmodat"]) &&
         keyname == "perm_mod"
@@ -5007,7 +5007,7 @@ queries:
         action == "always" &&
         list == "exit" &&
         fields.contains(key == "arch" && op == "=" && value == "b32") &&
-        fields.contains(key == "auid" && op == ">" && value == "=1000") &&
+        fields.contains(key == "auid" && op == ">=" && value == "1000") &&
         fields.contains(key == "auid" && op == "!=" && value == "4294967295") &&
         syscalls.containsAll(["chmod", "fchmod", "fchmodat"]) &&
         keyname == "perm_mod"
@@ -5016,7 +5016,7 @@ queries:
         action == "always" &&
         list == "exit" &&
         fields.contains(key == "arch" && op == "=" && value == "b64") &&
-        fields.contains(key == "auid" && op == ">" && value == "=1000") &&
+        fields.contains(key == "auid" && op == ">=" && value == "1000") &&
         fields.contains(key == "auid" && op == "!=" && value == "4294967295") &&
         syscalls.containsAll(["chown", "fchown", "fchownat", "lchown"]) &&
         keyname == "perm_mod"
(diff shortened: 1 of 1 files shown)


content/mondoo-linux-security.mql.yaml
```diff
@@ -4129,15 +4129,9 @@ queries:
...
-    props:
-      - uid: mondooLinuxSecurityAuditFiles
-        title: Return the content from all /etc/audit/rules.d and /etc/audit/audit.rules
-        mql: |
-          mondooLinuxSecurityAuditFiles = files.find(from: "/etc/audit/rules.d",regex:'.*\.rules$', type: "file").list.map(path) + ["/etc/audit/audit.rules"]
-          return mondooLinuxSecurityAuditFiles.map(file(_).content.lines.where( _ == /^[^#]/ ))
     mql: |
-      props.mondooLinuxSecurityAuditFiles.any(_.contains(/^(\s+)?\-w\s+\/etc\/sudoers\.d(\/?)\s+\-p\s+wa\s+\-k\s+scope(\s+)?$/))
-      props.mondooLinuxSecurityAuditFiles.any(_.contains(/^(\s+)?\-w\s+\/etc\/sudoers(\/?)\s+\-p\s+wa\s+\-k\s+scope(\s+)?$/))
+      auditd.rules.files.where(path == "/etc/sudoers").all(permissions == "wa" && keyname == "scope")
```
tas50: I'm realizing now this probably isn't what we want since it passes if the files aren't there

content/mondoo-linux-security.mql.yaml
```diff
@@ -4704,18 +4671,15 @@ queries:
   - uid: mondoo-linux-security-events-that-modify-the-systems-mandatory-access-controls-are-collected
     title: Ensure events that modify the system's Mandatory Access Controls are collected
     impact: 50
-    props:
-      - uid: mondooLinuxSecurityAuditFiles
-        title: Return the content from all /etc/audit/rules.d and /etc/audit/audit.rules
-        mql: |
-          mondooLinuxSecurityAuditFiles = files.find(from: "/etc/audit/rules.d",regex:'.*\.rules$', type: "file").list.map(path) + ["/etc/audit/audit.rules"]
-          return mondooLinuxSecurityAuditFiles.map(file(_).content.lines.where( _ == /^[^#]/ ))
+    filters: file("/etc/apparmor").exists || file("/etc/apparmor").exists
```
mm-weber:
```suggestion
    filters: file("/etc/apparmor").exists || file("/etc/selinux/").exists
```

mm-weber (changes requested):
There's a few things that need to be changed or thought about.
`keyname` fields are arbitrary, should we really check for them as fixed values?

content/mondoo-linux-security.mql.yaml
tas50: 🤦 Fixed thanks

content/mondoo-linux-security.mql.yaml
```diff
@@ -5001,78 +4942,93 @@ queries:
...
-            || split("-").containsAll(["F auid!=4294967295 "])
-          && split("-").containsAll(["F arch=b64 "])
-            || split("-").containsAll(["F arch=b32 "])
-          && split("-").containsAll(["F key=perm_mod"])
-            || split("-").containsAll(["k perm_mod"])
+      auditd.rules.syscalls.contains(
+        action == "always" &&
+        list == "exit" &&
+        fields.contains(key == "arch" && op == "=" && value == "b64") &&
+        fields.contains(key == "auid" && op == ">" && value == "=1000") &&
```
tas50: The splitting needs to get fixed in the resource, but I'm leaving it here for now as this is the way the resource returns the data.

tas50: Fixed in https://github.com/mondoohq/cnquery/pull/5858

content/mondoo-linux-security.mql.yaml
```diff
@@ -4560,29 +4527,18 @@ queries:
...
-      props.mondooLinuxSecurityAuditFiles.flat.unique.where(/\/etc\/localtime/).all(
-        split("-").containsAll(["w /etc/localtime ","p wa ",])
-          && split(" ").containsAll(["-F", "key=time-change"])
-            || split(" ").containsAll(["-k", "time-change"])
-      )
+      auditd.rules.syscalls.contains(
+        syscalls.containsAll(["adjtimex", "settimeofday", "clock_settime", "stime"]) &&
+        keyname == "time-change" &&
+        fields.contains(key == "arch" && op == "=" && value == "b32")
+        )
```
mm-weber, line 4535: Unfortunately this pattern only works if we assume that all desired syscalls are put into one rule.

This would my preferred pattern to deal with checks like this, because of the superior output :
```
syscalls_64bit = ["adjtimex", "settimeofday", "clock_settime"]

# This will find all auditd rules that fit our desired pattern and
# create a flat list of the matching syscalls that exist
# Note, we cannot replace .where with .all or .any because this 
# won't allow for more then 1 rule to fulfil the conditions
desiredRules64 = auditd.rules.syscalls.where(
  syscalls.in(syscalls_64bit)
  && fields.contains(key == "arch" && op == "=" && value == "b64")
  && keyname == "time-change"
).map(syscalls).flat

# Here we make sure all syscalls we are looking for actually exist
desiredRules64.containsAll(syscalls_64bit)

# Output Pass
[ok] value: []

# Output Fail missing rules
[failed] [].containsAll()
  expected: == []
  actual:   [
    0: "adjtimex"
    1: "settimeofday"
    2: "clock_settime"
  ]
```

mm-weber (changes requested):
In my opinion we should discuss auditd MQL patterns first, before we do the push here.

content/mondoo-linux-security.mql.yaml
```diff
@@ -4409,48 +4397,27 @@ queries:
...
-      props.mondooLinuxSecurityAuditFiles.flat.unique.any(_ == /\/var\/log\/lastlog/)
-      props.mondooLinuxSecurityAuditFiles.flat.unique.any(_ == /\/var\/log\/tallylog/)
-      props.mondooLinuxSecurityAuditFiles.flat.unique.where(_ == /\/var\/run\/faillock|\/var\/log\/lastlog|\/var\/log\/tallylog/).all(
-        split("-").contains(/p wa/)
-          && split(" ").containsAll(["-k","logins"])
-            || split(" ").containsAll(["-F","key=logins"])
-      )
+      auditd.rules.files.contains(path == "/var/run/faillock" && keyname == "logins" && permissions == "wa")
+      auditd.rules.files.contains(path == "/var/log/lastlog" && keyname == "logins" && permissions == "wa")
+      auditd.rules.files.contains(path == "/var/log/tallylog" && keyname == "logins" && permissions == "wa")
```
mm-weber, line 4408: This check ultimately needs a rework, as do all the auditd checks.
It's a faithful conversion of the original, but the `"/var/log/tallylog"` is only used in Debian 8,9 , SUSE-11/12, Ubuntu 14-18, and should be checked on Amazon or Redhat.

tas50:
Closing this out as it needs rework once the auditd resource is updated

Issue #1798, opened 2025-08-20
Improve the CLI scan experience


tas50:
This issue is a rollup of multiple issues I've encountered while scanning assets from the CLI:

- [ ] Data query output takes up a large amount of space and provides little value in its truncated form. This should be disabled by default in V12
- [ ] Compliance controls passing section lacks framework context and should be disabled by default in V12
- [ ] "Failing:" section should probably be "Failing checks:" since we don't have combined findings at this point

```
Failing:
! Error:          Ensure security auditing retention
✕ MEDIUM (60):    Reduce the sudo timeout period
````

- [ ] Failing section is B&W but Vulnerabilities but other sections are console default

<img width="483" height="244" alt="Image" src="https://github.com/user-attachments/assets/84c14f2c-f165-4dfa-834b-ed6771c3d829" />

- [ ] Failing checks lack a scale. This should be `✕ MEDIUM (60/100):`

<img width="424" height="35" alt="Image" src="https://github.com/user-attachments/assets/3b337122-2875-4089-a7b4-e8599732a832" />

- [ ] Skipped checks shouldn't show up in the Failing section

<img width="452" height="77" alt="Image" src="https://github.com/user-attachments/assets/632dc003-de20-4ce8-aad3-f009efe37173" />

- [ ] Risk not used in the Vulnerabilities section / it's unclear score is CVSS and not risk score.

<img width="160" height="120" alt="Image" src="https://github.com/user-attachments/assets/ee1071d7-fcaf-437c-928a-3597948c3c40" />

- [ ] We expose an overall CVSS score, which is not a thing we do on the console

<img width="203" height="37" alt="Image" src="https://github.com/user-attachments/assets/92254f00-5204-4ee8-a50b-e034cde70cc9" />

- [ ] If only a single asset is scanned the summary format is not a great use of space

<img width="286" height="103" alt="Image" src="https://github.com/user-attachments/assets/4505d912-b140-4390-b1c7-14a0cef91ecc" />

- [ ] Our console link call to action is `See more scan results and asset relationships on the Mondoo Console` which should probably be updated to stress digging in deeper and remediations instead of relationships.

PR #1799, opened 2025-08-20
🎉 v12.0.0-pre1

arlimus:
TODO:
- [x] remove compliance and data queries from the default output
- [ ] flip scoring to risk scoring
- [ ] introduce risk-threshold and deprecate score-threshold

see https://github.com/mondoohq/cnspec/issues/1798


arlimus:
remaining tracking in https://github.com/mondoohq/planning/issues/170#issuecomment-3232385243

PR #1803, opened 2025-08-26
⭐ policy require providers

arlimus:
This introduces a new keyword in policies: `require`. It allows you to specify which providers are required by a policy:

```yaml
policies:
  - uid: example1
    name: Example policy 1
    version: "1.0.0"
    require:
      # Specify which providers this policy requires
      - name: os
    ...
```

This is going to become the default throughout the next major version, i.e. we recommend adding `require` statements with the release of cnspec v12 and will make it mandatory in v13. 

Specifying required providers allows us to significantly reduce the pressure on commands like `lint` and `format`. Both of these currently required **all** providers that exist to be installed. That is annoying, slow, and doesn't scale long-term. Specifying providers allows us to verify requirements once and then work with the bundle.

Upcoming PRs will expand this with version constraints and auto-adding providers during format.


policy/bundle_require.go
```diff
@@ -0,0 +1,75 @@
...
+			if err := ensureRequirement(require, existing); err != nil {
+				return multierr.Wrap(err, "failed to validate policy '"+policy.Name+"'")
+			}
+		}
+	}
+
+	if len(missingRequires) != 0 {
+		log.Info().Strs("policies", missingRequires).Msg("policy doesn't specify required providers, defaulting to installing all default providers")
+		for _, v := range providers.DefaultProviders {
+			if _, err := providers.EnsureProvider(providers.ProviderLookup{ID: v.ID}, true, nil); err != nil {
```
chris-rock: we need to respect the auto-update flag here

chris-rock: added a commit that handles this

chris-rock:
A few things that we still need to check:

- [ ] policy scan does not seem to respect requirements (need to double-check)
- [ ] We need to add a lint warning, when the policy has not defined a requirement
- [ ] When a directory is linted, we see "policy doesn't specify required providers, defaulting to installing all default providers" for each policy file since linting happens for each file separately. We probably want to turn this into a debug message instead.

internal/bundle/bundle_ext.go
```diff
@@ -166,3 +167,26 @@ func (x *RiskMagnitude) UnmarshalYAML(node *yaml.Node) error {
 	}
 	return nil
 }
+
+func (x *HumanTime) UnmarshalYAML(node *yaml.Node) error {
```
chris-rock, line 242: I think we also need to implement MarshalYAML here as we've removed this earlier

chris-rock:
I fixed the cli tests. This issue was caused by the fact that cnquery providers package uses globals and the init function was loaded way before the execution, when we construct the command.

The two things I see left to complete this PR:

- [ ] use require for `cnspec scan`
- [ ] fix the unmarshal for HumanTime

chris-rock:
I update the code to fix the HumanTime handling. The only thing I see left is that require is not used for `cnspec scan`. We may decide that this is a good approach because we can safely test the new feature, enforce it during lint and then switch cnspec scan at a later point.

test/cli/testdata/cnspec_providers_list.ct
```diff
@@ -0,0 +1,6 @@
+$ cnspec providers list
```
chris-rock: I think we should remove this, because it includes specific version numbers

policy/bundle_require.go
```diff
@@ -0,0 +1,82 @@
...
+	for _, policy := range p.Policies {
+
+		// collect policies that do not specify any requirements and install default providers later
+		if len(policy.Require) == 0 && installIfNoRequire {
+			missingRequires = append(missingRequires, policy.Name)
+			continue
+		}
+
+		for _, require := range policy.Require {
+			if err := ensureRequirement(require, existing); err != nil {
```
jaym: I feel like the logic here might not be as expected. If auto-update is true, wouldn't we still expect that we try to call `EnsureProvider` to make sure it updates

policy/bundle_require.go
```diff
@@ -0,0 +1,82 @@
...
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+// ensureRequirement checks if a required provider is installed, and installs it if not.
+func ensureRequirement(require *Requirement, existing providers.Providers) error {
```
jaym: I'm not clear why we need this function. `providers.EnsureProvider` should do all of this, no?

internal/bundle/lint_rules_policy.go
```diff
@@ -58,6 +58,7 @@ const (
 	PolicyMissingVersionRuleID       = "policy-missing-version"
 	PolicyWrongVersionRuleID         = "policy-wrong-version"
 	PolicyRequiredTagsMissingRuleID  = "policy-required-tags-missing"
+	PolicyMissingRequireRuleID       = "policy-missing-require"
```
tas50, line 61: Do we want every user to need to think about this enough to make this a linting rule?

chris-rock: We can extract this to be a separate PR and add this once we've updated all policies. I think we should encourage users to add those as it speeds up the linting as we do not need to download all providers and makes the requirements more obvious.

policy/tooling/tooling.go
```diff
@@ -21,6 +21,12 @@ func Format(data []byte) ([]byte, error) {
 	return fmtData, nil
 }
 
-func Lint(schema resources.ResourcesSchema, filename string, data []byte) []*bundlefmt.Entry {
-	return bundlefmt.LintPolicyBundle(schema, filename, data)
+type LintOptions struct {
+	AutoUpdateProviders bool
```
chris-rock: use new name `SkipProviderDownload`

internal/bundle/lint_rules_policy.go
chris-rock: It is just a warning for now, so it has no impact.

policy/bundle_require.go
chris-rock: Can we just use https://github.com/mondoohq/cnquery/blob/299afc6433f4a0df252a00f72c98b22e911a7564/providers/providers.go#L351

policy/bundle_require.go
```diff
@@ -0,0 +1,82 @@
...
+				return multierr.Wrap(err, "failed to validate policy '"+policy.Name+"'")
+			}
+		}
+	}
+
+	// install default providers for policies that do not specify any requirements
+	if len(missingRequires) != 0 {
+		log.Debug().Strs("policies", missingRequires).Msg("policy doesn't specify required providers, defaulting to installing all default providers")
+		for _, v := range providers.DefaultProviders {
+			if _, err := providers.EnsureProvider(providers.ProviderLookup{ID: v.ID}, autoUpdate, nil); err != nil {
```
chris-rock, line 45: should also pass through the auto-update

internal/bundle/bundle_ext.go
chris-rock: Completed

content/mondoo-linux-workstation-security.mql.yaml
```diff
@@ -8,6 +8,8 @@ policies:
     tags:
       mondoo.com/category: security
       mondoo.com/platform: linux
+    require:
```
tas50, line 11: I have concerns that this name isn't clear enough to a new-ish user without a detailed understanding of how policies and providers work.

apps/cnspec/cmd/bundle.go
```diff
@@ -147,6 +142,11 @@ var policyPublishCmd = &cobra.Command{
 			log.Fatal().Err(err).Msg("could not load policy bundle")
 		}
 
+		// we only need to auto update providers when linting is disabled
+		if err = policyBundle.EnsureRequirements(true, autoUpdate && noLint); err != nil {
+			log.Fatal().Err(err).Msg("could not install requirements")
```
imilchev, line 151: if we introduce a new MQL resource won't linting fail if the provider isn't updated?

content/mondoo-linux-workstation-security.mql.yaml
chris-rock: What we could do: Instead of the following struct

```protobuf
message Requirement {
  string id = 1;
  string name = 2;
  string version = 3;
}
```

we use:

```protobuf
message Requirement {
  string provider =  1; # provider name or provider id
  string version = 3;
}
```

This would change the description to:

```yaml
policies:
  - uid: example1
    name: Example policy 1
    version: "1.0.0"
    require:
      # Specify which providers this policy requires
      - provider: os
```

or 

```yaml
policies:
  - uid: example1
    name: Example policy 1
    version: "1.0.0"
    require:
      - provider: aws
      - provider: terraform
```

apps/cnspec/cmd/bundle.go
chris-rock: Added additional comment to make this behavior more clear.

chris-rock (approved):
Great addition @arlimus

PR #1805, opened 2025-08-26
GCP policy: Swap MySQL for PostgreSQL in text

tas50:
This was copy pasta


arlimus (approved):
Thank you for catching it!

PR #1809, opened 2025-09-01
🌈 streamline CLI output on risk score

arlimus:
After: https://github.com/mondoohq/cnspec/pull/1808

Streamline all output on risk scores. This includes regular reporting formats like compact (default), report, full, and summary.

image

Also includes JSON and YAML outputs. These introduce a new field "riskScore" in each entry, alongside the existing "score". We expect to deprecate and sunset the old score field here as well in the next major release (v13+)


.gitignore
```diff
@@ -1,9 +1,9 @@
-cnquery
-cnspec
+./cnquery
+./cnspec
```
arlimus: fyi since these were not properly namespaced they caused us to ignore the named folders in apps (which e.g. search utilties started to ignore)

PR #1811, opened 2025-09-01
fix(ci): Skip installer dispatch hook on pre-release versions

philipbalinov:
Trigger the final release job only if we are not on a pre-release ref (`-pre`)


czunker (approved):
LGTM

PR #1840, opened 2025-09-23
🧹 update references to golangci.yml -> golangci.yaml

vjeffrey:
same as https://github.com/mondoohq/cnquery/pull/5955


.github/workflows/pr-extended-linting.yml
```diff
@@ -25,5 +25,5 @@ jobs:
         uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 # v8.0.0
         with:
           version: latest
-          args: --config=.github/.golangci.yml --timeout=30m
```
czunker, line 28: This was correct:
https://github.com/mondoohq/cnspec/blob/main/.github/.golangci.yml

To make this work and align it with cnquery, we'd also need to rename the file.

chris-rock: Why do we need two files in the first place?

chris-rock: We use that to enforce more rules for new PRs

chris-rock: I migrated the file to the new format too.

PR #1848, opened 2025-09-24
Removed benchmark workflow as its no longer required

username-is-already-taken2:


chris-rock (approved):
Thank you @username-is-already-taken2

PR #1851, opened 2025-09-26
⭐ Add a new mechanism to send results to the server

jaym:
This PR introduces a new way to send results to the server. Instead of sending data through the API, we instead upload a file to a link provided by the server.

Two new APIs are introduced: GetUploadURLs and ConfirmUploads.
- *GetUploadURLs* requests a place where it can upload the scan results
- *ConfirmUploads* is called once all files have been uploaded

This iteration uses a single file bundle. All results are written to an SQLite database. Tables are created for scores, data, risk factors and resources. SQLite was chosen for two reasons:
- It avoids needing to create a new file format
- SQLite deals with memory management for us. Related to the file format, but we for free get the ability to not have to deserialize all data into memory to process it. This will also be helpful during writes as we will not need to hold all data in memory (the graph executor and resource recorder would still need to be modified to not keep a copy of this and instead write directly to the store)

The implementation is less than ideal in some places. To avoid creating a whole new datalake, we give the datalake a writer for results. This lets us swap out the part that directly writes in memory with one that can write to any store of our choosing. Things like policies, resolved polciy, etc still get cached in the "in-memory" part of the data lake.

Requires https://github.com/mondoohq/cnquery/pull/5967


policy/scandb/scan_db.sql
```diff
@@ -0,0 +1,63 @@
+-- SQLite schema for cnspec policy upload files
+-- Optimized for fast writes on client side
+-- Version: 1.0
+
+-- Metadata table for versioning and asset information
+CREATE TABLE metadata (
```
chris-rock, line 9: Do we include the version of the schema in that table?

jaym: we do. search for the `insertMetadata` function

policy/cnspec_policy.proto
```diff
@@ -879,6 +879,10 @@ service PolicyResolver {
   rpc GetResolvedPolicy(Mrn) returns (ResolvedPolicy) {}
 
   rpc StoreResults(StoreResultsReq) returns (Empty) {}
+  
+  // New signed URL upload flow
+  rpc GetUploadURLs(GetUploadURLsReq) returns (GetUploadURLsResp) {}
+  rpc ConfirmUploads(ConfirmUploadsReq) returns (Empty) {}
```
chris-rock: Lets use `ReportUploadComplete` instead since only the server can confirm results.

jaym: - `GetUploadURLs` was replaced with `GetUploadURL` You can only get 1 file to upload
- `ConfirmUploads` was replaced with `ReportUploadComplete`

policy/scan/local_scanner.go
```diff
@@ -852,11 +853,69 @@ func (s *localAssetScanner) run() (*AssetReport, error) {
 		return nil, err
 	}
 
+	var scanDataStore policy.ScanDataStore
```
chris-rock: Should we not do this before? It would be nice if we could just a different implementation of the datastore. Then the runtime would not need to know if this is a local db or upstream reporting. We could implement this as a wrapper around the inmemory db

policy/scan/local_scanner.go
```diff
@@ -878,6 +937,74 @@ func (s *localAssetScanner) run() (*AssetReport, error) {
 	return ar, nil
 }
 
+func (s *localAssetScanner) uploadScanDataStore(scanDataPath string) error {
```
chris-rock: would be great if we would move this to be a special handling of the database itself.

internal/datalakes/sqlite/sqlite.go
```diff
@@ -0,0 +1,161 @@
...
+func NewServices(runtime llx.Runtime) (*inmemory.Db, *policy.LocalServices, error) {
+	// Reuses the inmemory datalake. All collected data is stored in sqlite via the ScanDataStoreWrapper,
+	// however everything else reuses the inmemory implementation
+	return inmemory.NewServices(runtime)
+}
+
+func WithServices(ctx context.Context, runtime llx.Runtime, assetMrn string, upstreamClient *upstream.UpstreamClient, f func(*policy.LocalServices) error) error {
+	err := withSqliteDataStore(assetMrn, func(scanDataStore scandb.ScanDataStore) error {
+		db, ls, err := NewServices(runtime)
+		db.SetDataWriter(scandb.NewScanDataStoreWrapper(scanDataStore, assetMrn))
```
preslavgerchev: we should move this after the err check, else `db` might be nil

internal/datalakes/inmemory/inmemory.go
```diff
@@ -37,14 +137,28 @@ func NewServices(runtime llx.Runtime) (*Db, *policy.LocalServices, error) {
 	return db, services, nil
 }
 
-// WithDb creates a new set of policy services and closes everything out once the function is done
-func WithDb(runtime llx.Runtime, f func(*Db, *policy.LocalServices) error) error {
-	db, ls, err := NewServices(runtime)
+func (db *Db) SetDataWriter(writer DataStore) {
```
preslavgerchev: nit: can we make this an opt that we pass in when constructing the db? unless this has to be set at a later point in time after the creation

policy/scandb/scan_data_store_wrapper.go
```diff
@@ -0,0 +1,113 @@
...
+	if assetMrn != w.assetMrn {
+		return nil, fmt.Errorf("asset MRN mismatch: expected %s, got %s", w.assetMrn, assetMrn)
+	}
+
+	return w.store.GetScore(ctx, scoreID)
+}
+
+// WriteData writes a single data result, verifying the asset MRN matches
+func (w *ScanDataStoreWrapper) WriteData(ctx context.Context, assetMrn string, data *llx.Result) error {
+	if assetMrn != w.assetMrn {
```
preslavgerchev: can we extract this to a helper fn? it seems equal for all the fns of the wrapper

PR #1860, opened 2025-10-03
🐛 Convert implicit props into explicit props

### LINKED ISSUE #1843: Implicit Properties are not supported
cnquery v12 introduced the ability to have implicit properties.

The following fails to compile:
```
policies:
  - uid: implicit-props
    name: implicit-props policy 1
    version: "1.0.0"
    authors:
      - name: Mondoo
        email: hello@mondoo.com
    groups:
      - filters: return true
        queries:
          - uid: query-1
    props:
      - uid: xtesting
        mql: return true

queries:
  - uid: query-1
    title: testing
    mql: props.xtesting == true
```

```
cnspec policy lint ./policy-new.yaml.mql          x1
→ lint policy bundle file=./policy-new.yaml.mql
→ installing provider 'ai' version=12.0.1
→ successfully installed ai provider path=/Users/atomic111/.config/mondoo/providers/ai version=12.0.1
  RULE ID                       LEVEL    FILE                 LINE  MESSAGE
  bundle-compile-error          error    policy-new.yaml.mql  1     Could not compile policy bundle: failed to validate query '//local.cnspec.io/run/local-execution/queries/query-1': failed to compile query 'props.xtesting == true': failed to compile: cannot find property 'xtesting', please define it first

  policy-required-tags-missing  warning  policy-new.yaml.mql  2     policy 'implicit-props' does not contain the required tag `mondoo.com/category`
  policy-required-tags-missing  warning  policy-new.yaml.mql  2     policy 'implicit-props' does not contain the required tag `mondoo.com/platform`
FTL invalid policy bundle
```

The issue is that when we compile, we try to look up global properties by name. However, they are stored in a map by mrn. I think we would also need to modify the resolver to pass those props. Right now, it assumes that all props are explicit.

I think the way this needs to work is during bundle compilation, we need to resolve implicit props into explicit props. The example above needs to somehow turn into:
```
policies:
  - uid: implicit-props
    name: implicit-props policy 1
    version: "1.0.0"
    authors:
      - name: Mondoo
        email: hello@mondoo.com
    groups:
      - filters: return true
        queries:
          - uid: query-1
    props:
      - mrn: //local.cnspec.io/policies/implicit-props/properties/xtesting
        mql: return true
        for:
        - mrn: //local.cnspec.io/queries/query-1/properties/xtesting

queries:
  - uid: query-1
    title: testing
    mql: props.xtesting == true
    props:
    - mrn: //local.cnspec.io/queries/query-1/properties/xtesting
      type: boolean
```

jaym:
cnquery v12 introduced the ability to have implicit properties.

The following failed to compile:
```
policies:
  - uid: implicit-props
    name: implicit-props policy 1
    version: "1.0.0"
    authors:
      - name: Mondoo
        email: hello@mondoo.com
    groups:
      - filters: return true
        queries:
          - uid: query-1
    props:
      - uid: xtesting
        mql: return true

queries:
  - uid: query-1
    title: testing
    mql: props.xtesting == true
```

This change converts implicit props to explicit props when the bundle is uploaded. A query can lookup from any of its parents. It goes up the chain in order, using the first matching prop. If one is found, we create the prop on the query. The matching prop on the parent references it through `for`.

The bundle above becomes:
```
policies:
  - uid: implicit-props
    name: implicit-props policy 1
    version: "1.0.0"
    authors:
      - name: Mondoo
        email: hello@mondoo.com
    groups:
      - filters: return true
        queries:
          - uid: query-1
    props:
      - mrn: //local.cnspec.io/policies/implicit-props/properties/xtesting
        mql: return true
        for:
        - mrn: //local.cnspec.io/queries/query-1/properties/xtesting

queries:
  - uid: query-1
    title: testing
    mql: props.xtesting == true
    props:
    - mrn: //local.cnspec.io/queries/query-1/properties/xtesting
      type: boolean
```

Fixes #1843

Requires https://github.com/mondoohq/cnquery/pull/6011

### DIFF
--- a/policy/bundle.go
+++ b/policy/bundle.go
@@ -1449,30 +1449,29 @@ func newQueryPropsResolver(query *explorer.Mquery, parents map[string][]parent) 
 
 func (r *QueryPropsResolver) Get(name string) *llx.Primitive {
 	// Check explicitlyl defined properties
-	for propName, queryProp := range r.nameToProp {
-		if propName == name {
-			if queryProp.Type == "" {
-				// Resolve a type if the prop was defined with just a name
-				r.walkParents(func(p hasProps) bool {
-					for _, parentProp := range p.GetProps() {
-						pn, err := explorer.GetPropName(parentProp.Mrn)
-						if err != nil {
-							continue
-						}
-						if pn == name && parentProp.Type != "" {
-							queryProp.Type = parentProp.Type
-							return true
-						}
+	queryProp, ok := r.nameToProp[name]
+	if ok {
+		if queryProp.Type == "" {
+			// Resolve a type if the prop was defined with just a name
+			r.walkParents(func(p hasProps) bool {
+				for _, parentProp := range p.GetProps() {
+					pn, err := explorer.GetPropName(parentProp.Mrn)
+					if err != nil {
+						continue
 					}
-					return false
-				})
-				if queryProp.Type == "" {
-					r.errors = append(r.errors, errors.New("property "+name+" has no type in query "+r.query.Mrn))
-					return nil
+					if pn == name && parentProp.Type != "" {
+						queryProp.Type = parentProp.Type
+						return true
+					}
 				}
+				return false
+			})
+			if queryProp.Type == "" {
+				r.errors = append(r.errors, errors.New("property "+name+" has no type in query "+r.query.Mrn))
+				return nil
 			}
-			return &llx.Primitive{Type: queryProp.Type}
 		}
+		return &llx.Primitive{Type: queryProp.Type}
 	}
 
 	found := struct {


policy/bundle.go
```diff
@@ -1365,19 +1402,19 @@ func (c *bundleCache) precompileQuery(query *explorer.Mquery, policy *Policy) *e
...
-			lookup, ok := c.lookupProps[prop.Mrn]
-			if !ok {
-				c.errors = append(c.errors, fmt.Errorf("property %s not found in bundle for query %s", prop.Mrn, query.Mrn))
-				return nil
-			}
-			query.Props[i] = lookup.Property
-		}
-	}
+	// for i := range query.Props {
+	// 	prop := query.Props[i]
```
imilchev: Can we delete this now?

jaym: yep, fixed in cnquery

policy/bundle.go
```diff
@@ -1399,33 +1422,167 @@ func (c *bundleCache) precompileQuery(query *explorer.Mquery, policy *Policy) *e
...
+		query:      query,
+		parents:    parents,
+		nameToProp: propsMap,
+		errors:     []error{},
+	}, nil
+}
+
+func (r *QueryPropsResolver) Get(name string) *llx.Primitive {
+	// Check explicitlyl defined properties
+	for propName, queryProp := range r.nameToProp {
```
czunker: Why do we have to loop?
`nameToProp` is a map. Couldn't we look up by the name directly?

policy/bundle.go
```diff
@@ -1399,33 +1422,167 @@ func (c *bundleCache) precompileQuery(query *explorer.Mquery, policy *Policy) *e
...
+	r.walkParents(func(p hasProps) bool {
+		for _, prop := range p.GetProps() {
+			propName, err := explorer.GetPropName(prop.Mrn)
+			if err != nil {
+				continue
+			}
+			if propName == name {
+				found.mrn = prop.Mrn
+				found.prop = prop
+				return true
```
czunker, line 1490: What if the prop is also used in another query?

I'm not sure I fully understood your comment and code in this case.
Could the same prop be used in multiple queries?

jaym: We pick the first one, walking up the chain. So if a query is a variant, we check its parent and then the policy. The use case not covered is if 2 policies reference the same query and each define the property.

PR #1866, opened 2025-10-07
🧹 Prevent panic in AdjustRiskScore

czunker:


policy/risk_factor.go
```diff
@@ -318,6 +318,9 @@ func cmpScoredRiskFactors(ri, rj *ScoredRiskInfo) int {
 // It merges the risk factors from the sortedRisks and applies them to the score.
 // Each array of ScoredRiskInfo is expected to be sorted using SortScoredRiskInfo.
 func AdjustRiskScore(score *Score, sortedRisks ...[]*ScoredRiskInfo) {
```
pratik-mondoo, line 320: maybe also check `len(sortedRisks) != 0`.

czunker: Good catch!

pratik-mondoo (approved):

PR #1882, opened 2025-10-21
:broom: Fix SARIF file missing rules error -> Fixes: failing pipeline

mm-weber:
Currently the policy release pipeline (`Lint Policies`) fails because it generates SARIF files with problematic rule indices:
```
"ruleIndex":18446744073709551615,"level":"error","message": ...
```

This:
- adds the missing rules

<s>We might need to temporarily build cnspec in the pipeline to pass this check (instead of using the cnspec/action)</s> -> Turns out we don't. But we'll need to do another release and update the `cnspec/lint` action, again to make subsequent Policy PRs pass.

### DIFF
--- a/test/cli/testdata/cnspec_scan.ct
+++ b/test/cli/testdata/cnspec_scan.ct
@@ -14,54 +14,13 @@ Usage:
   cnspec scan [command]
 
 Available Commands:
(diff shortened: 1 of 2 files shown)


syrull:
LGTM thanks @mm-weber

PR #1898, opened 2025-11-03
✨ Automate Hosted AWS Integration onboarding.

VasilSirakov:
@mati007thm copied over your PR and made some minor adjustments in separate commits. Thank you for this contribution 👍

#### In this PR:
- New `cnspec integrate aws` command which generates a Terraform file and applies it. The Terraform creates an AWS hosted integration in the designated space.
- Documents `cnspec integrate aws` command with a link to the Mondoo docs.
- Extends TF codegen with variable support.
- Adds unit tests.

Generated Terraform:
```tf
terraform {
  required_providers {
    mondoo = {
      source  = "mondoohq/mondoo"
      version = "~> 0.19"
    }
  }
}

provider "mondoo" {
  space = "dazzling-raman-355734"
}

variable "aws_access_key" {
  type        = string
  description = "AWS access key used for authentication"
  sensitive   = true
}

variable "aws_secret_key" {
  type        = string
  description = "AWS secret key used for authentication"
  sensitive   = true
}

resource "mondoo_integration_aws" "this" {
  credentials = {
    key = {
      access_key = var.aws_access_key
      secret_key = var.aws_secret_key
    }
  }
  name = "my-cool-aws-integration"
}

```

image
image

### DIFF
--- a/internal/tfgen/hcl_test.go
+++ b/internal/tfgen/hcl_test.go
@@ -218,9 +218,11 @@ func TestGenericBlockCreation(t *testing.T) {
 			"thing",
 			[]string{"a", "b"},
 			map[string]any{
-				"a": "foo",
-				"b": 1,
-				"c": false,
+				"a":       "foo",
+				"b":       1,
+				"c":       false,
+				"source":  "should be first",
+				"version": "should be second",
 				"d": map[string]any{ // Order of map elements should be sorted when executed
 					"f": 1,
 					"g": "bar",
@@ -251,9 +253,11 @@ func TestGenericBlockCreation(t *testing.T) {
 		assert.Equal(t, "a", data.Labels()[0])
 		assert.Equal(t, "b", data.Labels()[1])
 		expectedOutput := `thing "a" "b" {
-  a = "foo"
-  b = 1
-  c = false
+  source  = "should be first"
+  version = "should be second"
+  a       = "foo"
+  b       = 1
+  c       = false
   d = {
     e = true
     f = 1
--- a/internal/tfgen/hcl.go
+++ b/internal/tfgen/hcl.go
@@ -623,19 +623,22 @@ func HclCreateGenericBlock(hcltype string, labels []string, attr map[string]any,
 	// We need/want to guarantee the ordering of the attributes, do that here
 	keys := sortx.Keys(attr)
 
-	if sourceFound || versionFound {
-		var newKeys []string
-		if sourceFound {
-			newKeys = append(newKeys, "source")
+	var orderedKeys []string
+	if sourceFound {
+		orderedKeys = append(orderedKeys, "source")
+	}
+	if versionFound {
+		orderedKeys = append(orderedKeys, "version")
+	}
+
+	for _, key := range keys {
+		if key != "source" && key != "version" {
+			orderedKeys = append(orderedKeys, key)
 		}
-		if versionFound {
-			newKeys = append(newKeys, "version")
-		}
-		keys = append(newKeys, keys...)
 	}
 
 	// Write block data
-	for _, key := range keys {
+	for _, key := range orderedKeys {
 		val := attr[key]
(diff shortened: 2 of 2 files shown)


internal/tfgen/hcl_test.go
```diff
@@ -270,21 +270,6 @@ func TestGenericBlockCreation(t *testing.T) {
 }
 `
 		assert.Equal(t, expectedOutput, tfgen.CreateHclStringOutput(data))
-
-	})
-	t.Run("should fail to construct generic block with mismatched list element types", func(t *testing.T) {
```
VasilSirakov, line 275: Spoke with @jaym about this type mismatch check. I'm removing this for the time being.

Right now it's confusing because this check was only applied once in a very particular scenario and could be avoided depending on which Go type was used in code. Secondly, we don't really have a use case that would warrant such a check yet, and lastly even if we allow the generation of invalid TF it's something we can easily catch in the development phase.

If we identify in the future such a scenario, we can always add it but would rather have a use-case first.

apps/cnspec/cmd/integrate.go
```diff
@@ -55,6 +58,109 @@ var (
...
+		PreRunE: func(cmd *cobra.Command, _ []string) error {
+			errs := []error{
+				viper.BindPFlag("space", cmd.Flags().Lookup("space")),
+				viper.BindPFlag("output", cmd.Flags().Lookup("output")),
+				viper.BindPFlag("integration-name", cmd.Flags().Lookup("integration-name")),
+			}
+			return errors.Join(errs...)
+		},
+		RunE: func(_ *cobra.Command, _ []string) error {
+			var (
```
preslavgerchev: nit: why have those in a var block?

internal/onboarding/aws.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/cockroachdb/errors"
+	"github.com/google/uuid"
+	"github.com/rs/zerolog/log"
+
+	"go.mondoo.com/cnquery/v12/cli/theme"
+	"go.mondoo.com/cnspec/v12/internal/tfgen"
+)
+
+// AzureIntegration represents the configuration of an AWS integration to be created.
```
preslavgerchev:
```suggestion
// AwsIntegration represents the configuration of an AWS integration to be created.
```

internal/onboarding/aws.go
```diff
@@ -0,0 +1,110 @@
...
+type AwsIntegration struct {
```
preslavgerchev: nit: can we make this private and only expose via `NewAwsIntegration`?

internal/onboarding/aws.go
```diff
@@ -0,0 +1,110 @@
...
+	Name      string
+	Space     string
+	AccessKey string
+	SecretKey string
+}
+
+func NewAwsIntegration(name, space, accessKey, secretKey string) AwsIntegration {
+	if name == "" {
+		name = fmt.Sprintf("AWS Integration (%s)", uuid.New().String()[:7])
```
preslavgerchev: this looks a bit strange, can we make a hlper to givce us a random name?

internal/onboarding/aws.go
```diff
@@ -0,0 +1,110 @@
...
+	if a.AccessKey == "" {
+		errs = append(errs, errors.New("missing AWS access key"))
+	}
+	if a.SecretKey == "" {
+		errs = append(errs, errors.New("missing AWS secret key"))
+	}
+	return errs
+}
+
+// GenerateAzureHCL generates automation code to create an AWS integration.
```
preslavgerchev:
```suggestion
// GenerateAwsHCL generates automation code to create an AWS integration.
```

apps/cnspec/cmd/integrate.go
VasilSirakov: I think it's because the `var` block gives you nice horizontal alignment. Don't think it's something we use internally, so I'll change it back to individual assignments.

PR #1903, opened 2025-11-05
🧹 Stop exposing resolved policy cache

jaym:
- It looks unused. Each scan will get a single resolved policy. Caching doesn't make sense in this context


chris-rock (approved):
Thank you @jaym

PR #1934, opened 2025-11-28
update `mondoo-linux-security-window-system-is-not-installed`

syrull:


content/mondoo-linux-security.mql.yaml
```diff
@@ -812,7 +812,7 @@ queries:
     title: Ensure X Window System is not installed
     impact: 100
     mql: |
-      packages.none(name == /^xserver-xorg.*/ || name == /^xorg-x11/ || name == /^xserver/)
+      packages.none(name == /^xserver-xorg.*/ || name == /^xorg-x11/ && name != /^xorg-x11-fonts/ || name == /^xserver/)
```
chris-rock: Since we do not have precedence with brackets, this logic is difficult to understand. I recommend we split this into:

```
packages.none(name == /^xserver-xorg.*/)
packages.none(name == /^xorg-x11/ && name != /^xorg-x11-fonts/)
packages.none(name == /^xserver/)
```

chris-rock: I updated the policy to fix the MQL

PR #1945, opened 2025-12-05
🧹 Add input validation to GH workflows

czunker:
https://blog.gitguardian.com/github-actions-security-cheat-sheet/

https://docs.github.com/en/actions/reference/security/secure-use\#use-an-intermediate-environment-variable


czunker:
What we did for packer, also applied here: https://github.com/mondoohq/packer-plugin-cnspec/pull/468

PR #1962, opened 2025-12-12
⭐ Add a new check for kernel.kptr_restrict to Linux policy

tas50:
This acts as another layer of security over ASLR

### DIFF
--- a/content/mondoo-linux-security.mql.yaml
+++ b/content/mondoo-linux-security.mql.yaml
@@ -67,12 +67,17 @@ policies:
           asset.family.contains('linux')
         checks:
           - uid: mondoo-linux-security-window-system-is-not-installed
-          - uid: mondoo-linux-security-address-space-layout-randomization-aslr-is-enabled
           - uid: mondoo-linux-security-aide-is-installed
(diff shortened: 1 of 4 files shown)


content/mondoo-linux-security.mql.yaml
```diff
@@ -723,6 +728,79 @@ queries:
...
             sysctl -w kernel.randomize_va_space=2
             ```
+  - uid: mondoo-linux-security-restrict-kernel-address-access
+    title: Restrict access to kernel addresses
+    impact: 70
+    mql : |
+      kernel.parameters["kernel.kptr_restrict"] == 2
+    docs:
+      desc: |
+        This check verifies that access to kernel addresses is restricted by confirming that the kernel parameter kernel.kptr_restrict is set to 2, which limits the exposure of kernel pointers in user-space interfaces.
```
chris-rock, line 738: I think it is a great addition. We should also add the reference to https://access.redhat.com/articles/7133962 and also highlight, that this may affect existing tools.

tas50: Let me add that to the followup I have going

PR #1966, opened 2025-12-14
⭐ Add additional Linux Security policy kernel checks

tas50:
- General cleanup of different checks in the Linux policy including wording and spacing
- New Linux Security checks for information disclosure attacks as well as attacks against modern subsystems like ebpf and ptrace
- Fix the incorrect name of the nfs server service
- Make the /etc/CONFIG check bash remediation idempotent

### DIFF
--- a/content/mondoo-linux-security.mql.yaml
+++ b/content/mondoo-linux-security.mql.yaml
@@ -24,6 +24,7 @@ policies:
           - Hardening SSH configurations
           - Ensure users and groups are securely configured
           - Identifying misconfigured Kernel networking configurations
+          - Harden various Kernel parameters
 
         This policy has been developed for Red Hat (RHEL), Debian, Ubuntu, and SUSE (SLES) derivative distributions running on x64 architectures.
         Some queries may be skipped depending on your particular distribution, installation type, or underlying infrastructure.
@@ -78,6 +79,10 @@ policies:
         checks:
           - uid: mondoo-linux-security-address-space-layout-randomization-aslr-is-enabled
           - uid: mondoo-linux-security-restrict-kernel-address-access
+          - uid: mondoo-linux-security-prevent-mapping-low-memory-addresses
+          - uid: mondoo-linux-security-restrict-dmesg-access
+          - uid: mondoo-linux-security-prevent-nonprivileged-ebpf-access
+          - uid: mondoo-linux-security-prevent-nonprivileged-ptrace-access
       - title: Network Stack
         filters: |
           asset.family.contains('linux')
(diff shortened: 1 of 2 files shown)


content/mondoo-linux-security.mql.yaml
```diff
@@ -790,17 +804,365 @@ queries:
...
+        **Why this matters**
+
+        Low memory addresses (typically below 64KB) are often targeted by attackers to exploit vulnerabilities such as null pointer dereferences or buffer overflows. By mapping memory at these low addresses, malicious actors can manipulate program behavior, potentially leading to privilege escalation or arbitrary code execution.
+
+        By setting `vm.mmap_min_addr` to `65536` or higher:
+          - The system reduces the risk of exploitation through low memory address mappings.
+          - It enhances overall security by limiting the ability of unprivileged users to map memory in sensitive regions.
+          - It helps prevent attacks that rely on predictable memory layouts, thereby strengthening the system's defenses against various exploitation techniques.
+
+        Configuring `vm.mmap_min_addr` to at least `65536` is a proactive measure to safeguard against memory-based attacks and improve the security posture of Linux systems.
```
chris-rock, line 844: I think we need to mention what Debian mentions: 

The downside to preventing applications from mmap'ing low virtual memory addresses is that certain applications depend on this functionality. dosemu, qemu and wine are three such applications that exist in Debian. See the [application specific information](https://wiki.debian.org/mmap_min_addr#apps) below.

And we should probably ref https://wiki.debian.org/mmap_min_addr

content/mondoo-linux-security.mql.yaml
```diff
@@ -790,17 +804,365 @@ queries:
...
+              sed -i 's/^vm.mmap_min_addr.*/vm.mmap_min_addr = 65536/' /etc/sysctl.conf
+            else
+              echo "Setting vm.mmap_min_addr to 65536 in /etc/sysctl.d/99-mmap-min-addr.conf"
+              echo 'vm.mmap_min_addr = 65536' | tee /etc/sysctl.d/99-mmap-min-addr.conf
+            fi
+
+            echo "Applying vm.mmap_min_addr=65536"
+            sysctl -w vm.mmap_min_addr=65536
+            ```
+  - uid: mondoo-linux-security-restrict-dmesg-access
```
chris-rock, line 906: Maybe we add the ref to https://documentation.ubuntu.com/security/security-features/kernel-protections/

tas50: Let me see if I can find one just for dmesg until we get some of these others added. I have a big list that includes most of these

content/mondoo-linux-security.mql.yaml
tas50: Added a ref and a generic warning. Most of those apps are only impacted on Debian 5 though so folks should be good to set this

content/mondoo-linux-security.mql.yaml
tas50: I added the kernel docs

PR #1976, opened 2025-12-16
✨ Policy Migrations implementations

slntopp:


internal/bundle/testdata/fail-migration-missing-conditions.yaml
```diff
@@ -0,0 +1,226 @@
+policies:
```
chris-rock, line 1: Is it possible to reduce the samples to a more minimal set to meet the tests?

policy/cnspec_policy.proto
```diff
@@ -175,13 +175,21 @@ message Bundle {
...
+}
+
 message MigrationGroup {
-  // Information about the policy that is being migrated
-  Policy policy = 1;
+  // Human-readable group title
+  string title = 1;
   // Deprecated: single stage migration
+  // used in policymonkey at the moment
   repeated Migration migrations = 2;
```
chris-rock, line 188: I think we can remove it and just reserve it.

```
reserved 2;
```

policy/cnspec_policy.proto
```diff
@@ -206,6 +214,9 @@ message MigrationStage {
   string title = 1;
   // Migrations to execute, they are not guaranteed to be executed in order.
   repeated Migration migrations = 2;
```
chris-rock, line 219: should we call this `check_migrations`

policy/cnspec_policy.proto
```diff
@@ -175,13 +175,21 @@ message Bundle {
   repeated MigrationGroup migration_groups = 11;
 }
 
+message MigrationPolicyRef {
```
chris-rock, line 178: Should we introduce a simple `PolicyRef`: 

```proto
message PolicyRef {
  string uid = 1;
  string version = 2;
}

message MigrationConditions {
  PolicyRef sourcePolicy
  PolicyRef targetPolicy
}
```

policy/cnspec_policy.proto
slntopp: Or query? We could

chris-rock: lets use `query_migrations`

policy/cnspec_policy.proto
slntopp: still used in some policymonkey scripts, we can do a follow-up once those are cleaned up

chris-rock (approved):
Thank you @slntopp

PR #2013, opened 2026-01-07
Don't suggest disabling kernel modules you can't disable

tas50:
Only check to see if modules are disabled when those modules exist on the system. Some kernels compile these in: can't disable those. Some kernels don't ship them at all: nothing to fix.


mm-weber (approved):
LGTM, thanks @tas50 !

PR #2022, opened 2026-01-14
🧹 Improve cli printing when no checks are present.

### LINKED ISSUE #6390: Too much space in CLI output left if no checks are run
**Describe the bug**

Image

There should only be a single empty line when we scan an asset that has no checks.

preslavgerchev:
Fixes https://github.com/mondoohq/cnquery/issues/6390


tas50:
@preslavgerchev ping

tas50:
Fixed in https://github.com/mondoohq/cnspec/pull/2845

PR #2026, opened 2026-01-15
⭐ Improvements to the variants in the AWS policy

tas50:
- Remove variants from checks when there was only one
- Fix some incorrectly named "-single" variants that were actually account checks
- Add tags for the AWS resources to improve UI display
- Fix an incorrect filter for the RDS DB Cluster query that was checking for DB instances instead


arlimus (approved):
Awesome cleanup!

PR #2028, opened 2026-01-15
🧹 Remove all the project level GCP variants

tas50:
Resources + TF only. Avoid duplicate results on the project and the resource. This is the final component of our migration to fine grained scanning everywhere


arlimus (approved):
this will remove dupes - kudos!

PR #2051, opened 2026-01-24
🐛 Fix more policy bugs

tas50:
More typos in these checks that are causing them to fail to run as expected

### DIFF
--- a/content/mondoo-dns-security.mql.yaml
+++ b/content/mondoo-dns-security.mql.yaml
@@ -73,7 +73,7 @@ queries:
     mql: |
       dns.mx != empty
       dns.mx.all(domainName != regex.ipv4 && domainName != regex.ipv6)
-      dns.records.where(type == "NS") !=empty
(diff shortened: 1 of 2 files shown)


content/mondoo-tls-security.mql.yaml
```diff
@@ -67,7 +67,7 @@ queries:
       checkA1 = tls.certificates.first.subject.commonName == asset.fqdn
 
       if(tls.certificates.first.subject.commonName.contains(/^\*/)) {
-        checkA1 == asset.fqdn.contains(tls.certificates.first.subject.commonName.split("*.")[1])
+        checkA1 = asset.fqdn.contains(tls.certificates.first.subject.commonName.split("*.")[1])
```
chris-rock: that change is not correct, we want to do this assertions. Instead we should switch the `{}` syntax to `.all()`

tas50: reverted

PR #2064, opened 2026-02-01
🧹 Remove the check for the legacy google workspace asset platform

tas50:
This has been google-workspace for a while


chris-rock (approved):
Thank you @tas50

PR #2073, opened 2026-02-03
Update ansible remediation to use blockinfile for auditd rules

M-gre:
Customer noted that some of the ansible remediation snippets for audtid rules are not idempotent

### DIFF
--- a/content/mondoo-linux-security.mql.yaml
+++ b/content/mondoo-linux-security.mql.yaml
@@ -5292,18 +5292,18 @@ queries:
                   notify: Load audit rules
 
                 - name: Add Debian/Ubuntu-specific faillog audit rule
-                  ansible.builtin.blockinfile:
+                  ansible.builtin.lineinfile:
                     path: /etc/audit/rules.d/50-logins.rules
-                    block: "-w /var/log/faillog -p wa -k logins"
+                    line: "-w /var/log/faillog -p wa -k logins"
                     create: yes
                     insertafter: EOF
                   when: ansible_os_family == "Debian"
                   notify: Load audit rules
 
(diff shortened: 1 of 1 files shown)


content/mondoo-linux-security.mql.yaml
```diff
@@ -5292,18 +5292,18 @@ queries:
                   notify: Load audit rules
 
                 - name: Add Debian/Ubuntu-specific faillog audit rule
-                  ansible.builtin.lineinfile:
+                  ansible.builtin.blockinfile:
```
czunker: I think this was fine.

It was only a line, and according to the docs, this should be idempotent:
https://docs.ansible.com/projects/ansible/latest/collections/ansible/builtin/lineinfile_module.html

content/mondoo-linux-security.mql.yaml
```diff
@@ -5292,18 +5292,18 @@ queries:
...
-                    line: "-w /var/log/faillog -p wa -k logins"
+                    block: "-w /var/log/faillog -p wa -k logins"
                     create: yes
                     insertafter: EOF
                   when: ansible_os_family == "Debian"
                   notify: Load audit rules
 
                 - name: Add RedHat/Fedora/Amazon-specific faillock audit rule
-                  ansible.builtin.lineinfile:
+                  ansible.builtin.blockinfile:
```
czunker: Same here

czunker (approved):

content/mondoo-linux-security.mql.yaml
M-gre: I reverted the change

content/mondoo-linux-security.mql.yaml
M-gre: reverted

PR #2081, opened 2026-02-05
🧹 report error when we cannot store results

vjeffrey:


.github/workflows/spell-check.yaml
```diff
@@ -39,6 +39,11 @@ jobs:
             cspell:software-terms/src/software-tools.txt
             cspell:k8s/dict/k8s.txt
             mondoo:mondoo_dictionary.txt
+            cspell:software-terms/dict/softwareTerms.txt
```
tas50, line 42: We probably want to keep these as-is

vjeffrey: without that it told me there were 99 words it didnt know about

vjeffrey: Screenshot 2026-02-05 at 10 29 09

tas50:
Is this up to date with main. I'm not seeing these same issues with other PRs

PR #2083, opened 2026-02-05
only check CMK´s not aws managed keys

LittleSalkin1806:
acme is one of the view aws managed keys where we dont have 'kms:GetKeyRotationStatus' per policy.
As everything is AWS managed nevertheless we can do two things.
1. open support request towards AWS in order to change acme key policy (key was created august 2024)
2. filter every AWS managed key and trust aws that they will handle there keys properly.


tas50 (approved):

tas50:
Thanks @LittleSalkin1806

PR #2090, opened 2026-02-09
⭐ Expand our DNS, Email, HTTP, and TLS policies

tas50:
Fill in the gap with additional checks

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -90,6 +90,7 @@ fromjson
 frontbridge
 frontmatter
 fsetxattr
+fullchain
 fumadocs
 getstealthmode
 globalstate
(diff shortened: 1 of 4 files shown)


content/mondoo-dns-security.mql.yaml
```diff
@@ -229,3 +231,62 @@ queries:
...
+          2. If only one NS record exists, add a secondary nameserver through your DNS provider or a third-party DNS service.
+          3. Ensure the additional nameserver is on a different network or subnet to avoid correlated failures.
+          4. Verify the configuration using DNS validation tools to confirm all nameservers respond correctly.
+    refs:
+      - url: https://datatracker.ietf.org/doc/html/rfc1034#section-4.1
+        title: "RFC 1034: Domain Names - Concepts and Facilities"
+  - uid: mondoo-dns-security-reasonable-ttls
+    title: Ensure DNS record TTLs are not suspiciously low
+    impact: 50
+    mql: dns.records.where(type == "A" || type == "AAAA").all(ttl == null || ttl >= 60)
```
arlimus, line 267: I wonder what would cause TTL to be null....

tas50: The assumption here is it could be undefined but let me check if that's ever the case

PR #2091, opened 2026-02-09
⭐ Add additional checks to GitHub / GitLab policies

tas50:
Use our new resources and fields

### DIFF
--- a/content/mondoo-edr-policy.mql.yaml
+++ b/content/mondoo-edr-policy.mql.yaml
@@ -171,15 +171,15 @@ queries:
         **Why this matters**
 
         Endpoint Detection and Response (EDR) agents play a vital role in maintaining a robust security posture by providing continuous monitoring and protection against malicious activities. Ensuring that these agents are running is essential for:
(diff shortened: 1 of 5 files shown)


content/mondoo-github-security.mql.yaml
```diff
@@ -904,3 +913,363 @@ queries:
...
+    impact: 70
+    mql: github.organization.dependabotSecurityUpdatesEnabledForNewRepos == true
+    docs:
+      desc: |
+        This check ensures that Dependabot security updates are automatically enabled for all new repositories created in the organization.
+
+        **Why this matters**
+
+        Dependabot security updates go beyond alerting by automatically creating pull requests to update vulnerable dependencies to the minimum patched version. Enabling this by default provides proactive remediation:
+          •  It reduces the mean time to remediate (MTTR) for known vulnerabilities by automating the fix process.
```
arlimus: since we are MD-formatted, I think we need to use `-` as lists (some renderers may support '•', but it's non-standard)

arlimus (approved):
kudos! moar content yay

PR #2097, opened 2026-02-10
⭐ Add GKE / Redis / Firewall checks to the GCP policy

tas50:
Round out the capabilities of this policy


tas50:
Added to the larger PR now

PR #2098, opened 2026-02-10
:star: Add new checks to the Dockerfile policy

tas50:
- Don't run as root
- Don't build from latest tags


tas50:
The git ref for this branch name is causing copywrite to fail. Closing in favor of https://github.com/mondoohq/cnspec/pull/2107

PR #2113, opened 2026-02-13
⭐ Update GCP checks to use the new Redis platform

tas50:
Move these checks off the project and onto the assets

Depends on https://github.com/mondoohq/cnquery/pull/6612


tas50:
Moved this into https://github.com/mondoohq/cnspec/pull/2115. Closing this now

PR #2114, opened 2026-02-13
:star: Expand Azure security policy to 51 checks across 10 groups

tas50:
Add 25 new security checks and reorganize from 6 groups to 10, bringing Azure policy coverage in line with GCP/AWS. New groups cover Microsoft Defender for Cloud, AKS, App Service, Azure Cache for Redis, and Azure Network. Existing groups for Storage, SQL, Key Vault, and Compute gain additional checks for soft delete, TLS, threat detection, AD admin, RBAC, key rotation, and disk encryption.

- Add 8 Defender for Cloud checks (servers, app service, SQL, storage, key vault, containers, resource manager, auto-provisioning)
- Add 3 AKS checks (RBAC, authorized IP ranges, network policy)
- Add 5 App Service checks (HTTPS, FTP/SCM auth, managed identity, TLS)
- Add 2 Redis checks (SSL-only, public access)
- Add 3 Storage checks (blob/container soft delete, min TLS 1.2)
- Add 3 SQL checks (threat detection, AD admin, audit retention moved)
- Add 2 Key Vault checks (RBAC authorization, key auto-rotation)
- Add 1 Compute check (data disk encryption)
- Add 1 Network check (watcher flow logs)
- Move NSG checks from Compute to new Network group
- Rename "Azure Core" to "Azure Subscription"
- Update policy description with GCP-style service listing

### DIFF
--- a/content/mondoo-azure-security.mql.yaml
+++ b/content/mondoo-azure-security.mql.yaml
@@ -3273,7 +3273,7 @@ queries:
     filters: |
       asset.platform == "azure-sql-server"
     mql: |
-      azure.subscription.sql.server.azureAdAdministrators.length > 0
+      azure.subscription.sql.server.azureAdAdministrators != empty
     docs:
       desc: |
         This check ensures that an Azure Active Directory (Azure AD) administrator is configured for Azure SQL servers, enabling centralized identity management and stronger authentication.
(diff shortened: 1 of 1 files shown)


content/mondoo-azure-security.mql.yaml
```diff
@@ -2943,3 +2989,1604 @@ queries:
...
+              storage_account_type = "Standard_LRS"
+              create_option        = "Empty"
+              disk_size_gb         = 128
+              disk_encryption_set_id = azurerm_disk_encryption_set.example.id
+            }
+            ```
+    refs:
+      - url: https://learn.microsoft.com/en-us/azure/virtual-machines/disk-encryption-overview
+        title: Overview of managed disk encryption options
+  - uid: mondoo-azure-security-network-watcher-flow-logs-enabled
```
arlimus, line 3530: @tas50 flagging this one as a follow-up: If there are no network watchers, should this pass? It reads like we need flow logs and thus would also require the network watchers to exist, right? (I may be wrong, would love your insights)

PR #2123, opened 2026-02-16
✨ Add Exceptions segment to the scan output

### LINKED ISSUE #1973: cnspec scans are showing checks with exceptions as if they are active
**Describe the bug**
When a user sets an exception on Mondoo Platform and then runs a scan, those checks are still showing up as failing checks. We need to expose these as exceptions so the accurate state of the system is shown on the CLI:

Scan of a host showing critical findings:

Image

No critical findings shown on the asset:

Image

Approved exception set on this asset:

Image

slntopp:
> [!NOTE]
>Exceptions list is inferred from the checks impact action to avoid making extra requests to the server.
> Due to this limitation, exceptions of type `Disabled` are completely omitted in the scan output and we can't
> infer actual exception type nor it's TTL

Example scan output:
example scan output

Resolves #1973

### DIFF
--- a/cli/reporter/print_compact.go
+++ b/cli/reporter/print_compact.go
@@ -38,9 +38,6 @@ type defaultReporter struct {
 	// indicates if the StoreResourcesData MQL feature is enabled
 	isStoreResourcesEnabled bool
 
-	// faint dims the output for exceptions
-	faint bool
-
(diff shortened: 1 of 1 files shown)


tas50:
@slntopp Can we put failures below exceptions in the output please?

PR #2138, opened 2026-02-23
⭐ Add 68 new security checks for AWS and Azure and use resource-specific Azure platforms

tas50:
- Add 44 new AWS security checks across EC2, RDS, S3, EKS, Lambda, and more
- Add 24 new Azure security checks including VNet VM protection and AKS
  public network access
- Use resource-specific Azure platforms (azure-aks-cluster,
  azure-keyvault-vault, etc.) for targeted check filtering
- Modernize 3 Azure checks to use native typed fields instead of properties
  dict access (Key Vault purge protection, Key Vault public network access,
  subnet default outbound access)

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -26,6 +26,7 @@ ashburn
 asl
 aslr
 atlassian
+authnotrequired
 autoconfiguration
 autoremove
 awslogs
@@ -79,6 +80,7 @@ ctl
 CUSTOMERID
 CYAAAAAAAKEY
 cyserver
+databricks
 datalakes
 dccp
 ddos
@@ -116,6 +118,7 @@ examplestorageaccount
 exim
 faillock
 faillog
+fargate
 FCr
 FEMI
 fileless
@@ -130,6 +133,7 @@ ftruncate
 fullchain
 fumadocs
 functionapp
+Gbps
 getenv
 getstealthmode
 globalstate
@@ -143,10 +147,12 @@ Hostbased
 hostpath
 hotlink
 hvm
+Iaa
 iap
 idps
 idpuser
 iis
+IKEv
 inplace
 insertafter
 ipfw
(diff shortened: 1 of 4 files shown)


tas50:
## Re-Review: All FAIL Items Resolved ✅

All 5 issues from the initial review have been addressed:

| # | Check | Resolution |
|---|-------|-----------|
| 1 | `elasticache-serverless-cmk-encryption` | All 4 variants now use `!= empty` instead of `!= ""` |
| 2 | `ssm-parameter-encryption` | Title updated to "use KMS encryption" — no longer claims CMK enforcement |
| 3 | `timestream-database-encrypted` | Now uses `kmsKeyId != empty` |
| 4 | `aks-disk-encryption-set` | Duplicate `enableEncryptionAtHost` assertion removed, title updated |
| 5 | CosmosDB field access | All CosmosDB checks now use direct field access consistently |

### Additional Fixes (previously WARN, now resolved)
- `appgw-ssl-policy-tls12` — Null guard added via `.where()` + `.none()` two-assertion pattern
- `redis-auth-required` — Redundant `null ||` removed, simplified to `!= "yes"`
- `batch-pool-disk-encryption` — Added `&& targets != empty` to prevent false pass on empty config
- `guardduty-all-features-enabled` — Added `.length == 6` check to enforce all features present

### Remaining non-blocking observations
- Bracket notation on DMS (`_["PubliclyAccessible"]`), CodeBuild (`source["insecureSsl"]`), ECS (`configuration["containerInsights"]`) — may reflect actual cnquery provider schema for these resources
- MySQL flexible server uses `properties[]` access while PostgreSQL uses direct fields — likely a schema difference
- `rds-instance-iam-auth` changed to per-instance `aws-rds-dbinstance` scope — confirm this is intentional vs account-level `aws` scope used by other new checks
- Inspector checks use `length > 0` (passes with partial coverage) — design-level concern
- `ecs-cluster-fargate-encryption` will fail for EC2-only clusters — consider adding Fargate capability filter

LGTM! Thanks @tas50 🎉

---
🤖 Generated with [Claude Code](https://claude.ai/code)

PR #2161, opened 2026-03-03
Add platform ID to upstream headers for scan request

slntopp:


go.mod
```diff
@@ -49,7 +49,7 @@ require (
 	github.com/tliron/glsp v0.2.2
 	github.com/zclconf/go-cty v1.17.0
 	go.mondoo.com/mondoo-go v0.0.0-20260211131346-5a2e107d8ab4
-	go.mondoo.com/mql/v13 v13.0.0-rc2
+	go.mondoo.com/mql/v13 v13.0.0-rc2.0.20260303140528-8b2327051075
```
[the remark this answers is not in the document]
slntopp: this is to make the tests pass, until `mql` PR is merged

apps/cnspec/cmd/root.go
```diff
@@ -143,6 +147,35 @@ func init() {
 	_ = viper.BindEnv("providers_url")
 
 	config.Init(rootCmd)
+
+	// Register system info provider so that upstream HTTP requests include the
+	// Mondoo-PlatformID header. Without this, the server cannot correlate scan
+	// requests with managed agents, causing stale check-in times.
+	var (
+		sysInfoOnce   sync.Once
+		cachedSysInfo *rangerUtils.ClientSysInfo
```
VasilSirakov, line 156: do you need to check `if cachedSysInfo != nil` and return if present before `sysinfo.Get()`? At the same time you have `sync.Once`, so probably not but still wondering where `cachedSysInfo` is used?

VasilSirakov: nvm, `sync.Once` supports a void method, got confused. You can also cache this within the sys info provider function and not as a `var` up top.

PR #2192, opened 2026-03-10
Use go.mod as single source of truth for Go version in CI

czunker:
Read Go version from go.mod via go-version-file instead of .github/env, and bump Go from 1.25.1 to 1.25.8.

### DIFF
--- a/.github/workflows/pr-test-lint.yml
+++ b/.github/workflows/pr-test-lint.yml
@@ -21,12 +21,10 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-      - name: Import environment variables from file
-        run: cat ".github/env" >> $GITHUB_ENV
       - name: Install Go
         uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
         with:
-          go-version: ">=${{ env.golang-version }}"
+          go-version-file: go.mod
           cache: false
       - name: Check go mod
         run: |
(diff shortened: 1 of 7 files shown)


slntopp (approved):

.github/workflows/pr-test-lint.yml
```diff
@@ -168,12 +161,10 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-      - name: Import environment variables from file
-        run: cat ".github/env" >> $GITHUB_ENV
       - name: Install Go
         uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
         with:
-          go-version: ">=${{ env.golang-version }}"
+          go-version-file: go.mod
```
imilchev, line 170: i think this will not do what we want it to do. `go.mod` is specifying the minimum go version required to build our binary. However, that doesn't mean we cannot use a newer version to build it.

If we pair these 2 it means we always need to update the `go.mod` version to latest possible Go complier, to use the latest compiler improvements in the toolchain. When we have 2 separate number, we can still make sure the server builds with older versions but use newer Go versions just for building in CI

czunker: I don't understand this part:
> When we have 2 separate number, we can still make sure the server builds with older versions ...

What do we need the old versions for?

When we want the latest patch version, we could use: https://github.com/actions/setup-go/blob/main/docs/advanced-usage.md#check-latest-version

imilchev: we are importing this in other projects as well. Forcing the latest go version in the go.mod would mean anywhere where we import cnspec we will need to be updating the go version in the `go.mod`. As we have seen in the past that may not always be possible if other packages don't build with latest golang version

czunker: @imilchev Please have another look.
I changed how we use the version.

PR #2209, opened 2026-03-13
fix: allow non-nil datapoint results to override nil from short-circuit evaluation

jaym:
## Summary
- When `&&` short-circuits in MQL, it caches nil for unevaluated branches. If two queries share a datapoint (same checksum) and the short-circuiting query reports first, its nil poisons the shared `DatapointNode` (first-write-wins), causing subsequent queries to get incorrect nil results.
- Fix by allowing non-nil results to override previously-reported nil results in `DatapointNodeData`, `ReportingQueryNodeData`, and `ReportingJobNodeData`.
- Adds 6 new test cases covering the nil override behavior and ensuring real results are never overridden.

## Test plan
- [x] `go test ./policy/executor/internal/... -count=1` — all pass
- [x] `go test ./policy/executor/... -count=1` — all pass
- [x] `go test ./policy/... -count=1` — all pass (one pre-existing flaky test in `pdque` unrelated)
- [ ] Manual test: `cnspec scan local` with policies containing `&&` short-circuit queries sharing datapoints with other queries

### DIFF
--- a/policy/executor/internal/nodes.go
+++ b/policy/executor/internal/nodes.go
@@ -14,6 +14,19 @@ import (
 	"go.mondoo.com/mql/v13/utils/multierr"
 )
 
+// isNilResult returns true if the result has nil value and no error,
+// which indicates it came from short-circuit evaluation.
+func isNilResult(res *llx.RawResult) bool {
+	return res != nil && res.Data != nil &&
+		res.Data.Value == nil && res.Data.Error == nil
+}
+
+// hasRealData returns true if the result contains a non-nil value or an error.
+func hasRealData(res *llx.RawResult) bool {
+	return res != nil && res.Data != nil &&
+		(res.Data.Value != nil || res.Data.Error != nil)
+}
+
 const (
 	// ExecutionQueryNodeType represents a node that will execute
 	// a query. It can be notified by datapoint nodes, representing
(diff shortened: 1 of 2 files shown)


policy/executor/internal/nodes.go
```diff
@@ -249,6 +256,9 @@ func (nodeData *DatapointNodeData) recalculate() *envelope {
 
 	nodeData.invalidated = false
 
+	if nodeData.dump {
```
jaym: better to make this Trace

policy/executor/internal/nodes.go
```diff
@@ -465,6 +483,14 @@ func (nodeData *ReportingJobNodeData) consume(from NodeID, data *envelope) {
 		if !ok {
 			panic("invalid datapoint report")
 		}
+		// If the previously-reported result was nil (from short-circuit) and
+		// we're now getting real data, reset completed so the score can be
+		// recalculated with the actual data.
+		if nodeData.completed && dp.res != nil && dp.res.Data != nil &&
```
jaym: pull out a helper for this check

policy/executor/graph.go
```diff
@@ -55,15 +55,17 @@ func ExecuteResolvedPolicy(ctx context.Context, runtime llx.Runtime, collectorSv
 		return err
 	}
 
-	ge.Debug()
+	ge.Debug("resolved-policy")
 
 	return ge.Execute()
 }
```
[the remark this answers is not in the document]
jaym: it is intentional

policy/executor/internal/nodes.go
```diff
@@ -214,16 +215,22 @@ func (nodeData *DatapointNodeData) initialize() {
...
 	if data == nil || data.res == nil {
 		// This can be triggered with no data by the execution query nodes. These
 		// messages are not the ones we care about
 		return
 	}
+	if nodeData.isReported {
+		// Allow a real result to override a nil one. This handles the case
+		// where short-circuit evaluation (e.g. &&) reports nil for an
+		// unevaluated branch, and a later query reports the actual value.
+		if nodeData.res != nil && nodeData.res.Data != nil &&
```
preslavgerchev: this `if` eval. is hard to read. can we extract to vars and then run the if check on those?

PR #2378, opened 2026-04-16
✨ Add MQL queries to risk factors.

VasilSirakov:


arlimus (approved):
Kudos for expanding the risk factors to support data queries, now we can collect a lot more context about these!

PR #2400, opened 2026-04-20
Wire ScoredRiskFactor.Data through the BufferedCollector

preslavgerchev:
## Summary
- Populate `ScoredRiskFactor.Data` at collection time so it's present when sent upstream via `StoreResults`, not only when enriched later by the inmemory `GetReport` path.
- Unify the two accumulator maps (`risksIdx map[string]bool` and `riskData map[string]map[string]*llx.Result`) into a single `map[string]*ScoredRiskFactor` that is progressively built from both detection scores and data query results.

## Test plan
- [x] Existing `TestResolveV2_RiskFactorDataInReport` passes (full pipeline: resolve → execute → collect → report)
- [x] All `policy/...` tests pass
- [x] All `internal/...` tests pass

### DIFF
--- a/policy/executor/internal/collector.go
+++ b/policy/executor/internal/collector.go
@@ -7,6 +7,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
+	"slices"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


preslavgerchev:
Closing as this is currently not required.

PR #2405, opened 2026-04-21
Add rescoring support to GraphExecutor via ExecutionManager interface

preslavgerchev:
## Summary

Add rescoring support to the graph executor, enabling server-side score re-evaluation without query execution.

- Extract `ExecutionManager` interface from the concrete execution manager. `DefaultExecutionManager` wraps an `llx.Runtime` for scanning. `NewNoopExecutionManager` returns nil channels for rescoring
- `ExecuteResolvedPolicy` takes `ExecutionManager` instead of `llx.Runtime`
- Add `RescoreResolvedPolicy(assetMrn, resolvedPolicy, scores, scoreCollector)` — injects pre-computed scores into `ReportingJobNode`s, runs the same `Execute()` loop, and sends rolled-up results to the provided `ScoreCollector`
- `nodeData` interface gains `preseed(data *envelope)` method. `ReportingJobNodeData` uses it to set injected scores; other node types are no-ops. `InjectData(data map[string]*envelope)` on `GraphExecutor` finds nodes by `queryID` and calls `preseed` — no type assertions needed
- `envelope` stays private. Exposed via `NewEnvelope(res, score)` constructor for external callers. `WithScores` on the builder converts score maps to envelopes internally
- `CalculateReportingJobScore` — shared scoring function used by graph executor nodes. Single code path for forward-score and aggregate scoring. CONTROL remapping moved from `consume()` to scoring time
- `EnableScoreRisk` opt-in: rolls up `RiskScore` alongside `Value` through the hierarchy using the same scoring system per level
- Both entry points share `buildAndExecuteResolvedPolicy` — all options (scores, risk scoring, features, collectors) passed as parameters

## Test plan

- [x] 10 rescore tests via `RescoreResolvedPolicy`: average, worst, multi-level (CHECK→POLICY→CONTROL→FRAMEWORK→root), exceptions (Skip/Disabled/OutOfScope), CONTROL remapping, impact floor, updated scores, Value+RiskScore rollup, full hierarchy with risk, real RP fixture (139 reporting jobs)
- [x] All existing `executor/internal` tests pass (67 tests)
- [x] All existing `executor` integration tests pass
- [x] `go build ./...` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2414, opened 2026-04-23
Score rollup via graph executor (rescoring) with Producer refactor

jaym:
## Summary

This branch supersedes #2405. It carries the rescoring functionality from that PR plus a follow-up refactor that removes the fragile `preseed` mechanism and replaces the `ExecutionManager` interface with a focused `Producer` abstraction.

Two commits:

1. **Add rescoring support to GraphExecutor** (1aaf8271, from #2405) — original rescoring implementation with `ExecutionManager` interface and per-node `preseed`.
2. **Refactor GraphExecutor to use Producer abstraction; remove preseed** — replaces `ExecutionManager` with `Producer` and feeds rescore scores in as a normal batch through the same channel the runtime producer uses.

### Why the refactor

`ReportingJobNodeData.preseed` hijacked the node's `childScores` map by replacing it with a single `__preseed__` entry. The code comment on `preseed` admitted this was fragile — any later `consume()` from a real edge would panic.

The graph already has the slots needed: every parent `ReportingJobNode` has `childScores[childUuid]` populated from `rj.ChildJobs` in `addReportingJobNode`. `BatchScoreProducer` just addresses each input score at the parents listed in `leaf.Notify` (and the well-known `ScoreCollectorID`), with `from = leaf.Uuid`. The existing `consume()` path slots the score in. No node modifications, no slot reservation, no special-case scoring code.

The main-loop `DRAIN` already batches incoming envelopes before the priority queue starts processing, so emitting all rescore scores in one shot causes every `consume()` to run before any aggregate `recalculate` — exactly the desired behavior.

The refactor also splits the run-queue input out of the producer contract. `ExecutionQueryNode`s get the run queue via `builder.WithRunQueue`; producers that don't execute queries (`BatchScoreProducer`) don't need to provide one.

### Net code impact (refactor commit)

- `+278 / -264` (9 files)
- `policy/executor/internal/execution_manager.go` → renamed `producer.go`
- `policy/executor/internal/batch_producer.go` (new, ~95 lines)
- `nodeData.preseed` interface method and 7 implementations deleted
- `GraphExecutor.InjectData` deleted
- `GraphBuilder.WithScores` / `initialData` / preseed branch deleted

## Test plan

- [x] `go test ./policy/executor/...` — all pass, including the 12 `TestRescore_*` tests
- [x] `TestRescore_RealResolvedPolicy` — validates rollup against real server-computed scores from `testdata/scores.json` against `testdata/resolved_policy.json`
- [x] `go test ./policy/...` — all pass (the one failing `policy/scan/pdque/TestNextAvailableFilename` reproduces on `main` — pre-existing flake unrelated to this branch)
- [ ] End-to-end manual: `cnspec scan local` to exercise the `DefaultProducer` runtime path through `local_scanner.go`


policy/executor/internal/batch_producer.go
```diff
@@ -0,0 +1,97 @@
...
+
+// NewBatchScoreProducer builds a BatchScoreProducer from the resolved
+// policy's reporting jobs and a map of pre-computed scores keyed by
+// QrId. Honors the "root" → assetMrn rename that addReportingJobNode
+// performs so input keyed by the asset MRN finds its leaf.
+func NewBatchScoreProducer(scores map[string]*policy.Score, rjs map[string]*policy.ReportingJob, assetMrn string) *BatchScoreProducer {
+	leafByQrId := make(map[string]*policy.ReportingJob, len(rjs))
+	for _, rj := range rjs {
+		qid := rj.QrId
+		if qid == "root" {
```
preslavgerchev, line 35: we have this i think in a few places, can we extract?

policy/executor/internal/batch_producer.go
```diff
@@ -0,0 +1,97 @@
...
+	stopOnce sync.Once
+	stopChan chan struct{}
+	wg       sync.WaitGroup
+}
+
+// NewBatchScoreProducer builds a BatchScoreProducer from the resolved
+// policy's reporting jobs and a map of pre-computed scores keyed by
+// QrId. Honors the "root" → assetMrn rename that addReportingJobNode
+// performs so input keyed by the asset MRN finds its leaf.
+func NewBatchScoreProducer(scores map[string]*policy.Score, rjs map[string]*policy.ReportingJob, assetMrn string) *BatchScoreProducer {
```
preslavgerchev, line 31: could we map this a map[string]ReportingJobWithScore?
if i read it right, we expect both maps to have matching keys. we could have a util that merges the 2 maps outside and pass them in

PR #2459, opened 2026-04-30
:sparkles: Add Okta Terraform variants for 5 previously-deferred checks

tas50:
## Summary
Five parents in `content/mondoo-okta-security.mql.yaml` previously carried `# No Terraform variants: not yet implemented` comments listing a candidate Okta TF resource (left behind by #2458 as documented follow-up work). This PR delivers that follow-up — each parent now has full `terraform-hcl` / `terraform-plan` / `terraform-state` variants plus an `id: terraform` remediation block with example HCL.

| check | resource | argument |
|---|---|---|
| `okta-mfa-access` | `okta_policy_mfa_default` / `okta_policy_factor_enrollment` | resource exists |
| `okta-enforce-session-lifetime` | `okta_policy_rule_signon` | `session_idle <= 120`, `session_lifetime <= 1440` |
| `threatinsight-action-block` | `okta_threat_insight_settings` | `action = "block"` |
| `trusted-origins-https` | `okta_trusted_origin` | `origin` matches `^https://` |
| `signon-requires-mfa` | `okta_policy_rule_signon` | `mfa_required = true` (or the newer `factor_required` argument) |

The two remaining `# No Terraform variants:` comments — `limit-super-admins` and `limit-org-admins` — are intentionally untouched. Counting role assignments across the org has no clean Terraform analog: per-user `okta_user_admin_roles` HCL would only see TF-managed grants and miss admin assignments made via the API or Admin Console, so the runtime check against the live tenant remains the source of truth.

Policy version stays at `2.3.0`.

## Test plan
- [x] `cnspec policy lint content/mondoo-okta-security.mql.yaml` → valid policy bundle(s)
- [x] `inspect-policies.py` reports the 5 newly-equipped parents now satisfy `{console, terraform}`; only the two documented count-based gaps remain
- [ ] Run `cnspec scan terraform-hcl <fixture>` against a small fixture containing each new resource type and confirm the variants evaluate correctly
- [ ] Spot-check rendered remediation Markdown in the UI for the 5 parents (especially `signon-requires-mfa` because the HCL note about the deprecated `mfa_required` vs newer `factor_required` argument is non-trivial)

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -640,6 +640,7 @@ vty
 vzdump
 wafv
 wazuh
+webauthn
 webfilter
 webroot
 webui
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -480,7 +480,6 @@ rebrands
 Redir
 rediraccept
 reissuance
-rejectattr
 remoteadministrator
 removexattr
 renameat
@@ -553,7 +552,6 @@ solr
 sourcegraph
 sourceroute
 spdx
-splitlines
 spo
 SProtection
 sqladmin
@@ -642,7 +640,6 @@ wafv
 wazuh
 webauthn
 webfilter
-webroot
 webui
 widescale
 WIFI
--- a/content/mondoo-okta-security.mql.yaml
+++ b/content/mondoo-okta-security.mql.yaml
@@ -723,8 +723,8 @@ queries:
   - uid: mondoo-okta-security-okta-enforce-session-lifetime-terraform-hcl
     filters: asset.platform == "terraform-hcl" && terraform.providers.any( nameLabel == "okta" )
     mql: |
-      terraform.resources.where( nameLabel == "okta_policy_rule_signon" ).all( arguments['session_idle'] == /var/ || arguments['session_idle'] <= 120 )
-      terraform.resources.where( nameLabel == "okta_policy_rule_signon" ).all( arguments['session_lifetime'] == /var/ || arguments['session_lifetime'] <= 1440 )
+      terraform.resources.where( nameLabel == "okta_policy_rule_signon" ).all( arguments['session_idle'] == /^var\./ || arguments['session_idle'] <= 120 )
+      terraform.resources.where( nameLabel == "okta_policy_rule_signon" ).all( arguments['session_lifetime'] == /^var\./ || arguments['session_lifetime'] <= 1440 )
   - uid: mondoo-okta-security-okta-enforce-session-lifetime-terraform-plan
     filters: asset.platform == "terraform-plan" && terraform.plan.resourceChanges.contains( providerName == /okta/ )
     mql: |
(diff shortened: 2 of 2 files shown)


tas50:
Addressed in 794e9c0d:

**🟡 `/var/` regex** — Tightened the three lines this PR introduced (`session_idle`, `session_lifetime`, `mfa_required`) to `/^var\./` so they only match actual Terraform variable references and won't accidentally match a literal like `"varnish"`. The looser `/var/` form is used widely as the long-standing convention in this policy file (e.g., the existing password-settings variants — see `arguments['email_recovery'] == /var/`, `arguments['password_min_length'] == /var/`, etc.). Cleaning those up wholesale would be real scope creep into untouched code; I left them alone and queued for a separate convention-tightening PR.

**🔵 MFA-access HCL presence-only** — Intentional. The runtime MQL on this check is itself a presence/applicability check:

```mql
everyoneGroupId = okta.groups.where(profile['name'] == "Everyone").map(id)[0]
okta.policies.mfaEnroll.one(status == "ACTIVE" && conditions['people']['groups']['include'].contains(everyoneGroupId))
```

It verifies that an MFA enrollment policy *exists and applies to Everyone*, not that it actually has any factor configured with `enroll = "REQUIRED"`. The TF variants mirror that strictness by design — making them stricter ("at least one factor has `enroll = REQUIRED`") would create a semantic gap where a tenant could pass the runtime check but fail the HCL check, or vice-versa, on the same configuration. Tightening should be a co-ordinated change to both the runtime check and the TF variants in a follow-up PR (this is the "factor-required vs policy-applies" distinction).

The existing `mondoo-okta-security-enable-okta-verify-with-push-for-mfa-terraform-state` check that the reviewer pointed at is a *separate* check that specifically asserts the `okta_otp` factor is required — that lives at a different uid because it's a different control objective.

tas50:
Addressed in 38f98fc7:

**🟡 MFA-access still passes if the resource is just declared (line 636).** Tightened the HCL / plan / state variants. They no longer pass on mere existence of `okta_policy_mfa_default`; they now require that at least one strong factor on the resource has `enroll = "REQUIRED"`:

- `okta_otp` (TOTP)
- `okta_verify`
- `fido_webauthn`
- `webauthn`
- `google_otp`

Dropped the `okta_policy_factor_enrollment` alternative from the variant assertion — that resource defines per-rule enrollment, and validating it correctly is more than the runtime check itself enforces. The runtime check still catches tenants that rely solely on `okta_policy_factor_enrollment`.

**🔵 HCL `var.` asymmetry between HCL and plan/state (line 724).** Added a YAML comment above the session-lifetime HCL variant explaining the asymmetry: the HCL variant evaluates static source where a `var.something` reference's resolved value isn't known at parse time, so the variant accepts a `var.` reference as a pass-through. Plan and state variants evaluate against resolved values and intentionally omit the escape.

**🔵 Boolean `mfa_required == /^var\./` works correctly (line 822).** Quick verification: when the HCL parser encounters `mfa_required = var.foo`, it stores `arguments['mfa_required']` as a string (the textual expression `var.foo`). When it encounters `mfa_required = true`, it stores it as a bool. The expression `arguments['mfa_required'] == /^var\./ || arguments['mfa_required'] == true` covers both cases:

| HCL source | `arguments['mfa_required']` | first match | second match | result |
|---|---|---|---|---|
| `mfa_required = var.foo` | `"var.foo"` (string) | regex matches | type-mismatch (false) | **pass** |
| `mfa_required = true` | `true` (bool) | type-mismatch (false) | bool match | **pass** |
| `mfa_required = false` | `false` (bool) | type-mismatch (false) | type-mismatch (false) | **fail** ✓ |
| _no `mfa_required`, but `factor_required = true`_ | _absent_ | _absent_ | _absent_ | third clause matches → **pass** |

MQL's `==` returns false on a type mismatch rather than raising, so the disjunction reads as expected. This is also the same pattern used across the rest of this policy (e.g., `arguments['password_min_length'] == /var/` against an integer field).

I also relocated a stranded `# No Terraform variants:` comment that ended up above `signon-requires-mfa-terraform-hcl` instead of `limit-org-admins` — ruamel.yaml's comment tracking attached the comment to the original index when this PR inserted entries before it.

PR #2487, opened 2026-05-01
:sparkles: Capture self-contained scan databases for offline replay

jaym:
## Summary
- Extends the cnspec scan database (SQLite) so it carries everything needed to replay a scan offline: the inventory.Asset proto and the asset filters the scanner used in ResolveAndUpdateJobs.
- Adds a hidden `cnspec scan --output-scan-db <dir>` flag that saves a copy of each captured scan db locally instead of just uploading it.
- Plumbs a context-based filter-capture hook (`scandb.WithFilterCapture` / `scandb.CaptureFilters`) so external tooling can subscribe to filter info without bloating regular uploads.

## Why
We need self-contained, replayable scan databases for offline debugging and synthetic-load workflows. The existing scan-db format only carried `asset_mrn` in metadata, which isn't enough to call SynchronizeAssets or reproduce ResolveAndUpdateJobs.

## What's in scope
- **Schema additions (additive, backwards-compatible)**
  - `policy/scandb/scan_db.sql`: new `asset` and `asset_filters` tables. `SchemaVersion` is unchanged so existing readers keep working unchanged.
- **Public surface on `*SqliteScanDataStore`** (concrete type, no interface changes)
  - `WriteAsset` / `GetAsset`
  - `WriteAssetFilters` / `GetAssetFilters` (stores the full `policy.Mqueries` blob — the resolver compiles each filter's MQL, so code_ids alone aren't enough)
  - Both `Get*` variants return new sentinel errors (`policy.ErrAssetNotFound`, `policy.ErrAssetFiltersNotFound`) for older databases that lack the optional tables.
- **Filter-capture hook**
  - `policy/scandb/filter_capture.go`: ctx-based `FilterCaptureFunc` with `WithFilterCapture` / `CaptureFilters` helpers. Cost is zero when no consumer installs the hook.
  - `policy/scan/local_scanner.go`: scanner invokes `CaptureFilters` once per asset right before `ResolveAndUpdateJobs`.
- **Internal datalake plumbing**
  - `internal/datalakes/sqlite/WithServices` and `internal/datalakes/inmemory/WithServices` now take the full `*inventory.Asset` and a `func(ctx, *policy.LocalServices) error` callback so the enriched ctx (including filter capture) reaches the scanner. Internal-only signature change.
  - When `--output-scan-db` is set, sqlite.WithServices installs the filter-capture callback so the captured filters land in the scan db. Otherwise it's a no-op.
- **CLI**
  - `cnspec scan --output-scan-db <dir>` (hidden). Auto-enables the `UploadResultsV2` feature so the SQLite datalake path runs and produces a captured file.

## Backwards compatibility
This library is consumed downstream, so changes are additive:
- `SchemaVersion` is unchanged.
- The `ScanDataStore` interface is unchanged; the new methods live on the concrete `*SqliteScanDataStore`.
- Old databases without the optional tables read fine — `GetAsset` / `GetAssetFilters` return their respective sentinel errors and consumers handle them.
- Only `internal/` packages had their signatures changed (other modules can't import `internal/`).

## Test plan
- [x] `go test ./policy/scandb/...` (schema additions don't regress existing tests)
- [x] `go build ./...`
- [ ] `cnspec scan host example.com --output-scan-db ./seeds` produces a SQLite file under `./seeds`; `sqlite3 seeds/*.db .schema` shows the new tables; `.dump asset` and `.dump asset_filters` are non-empty.


apps/cnspec/cmd/scan.go
```diff
@@ -152,6 +156,13 @@ To manually configure a policy, use this:
 
 var scanCmdRun = func(cmd *cobra.Command, runtime *providers.Runtime, cliRes *plugin.ParseCLIRes) {
 	ctx := context.Background()
+
+	// Hidden dev flag — preserve a copy of each scan database under this dir
+	// so it can be replayed by `cnspec loadtest`.
+	if dir := viper.GetString("output-scan-db"); dir != "" {
+		sqlite.SetOutputDir(dir)
```
jaym: Can we inject this through the context instead of a global variable

internal/datalakes/sqlite/sqlite.go
```diff
@@ -18,12 +18,49 @@ import (
...
+			}
+		}
+
+		// When --output-scan-db is set, install the filter-capture hook so
+		// the scanner's ResolveAndUpdateJobs filters land in the scan db too.
+		// We deliberately gate this on outputDir so regular scans don't pay
+		// the marshalling cost or send filters to the platform.
+		scanCtx := ctx
+		if outputDir != "" {
+			scanCtx = scandb.WithFilterCapture(scanCtx, func(filters *policy.Mqueries) {
```
jaym: The filters are pretty small, lets always collect them if we're using the scandb

policy/scandb/scan_data_store.go
```diff
@@ -218,6 +220,73 @@ func (s *SqliteScanDataStore) WriteResource(ctx context.Context, resource *llx.R
...
+// compiles each filter's MQL during resolution; sending a bare code_id
+// causes ResolveAndUpdateJobs to fail with "query is not implemented".
+func (s *SqliteScanDataStore) WriteAssetFilters(ctx context.Context, filters *policy.Mqueries) error {
+	if s.readOnly {
+		return fmt.Errorf("cannot write asset filters in read-only mode")
+	}
+	if filters == nil {
+		return fmt.Errorf("cannot write nil asset filters")
+	}
+	data, err := proto.Marshal(filters)
```
jaym: I don't want to capture a proto encoded field. Lets create a table with just code_id. code_id should be unique so we do not need to keep repeating. Update `ChecksumAssetFilters` to skip `RefreshAsFilter` if we don't have the code.

internal/datalakes/sqlite/sqlite.go
```diff
@@ -18,12 +18,59 @@ import (
...
+			if err := scanDataStore.WriteAsset(ctx, asset); err != nil {
+				log.Warn().Err(err).Msg("failed to persist asset to scan data store")
+			}
+		}
+
+		// Always install the filter-capture hook on the SQLite datalake path —
+		// filter code_ids are tiny and we want them in every scan db so any
+		// downstream replay can call ResolveAndUpdateJobs without re-deriving
+		// them.
+		scanCtx := scandb.WithFilterCapture(ctx, func(codeIDs []string) {
```
jaym: Capturing filters should be part of the datalake.

`SetAssestFilters(ctx, Mqueries)`. The implementation can pull the code ids and write it to the scandb

policy/cnspec_policy.pb.go
```diff
@@ -3,7 +3,7 @@
 
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.11
+// 	protoc-gen-go v1.36.11-devel
```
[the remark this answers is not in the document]
jaym: we don't control this. One of our dependencies is deciding

PR #2512, opened 2026-05-08
:sparkles: Add Azure security checks for Container Apps and Container Instances

tas50:
## Summary

- Add three native Azure Container Apps checks: HTTPS-only ingress, managed identity for registry pulls, and digest-pinned container images.
- Add one native Azure Container Instances check that requires images come from a private Azure Container Registry rather than a public registry.
- Each check is a single self-contained query against the Azure ARM API. No Terraform variants and no `compliance/*` tags were added.

## Test plan

- [x] `go run ./apps/cnspec policy lint ./content/mondoo-azure-security.mql.yaml` passes.
- [ ] Verify the four new check UIDs surface against an Azure subscription that has Container Apps and Container Instances deployed.


tas50:
Addressed bot review feedback in fe500ba1:

- **Title/regex mismatch on `container-instance-private-registry`** — narrowed the title to "Ensure Container Instances pull images from Azure Container Registry" so it matches the `*.azurecr.io/` regex.
- **Spell-check** — added `myacr`, `repointed`, and `Sourcing` to `.github/actions/spelling/expect.txt`.

**Not addressing:** The suggestion to add `azure.subscription.containerApp.containerApps.length > 0` to the `filters` block. Every existing subscription-level Azure check in this policy uses bare `asset.platform == "azure"` (e.g., `cloudDefender.securityContacts`, `monitor.diagnosticSettings`, `network.watchers.flowLogs`, etc.) — adding a per-resource existence guard would deviate from the established convention. MQL's `.all()` returns true on empty lists, so subscriptions without Container Apps trivially pass without noise.

PR #2539, opened 2026-05-15
:sparkles: Add linter rule warning on check titles over 75 characters

tas50:
## Summary

Adds a new `query-title-too-long` lint rule that warns when a check's `title` exceeds 75 characters.

- `content/CLAUDE.md` already documents the rule ("Check `title` fields must be 75 characters or fewer"); this makes `cnspec policy lint` surface violations automatically.
- The rule is **warning-level**, so it does not fail linting on existing policies.
- It applies only to queries used as checks, counting Unicode characters (`utf8.RuneCountInString`) rather than bytes.

Running the linter against `./content` currently flags 4 existing checks that exceed the limit (2 in `mondoo-kubernetes-security`, 2 in `mondoo-vmware-vsphere-esxi`). These are surfaced as warnings and can be addressed separately.

## Test plan

- Added `internal/bundle/testdata/fail-query-title-too-long.mql.yaml` and a `fail-query-title-too-long` subtest in `lint_test.go`.
- `go test ./internal/bundle/...` passes.

### DIFF
--- a/internal/bundle/lint_rules_queries.go
+++ b/internal/bundle/lint_rules_queries.go
@@ -5,18 +5,23 @@ package bundle
 
 import (
 	"fmt"
+	"unicode/utf8"
 )
 
 // Query Rule ID Constants
 const (
 	QueryUidRuleID                         = "query-uid"
 	QueryTitleRuleID                       = "query-name" // Original ID was query-name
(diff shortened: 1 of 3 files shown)


chris-rock (approved):
Nice addition @tas50

PR #2649, opened 2026-05-28
🐛 compile querypack MQL during bundle lint

M-gre:
## Summary

- `cnspec policy lint` parsed querypacks but never compiled their MQL, because `Bundle.CompileExt` only walks `p.Policies`/`p.Queries`/risk factors — it never iterates `p.Packs`.
- As a result, invalid MQL inside a querypack (e.g. using `:` instead of `=` for block-assignments like `org { counterSum: a + b }`) passed local lint and only failed server-side during `SetBundleMap` with `expected closing '}', got ':'`.
- Fix: call `Bundle.Prepare()` before `CompileExt` in the linter, mirroring what `apps/cnspec/cmd/scan.go` already does. `ConvertQuerypacks` turns packs into `DATA_ONLY` policies whose grouped queries then flow through the existing compile loop.

## Test plan

- [x] `go test ./internal/bundle/... -count=1`
- [x] New regression test `fail-querypack-compile-error` exercises a pack-only bundle with invalid MQL and asserts `bundle-compile-error` is raised with the same message the server compiler produces.
- [x] Manual repro: pack-only bundle with `counterSum: ...` now reports `expected closing '}', got ':'` at lint time instead of passing.

### DIFF
--- a/internal/bundle/lint.go
+++ b/internal/bundle/lint.go
@@ -125,6 +125,12 @@ func LintPolicyBundle(schema resources.ResourcesSchema, filename string, data []
 
 	// check if the file is compilable
 	if policyBundleForCompilation != nil {
+		// Mirror the scan path: convert querypacks to policies and generate
+		// evidence so that pack/evidence MQL is compiled here. Without this,
+		// invalid MQL in querypacks slips past the linter and only fails on
+		// the server.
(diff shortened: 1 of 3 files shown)


arlimus (approved):
👍 nice fix, kudos for this!

PR #2773, opened 2026-06-10
⭐ executor: attach the running query to panic reports

jaym:
## What

The policy executor's run goroutine now tracks the in-flight `*llx.CodeBundle` and, when a query panics the engine, attaches `queryCodeID` + `querySource` tags to the panic report via `health.ReportPanicWithTags`.

## Why

During INC-2026-06-10-client-panic-k8s (mondoohq/mql#8319), the panic report identified *where* the engine died (`llx.arrayContainsAll`, `builtin_array.go:935`) but not *which query* it was evaluating. Root-causing it to the su-restriction check took a policy-release diff plus compiled-chunk-ref forensics; with these tags the report itself names the query. Slow-query reports already carry code ID + query text — this brings panic reports to parity.

The recover stays at the goroutine top so the stacktrace is preserved intact; the deferred reporter just snapshots the tracked bundle at crash time.

## Dependencies

- ⚠️ **Depends on mondoohq/mql#8330** (`ReportPanicWithTags` / `QueryPanicTags`) — needs a `go.mondoo.com/mql/v13` bump here once that merges and releases. Verified compiling against the mql branch via a local go.work.
- Companion server PR maps the tags to the existing `mondoo.query.text` / `mondoo.query.code_id` obslog fields on `panic.received` events.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -49,7 +49,7 @@ require (
 	github.com/tliron/glsp v0.2.2
 	github.com/zclconf/go-cty v1.18.1
 	go.mondoo.com/mondoo-go v0.0.0-20260610003616-faefada3536e
-	go.mondoo.com/mql/v13 v13.22.1
+	go.mondoo.com/mql/v13 v13.22.2-0.20260610222949-17d6d58f1814
 	go.mondoo.com/ranger-rpc v0.8.0
 	gocloud.dev v0.46.0
 	golang.org/x/sync v0.21.0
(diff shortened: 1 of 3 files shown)


jaym:
mondoohq/mql#8330 merged — bumped `go.mondoo.com/mql/v13` to the pseudo-version at its merge commit (`v13.22.2-0.20260610222949-17d6d58f1814`) and marked ready for review. Happy to re-bump to the next tagged mql release if that is preferred. Also includes the same review fix as the mql side: `current` is cleared once `executeCodeBundle` returns so panics in non-query code are not misattributed to the last completed query.

PR #2775, opened 2026-06-11
🐛 linux: fix incorrect and incomplete remediations across the security policy

tas50:
## Summary

A full review of every remediation (cli, bash, and ansible entries) in the Linux Security policy, comparing each against its check's MQL to verify that following the remediation actually makes the check pass and does not harm the host. Bumps the policy to 2.7.2. No check logic (`mql`/`filters`) was changed.

All 477 embedded bash scripts pass `bash -n` and all 115 embedded ansible playbooks parse as valid YAML; `cnspec policy lint` passes.

## Dangerous remediations removed

- **root-group-is-empty (ansible)**: the playbook looped every non-root user through `ansible.builtin.user` with `state: absent`, which deletes the accounts instead of removing them from the root group. It also referenced `ansible_users`, a fact that does not exist. Replaced with `gpasswd -d` driven by actual root group membership.
- **shadow-group-is-empty (ansible)**: same `state: absent` account-deletion bug and undefined `ansible_users` variable. Same `gpasswd` replacement. The bash script also hardcoded GID 42 for the shadow group and silently no-opped on a literal `NEWGROUP` placeholder behind `|| true`; it now resolves the GID via `getent` and the placeholder is an explicit instruction.
- **access-to-the-su-command-is-restricted (bash)**: the script applied both mutually exclusive options at once, requiring membership in the wheel group and an empty `sugroup`, which locks every user out of `su`. It also appended a raw `wheel:x:10:root,<user list>` line to `/etc/group`, corrupting the group database. Now it configures `pam_wheel.so` idempotently and presents the lockdown option separately. The cli's `sed` on `/etc/group` (which assumed GID 10 and a bare `root` member list) is replaced with `usermod -aG wheel`.
- **ssh-access-is-limited (bash, cli)**: the script wrote `AllowUsers user1 user2`, `AllowGroups`, `DenyUsers`, and `DenyGroups` with placeholder values verbatim, which denies SSH access to every real account. Now it sets a single `AllowGroups` with an explicit warning to include all administrative accounts first, validates with `sshd -t`, and the cli explains the lockout risk.
- **system-accounts-are-non-login (bash)**: the loop only exempted root, so it overwrote the shells of `sync`, `shutdown`, and `halt`, which the check itself deliberately exempts. The ansible playbook used undefined `ansible_users` and the invalid `state: locked` (not a valid `user` module state); it now finds accounts via the same awk filter as the check and uses `password_lock: true`.

## Remediations that could not make their check pass

- **window-system-is-not-installed (Debian)**: `apt-get purge xserver-xorg xserver-common` leaves `xserver-xorg-core` and the driver packages installed, but the check requires `packages.none(name == /^xserver/)`. Now purges `'^xserver'` (cli/bash) and `xserver*` (ansible).
- **mail-transfer-agent-local-only (bash)**: the Postfix `sed` only matched `inet_interfaces = ` with spaces and did nothing when the directive was absent (Postfix then defaults to `all`); replaced with `postconf -e`. The Exim `sed` required spaces around `=` but Debian writes `dc_local_interfaces='...'` without them, so it never matched; pattern fixed. Added Sendmail loopback-binding steps to the cli since the check also fails on any non-loopback port 25 listener.
- **chrony-time-source (cli, ansible)**: both hardcoded `/etc/chrony.conf` and the `chronyd` unit. On Debian, chrony reads `/etc/chrony/chrony.conf`, so the remediation wrote a file the daemon never reads while the scanner read the new file and reported a pass. Both now select the correct path and unit per distro; the bash restart got the `chrony` fallback.
- **gpgcheck-is-globally-activated**: on dnf systems `/etc/yum.conf` is a symlink to `/etc/dnf/dnf.conf`, and `sed -i` (and module-based file replacement) breaks the symlink, so the scanner sees `gpgcheck=1` while dnf keeps reading the unchanged real file. All three entries now target `/etc/dnf/dnf.conf` first.
- **apt-repositories-enforce-signature-verification**: the regex `\[[^]]*trusted=yes[^]]*\]` deleted the entire option block including `arch=` and `signed-by=`, and no entry handled deb822 `.sources` files (`Trusted: yes`), which the check parses and which are the Ubuntu 24.04 default. Now removes only the trusted flag, cleans empty brackets, covers `.sources` files, and includes `/etc/apt/sources.list` in the playbook.
- **journald checks (forward-to-rsyslog, compress, persistent storage)**: the checks require the setting inside the `[Journal]` section, but the bash scripts appended to the end of the file and the playbooks could create a file with no section header, producing settings journald ignores. All entries now ensure the `[Journal]` header exists and the cli says where the line must go.
- **permissions-on-ssh-private-host-key-files**: all three entries used `chown root:ssh_keys`, a group that only exists on RHEL-family systems, so the commands fail on Debian, Ubuntu, and SUSE. The check only validates permission bits; now uses `root:root` with mode 0600, which passes everywhere.
- **permissions on /etc/shadow, /etc/gshadow, /etc/shadow- **: `chown root:shadow` fails on RHEL-family systems that have no `shadow` group, aborting the bash scripts under `set -e`. All entries are now distro-aware (Debian `root:shadow` 0640, RHEL `root:root` 0000), and the gshadow ansible entry no longer contradicts its own cli entry. The shadow tmpfiles `sed` pattern also matched the `/etc/shadow-` entry; anchored it.
- **permissions-on-etcsshsshd-config (cli, bash)**: `chmod og-rwx` leaves a user execute bit in place, but the check requires `user_executable == false`; now `chmod u-x,og-rwx`.
- **only-strong-ciphers (all entries)**: the check's allowed list is version dependent (OpenSSH 6.x permits only the CTR ciphers), but the remediation always wrote the full six-cipher list, which can never pass on 6.x and stops sshd from starting on pre-6.5. The bash script now selects by version, and cli/ansible document both lists.
- **auditing-prior-to-auditd (GRUB)**: the cli edited `GRUB_CMDLINE_LINUX_DEFAULT` while bash/ansible edited `GRUB_CMDLINE_LINUX` (now aligned). On RHEL 8+ the active kernel command line lives in `grubenv`/BLS entries that `grub2-mkconfig` does not reliably update, and the check reads `grubenv` first, so all three entries now add `grubby --update-kernel=ALL --args="audit=1"`. The bash EFI branch that ran `grub2-mkconfig` against every `grub.cfg` under `/boot/efi/EFI` (overwriting protected stub configs on RHEL 8+) is replaced with the grubby path.
- **login-and-logout-events (bash)**: the Debian `faillog` rule was gated on `[[ -d /var/log/faillog ]]`, a directory test against a regular file that is always false, so the rule the Debian check requires was never written. Now gated on the distro family, matching the per-distro checks.

## Scripts and playbooks that error out before remediating

- **Alternative-unit service checks (dns, ftp, http, imap/pop3, samba, http proxy, nis, talk)**: each check probes multiple mutually exclusive unit names (`named`/`bind9`, `smb`/`smbd`, `vsftpd`/`proftpd`/`pure-ftpd`, ...), but the bash scripts ran `set -e` with unconditional `systemctl stop` for every name. `systemctl stop` exits nonzero for a unit that does not exist, so on any real host the script aborted at the first absent unit, often before touching the daemon that is actually running. The ansible loops failed the host the same way. Bash scripts now iterate with a `list-unit-files` guard; playbooks tolerate only the missing-unit error via `failed_when`. The cli entries note that errors for absent units are expected.
- **nfs-and-rpc-are-not-enabled**: same aborting pattern (hosts commonly have `rpcbind` without `nfs-server`), plus `rpcbind.socket` was never handled, so socket activation restarted rpcbind after the stop. All entries now cover the socket and mask the units.
- **secure-icmp-redirects (bash)**: the script ran `sysctl -w` on `net.ipv6.conf.*.secure_redirects`, keys that do not exist in any kernel (`secure_redirects` is IPv4-only), so under `set -e` it always exited with an error after the real work. Lines removed.
- **auditd.conf checks (storage size, keep_logs, disk-full actions)**: `systemctl restart auditd` is refused on RHEL-family systems because the unit sets `RefuseManualStop=yes`, so the bash scripts aborted and the playbooks failed. All entries now use `systemctl restart auditd 2>/dev/null || service auditd restart`.
- **rsyslog-default-file-permissions (bash)**: the unguarded `for file in /etc/rsyslog.d/*.conf` glob either created a junk file named `*.conf` or aborted when the directory was missing; guarded.

## Ansible playbook bugs

- **Sysctl checks (all 11 network checks)**: every playbook looped `net.ipv4.route.flush` (and v6) through the sysctl module with `state: present`. The flush key is write-only, so the module's read-back verification fails the play, and the one-shot flush directive was persisted into `/etc/sysctl.d/`, flushing the route cache on every reload and boot. Removed from all loops; the tcp-syncookies and IPv6 router-advertisement bash scripts that persisted the same directive into config files are fixed the same way.
- **ciphers / MACs / KEX playbooks**: `register:` was indented inside the `lineinfile` module arguments and `when:` inside the `systemd` arguments, so every run fails with unsupported-parameter errors. Dedented to task level; the KEX value also gets `| trim` to strip whitespace from the folded Jinja expression.
- **audit-log-storage-size**: the playbook wrote the literal string `max_log_file = <MB>` into `auditd.conf`, an invalid value that prevents auditd from starting. Now a documented `max_log_file_mb` variable.
- **system-disabled-when-audit-logs-full**: `blockinfile` appended duplicate `space_left_action`/`admin_space_left_action` settings after the defaults that ship in the file; now `lineinfile` per key.
- **sudoers-scope**: `state: reloaded` SIGHUPs auditd, which re-reads `auditd.conf` but never loads new rules from `rules.d`; now `augenrules --load` plus the immutable-mode reboot warning the sibling playbooks have.
- **network-environment events**: the `failed_when` referenced `audit_immutable`, a variable never defined, so any real `augenrules` failure surfaced as an undefined-variable error. Replaced with the standard immutable-mode check. Also fixed the cli's copy-pasted "for SELinux:" lead-in from the MAC-policy check.
- **DCCP module**: the `copy` task's `content:` was missing the `|` block scalar, so YAML folded the two modprobe lines into one invalid line and the check kept failing. Sibling module checks were already correct.
- **filesystem-integrity (AIDE)**: the playbook scheduled a check timer without installing AIDE or initializing its database, so the timer fired a service that fails immediately; it now installs, initializes, and moves the database into place. The cli was missing the `aide.db.new.gz` move in both procedures, and the bash cron line hardcoded `/usr/sbin/aide`, which is `/usr/bin/aide` on Debian; the cron entry now resolves the binary with `command -v`.
- **nfs-exports-root-squash**: the playbook only edited `/etc/exports` while the check and the other entries cover `/etc/exports.d/`; now both.
- **shadow-group / root-group / system-accounts / timesyncd**: covered above.

## Hardening and clarity improvements

- **Socket-activated services (avahi, cups, rsync)**: stop/disable/mask now include the activation units (`avahi-daemon.socket`, `cups.socket`, `cups.path`, `rsyncd.socket`) that would otherwise restart the daemon on the next request.
- **All sshd configuration checks**: cli entries now note that drop-ins under `/etc/ssh/sshd_config.d/` win over the main file (sshd uses the first value it reads, and the `Include` is at the top), with a `grep` to find conflicts; this is why editing `sshd_config` can silently do nothing on Ubuntu 22.04+, Debian 12+, and RHEL 9+. Every entry that edits sshd configuration now validates with `sshd -t` before restarting, and the cipher, MAC, idle-timeout, login-grace-time, access-limited, and banner cli entries gained the previously missing restart instructions.
- **core-dumps cli**: added the `systemd-coredump` (`Storage=none`, `ProcessSizeMax=0`) step that the check conditionally requires and that only the bash and ansible entries covered.
- **prelink**: all entries now run `prelink -ua` before package removal (binaries otherwise stay modified on disk, corrupting AIDE baselines) and gained SLES/openSUSE coverage to match sibling checks.
- **timesyncd cli**: the bare `sed` did nothing when no `NTP=`/`#NTP=` line existed; now appends the setting when missing, matching the bash entry.
- **ssh-idle-timeout bash**: fixed an intro sentence copy-pasted from the LoginGraceTime check.

## Known issue not addressed here (check logic, not remediation text)

The DAC-permission-modification and unsuccessful-file-access checks unconditionally require the `chmod`/`chown` and `creat`/`open` syscalls in their audit rules, but those syscalls do not exist on aarch64 and rules naming them fail to load, so these checks cannot pass on ARM no matter what the remediation says. The file-deletion check already has an ARM carve-out in its MQL; these two need the same treatment in a follow-up, since fixing it changes check semantics rather than documentation.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -324,6 +324,7 @@ Firewalling
 firstuser
 FLUSHALL
 fmask
+fnmatch
 forti
 fortianalyzer
 fortiguard
@@ -402,6 +403,7 @@ ikev
 imds
 impersonatable
 inboxes
+initscripts
 inodes
 insertafter
 installable
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -107,6 +107,7 @@ breakglass
 bsdutils
 bsudo
 btmp
+btrusted
 bucketpolicychanges
 Burstable
 bypassable
@@ -242,6 +243,7 @@ disableforwarding
 dlp
 dlq
 dnd
+dnl
 DNSKEY
 docdb
 dotfile
@@ -533,7 +535,6 @@ networkaclchanges
 networkdevice
 networkdiscovery
 networkgatewaychanges
-NEWGROUP
 newkey
 nexthop
 nft
@@ -637,6 +638,7 @@ portgroup
 posix
 postconf
 postgre
+prelinked
 pricings
 privateca
 privkey
@@ -727,6 +729,7 @@ securepassword
 SECURITYADMIN
 securitygroupchanges
 sef
+sendmail
 sentinelagent
 sentineld
 sentinelone
@@ -783,7 +786,6 @@ stylesheet
 subnetted
 subnetting
 subpaths
-sudoer
 sudolog
 sugroup
 suki
--- a/content/mondoo-linux-security.mql.yaml
+++ b/content/mondoo-linux-security.mql.yaml
@@ -530,7 +530,9 @@ queries:
                   when: not aide_db.stat.exists
 
                 - name: Copy the new AIDE database into place
-                  ansible.builtin.command: mv /var/lib/aide/aide.db.new.gz /var/lib/aide/aide.db.gz
+                  ansible.builtin.command:
+                    cmd: mv /var/lib/aide/aide.db.new.gz /var/lib/aide/aide.db.gz
+                    creates: /var/lib/aide/aide.db.gz
                   when: not aide_db.stat.exists
 
                 - name: Create systemd service unit for AIDE check
@@ -6362,10 +6364,20 @@ queries:
                   register: auditd_logsize_changed
 
                 - name: Restart auditd service to apply changes if not immutable
-                  ansible.builtin.command: service auditd restart
+                  ansible.builtin.service:
+                    name: auditd
+                    state: restarted
                   when:
                     - immutable_check.rc != 0
                     - auditd_logsize_changed.changed
+                    - ansible_facts['os_family'] != 'RedHat'
+
+                - name: Restart auditd on RHEL-based systems where systemd refuses manual restarts
+                  ansible.builtin.command: /usr/libexec/initscripts/legacy-actions/auditd/restart
+                  when:
+                    - immutable_check.rc != 0
+                    - auditd_logsize_changed.changed
+                    - ansible_facts['os_family'] == 'RedHat'
 
                 - name: Warn if reboot is required due to immutable audit config
                   ansible.builtin.debug:
@@ -6495,10 +6507,20 @@ queries:
                   register: auditd_logfile_action_changed
 
                 - name: Restart auditd service to apply changes if not immutable
-                  ansible.builtin.command: service auditd restart
+                  ansible.builtin.service:
+                    name: auditd
+                    state: restarted
                   when:
                     - immutable_check.rc != 0
                     - auditd_logfile_action_changed.changed
+                    - ansible_facts['os_family'] != 'RedHat'
+
+                - name: Restart auditd on RHEL-based systems where systemd refuses manual restarts
+                  ansible.builtin.command: /usr/libexec/initscripts/legacy-actions/auditd/restart
+                  when:
+                    - immutable_check.rc != 0
+                    - auditd_logfile_action_changed.changed
+                    - ansible_facts['os_family'] == 'RedHat'
 
                 - name: Warn if reboot is required due to immutable audit config
                   ansible.builtin.debug:
@@ -6655,10 +6677,20 @@ queries:
                   changed_when: false
 
                 - name: Restart auditd service to apply changes if not immutable and config changed
-                  ansible.builtin.command: service auditd restart
+                  ansible.builtin.service:
+                    name: auditd
+                    state: restarted
                   when:
                     - immutable_check.rc != 0
                     - auditd_params_changed.changed
+                    - ansible_facts['os_family'] != 'RedHat'
+
+                - name: Restart auditd on RHEL-based systems where systemd refuses manual restarts
+                  ansible.builtin.command: /usr/libexec/initscripts/legacy-actions/auditd/restart
+                  when:
+                    - immutable_check.rc != 0
+                    - auditd_params_changed.changed
+                    - ansible_facts['os_family'] == 'RedHat'
 
                 - name: Warn if reboot is required due to immutable audit config
                   ansible.builtin.debug:
@@ -14089,10 +14121,12 @@ queries:
 
             SHADOW_GID=$(getent group shadow | cut -d: -f3)
             echo "Changing primary group for users whose primary group is shadow"
-            for user in $(awk -F: -v gid="$SHADOW_GID" '$4 == gid {print $1}' /etc/passwd); do
-              # Replace users with the desired primary group before running this loop
-              usermod -g users "$user"
-            done
+            while IFS=: read -r user _ _ gid _; do
+              if [ "$gid" = "$SHADOW_GID" ]; then
+                # Replace users with the desired primary group before running this loop
+                usermod -g users "$user"
+              fi
+            done < /etc/passwd
             ```
         - id: ansible
           desc: |
--- a/content/mondoo-linux-security.mql.yaml
+++ b/content/mondoo-linux-security.mql.yaml
@@ -496,7 +496,11 @@ queries:
             fi
 
             # Add cron job for daily check at 5AM if not already present
-            AIDE_BIN="$(command -v aide)"
+            AIDE_BIN="$(command -v aide || true)"
+            if [ -z "$AIDE_BIN" ]; then
+              echo "ERROR: aide not found" >&2
+              exit 1
+            fi
             CRON_LINE="0 5 * * * $AIDE_BIN --check"
             (crontab -l 2>/dev/null | grep -v -F "aide --check" ; echo "$CRON_LINE") | crontab -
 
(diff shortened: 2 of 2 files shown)


tas50:
Addressed the CI failures and review feedback in a2b6fe8e:

**Fixed**
- `validate-bash`: the shadow-group primary-group loop is now a `while IFS=: read` over `/etc/passwd` instead of `for user in $(awk ...)`, resolving SC2013.
- `validate-ansible`: the three `command: service auditd restart` tasks are split into the `service` module for distros where systemd permits the restart, plus the `/usr/libexec/initscripts/legacy-actions/auditd/restart` command on RHEL, which is the path `service` redirects to there and the reason the plain module cannot restart auditd on RHEL. This resolves `command-instead-of-module` without losing the RHEL workaround.
- Spell check: added `btrusted`, `dnl`, `prelinked`, and `sendmail` to the allowlist; removed `NEWGROUP` and `sudoer`, which no longer appear in content.
- AIDE playbook: the database move now carries `creates: /var/lib/aide/aide.db.gz`, so reruns are idempotent.

**Declined, with reasons**
- `AIDE_BIN` empty-string guard: the script runs under `set -e`, and a failing `command -v aide` in the assignment terminates the script before the cron line is built, so the broken-cron scenario cannot occur.
- `apt` module glob `xserver*`: `ansible.builtin.apt` documents fnmatch wildcard support in `name`, so the glob is valid; the suggested shell task would reintroduce the lint problem this PR fixes elsewhere.
- `sshd -v` version probe: `sshd` prints its version string in the usage output on stderr, which the existing key-exchange script in this policy has relied on for some time; the cipher script intentionally uses the identical pattern, and both pass validation.
- `/usr/sbin/nologin` vs `/sbin/nologin`: `/sbin` is a symlink into `/usr/sbin` on every distro with the merged-usr layout, which covers RHEL 7 and later, so the single path resolves everywhere this policy targets.

PR #2778, opened 2026-06-11
🧹 spelling: add logouts to expect list

tas50:
PR #2776 introduced the word `logouts` in the Linux policy check descriptions (`content/mondoo-linux-security.mql.yaml:7112`) without adding it to the spelling expect list. The spell-check workflow now fails on every branch containing that commit. This adds `logouts` to `.github/actions/spelling/expect.txt` in sorted position.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -471,6 +471,7 @@ logingracetime
 logingracetime
 logname
 logons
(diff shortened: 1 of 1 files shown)


tas50:
Superseded — `logouts` landed in main via #2779/#2765, so this is no longer needed.

PR #2784, opened 2026-06-11
🐛 oci: fix incorrect and incomplete remediations across the security policy

tas50:
## Summary

A full review of every remediation entry in the OCI Security policy: 78 checks with console, cli, and terraform guidance compared against the check logic of every variant (live OCI API and Terraform). 22 findings were verified and fixed; reviewers verified commands against the OCI CLI reference and the oracle/oci Terraform provider documentation. Bumps the policy patch version.

Same review question as the AWS (#2780), GCP (#2781), Azure (#2782), and Kubernetes (#2783) passes: if a user follows the remediation exactly, does the check pass afterward, without breaking anything else?

## A check that could never pass, reworked

**IAM auth tokens**: the check required `expires != null`, but OCI auth tokens cannot carry an expiration date at all — Oracle documents that tokens do not expire, and neither the console dialog nor `oci iam auth-token create` accepts one — so every tenancy with an active token failed permanently and the remediation promised a setting that does not exist. The check now asserts active tokens were created within the last 90 days, matching the CIS OCI rotation control, and the title, description, audit, and remediation all describe rotation consistently.

## Recreate-and-migrate guidance replaced with supported in-place updates

- **OKE API endpoint**: `oci ce cluster update-endpoint-config --is-public-ip-enabled false` disables the public IP on a VCN-native cluster in place; the entries claimed the cluster had to be recreated.
- **OKE node pool subnets**: the placement configuration is editable; workers are recreated in the new subnets after a cordon and drain, with no replacement pool needed.
- **Autonomous Database network access**: serverless databases switch from public to private endpoint via Update network access or `oci db autonomous-database update`; the entries instructed provisioning a replacement database and migrating.
- **DB system and database CMK**: the entries told users to rotate keys (which only works on databases already using customer-managed keys) or restore from backup; the real paths are `configure-key`, `migrate-vault-key`, and the console's Administer Encryption Key with Change Key Management Type.

## Commands that fail or do something else

- `oci compute instance update --metadata '{}'` is rejected by OCI because metadata updates cannot add, change, or remove the SSH key or user data entries; the entry now retrieves the map and resubmits it with only credential keys removed.
- The stale-users "disable" command only applied a defined tag; replaced with credential removal since the CLI cannot disable users.
- The Vault secret rotation command configured version expiry rules; the check reads the rotation configuration, now set via `--rotation-config`.
- Missing required parameters that fail the command outright: `--vnics` on container instance create, `--compartment-id` on the cache backup, `--display-name` on the security zone update, and the rotation interval details when enabling KMS auto rotation on a key without a schedule.
- The retention rule lock example used a timestamp now in the past, which OCI rejects; locked rules also can only have their duration increased, the opposite of what the console steps claimed.

## Safety warnings added

Four VCN security list entries used `--ingress-security-rules` with a single-rule example; that parameter replaces the entire ingress rule set, so following them verbatim deleted every other rule. Each entry now says to export the current rules and include everything being kept. The Functions configuration entry had the same whole-map replacement hazard.

## Consistency with the checks

The cross-tenancy policy entries showed narrowed `Endorse`/`Admit` statements as acceptable end states, but the checks fail any policy containing those statements; the entries now show in-tenancy alternatives and state that retained cross-tenancy grants need a documented exception. The KMS auto-rotation Terraform queries accepted a rotation details block with rotation still disabled; they now also require `is_auto_rotation_enabled == true`, and the snippet sets it.

## Known vacuous check, deliberately untouched

The certificate-authority customer-managed-key check can never fail: OCI requires a Vault KMS key to create any CA, in both the API and the Terraform provider. Strengthening it (for example, asserting HSM protection of the CA key) needs a provider field that does not exist yet, so the check is left as is and flagged here for a maintainer decision.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -292,6 +292,7 @@ eventhub
 Exadata
 exampleadb
 exampledb
+examplenamespace
 exampleregistry
 examplestorage
 examplestorageaccount
@@ -388,6 +389,7 @@ huggingface
 hvm
 hwaddr
 Iaa
+iad
 iampolicychanges
 iap
 Idempotently
@@ -860,6 +862,7 @@ uvx
 valkey
 VALUEX
 VAULTNAME
+vaultsecret
 vcpus
 vda
 vertexai
--- a/content/mondoo-oci-security.mql.yaml
+++ b/content/mondoo-oci-security.mql.yaml
@@ -2987,7 +2987,7 @@ queries:
             Issue a replacement token and record its creation date so it can be rotated again within 90 days:
 
             ```bash
-            oci iam auth-token create --user-id <user-ocid> --description "Rotated on 2026-06-11"
+            oci iam auth-token create --user-id <user-ocid> --description "Rotated on $(date +%Y-%m-%d)"
             ```
         # No Terraform remediation: auth token expiry and creation timestamps are runtime-only fields; the `oci_identity_auth_token` Terraform resource exposes neither `expires` nor `time_created` as configurable arguments.
     refs:
@@ -5877,7 +5877,7 @@ queries:
 
               containers {
                 display_name = "app"
-                image_url    = "iad.ocir.io/mytenancy/app:1.4.2"
+                image_url    = "iad.ocir.io/examplenamespace/app:1.0.0"
               }
             }
             ```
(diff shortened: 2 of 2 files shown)


tas50:
Addressed in 65a4fb30: the rotation description now uses `$(date +%Y-%m-%d)` instead of a literal date, and the retention lock example uses a nearer timestamp.

On the Terraform `image_url`: angle-bracket placeholders cannot be used inside HCL strings here for two mechanical reasons — the snippet validator rewrites `<placeholder>` tokens even inside quoted strings, producing invalid HCL, and the check's own regex requires a literal `*.ocir.io` host, which an interpolated or placeholder value can never match during static analysis. The value is now `iad.ocir.io/examplenamespace/app:1.0.0` with the note explaining the values must be replaced, which keeps the example obviously exemplary while remaining valid for both the validator and the check.

PR #2861, opened 2026-06-17
fix(policy): make framework graph checksum independent of framework-map order

jaym:
## What

`Framework.updateGraphChecksums` folds each `FrameworkMap`'s local checksums into the graph checksums with an **order-sensitive** rolling hash, but iterates `f.FrameworkMaps` in slice order — while `Dependencies` in the same function are sorted by `Mrn` first. Framework maps are loaded without a stable `ORDER BY`, so a framework with **≥2 maps** computes a different `graph_content_checksum` / `graph_execution_checksum` depending on row order.

This sorts `f.FrameworkMaps` by `Mrn` before folding, mirroring the existing `Dependencies` sort.

## Why

The non-deterministic framework checksum makes the server-side PolicyBundle rebuild persist-CAS (which compares the stored framework `graph_content_checksum` column against a freshly recomputed value) **never converge** for spaces with a compliance framework: each rebuild recomputes a differently-ordered checksum, mismatches → drift, and the workflow exhausts and re-loops indefinitely.

## Test

`TestFrameworkGraphChecksum_StableAcrossFrameworkMapOrder` builds the same framework with its maps in two different input orders and asserts equal graph checksums. Verified it **fails without** the sort and **passes with** it. Existing `Framework`/`Checksum`/`Compliance` tests still pass.

### DIFF
--- a/policy/framework.go
+++ b/policy/framework.go
@@ -303,6 +303,15 @@ func (f *Framework) updateGraphChecksums(
 			AddUint(uint64(dep.Action))
 	}
 
+	// The graph checksum is an order-sensitive rolling hash, so framework maps
+	// must be folded in a deterministic order — exactly like Dependencies are
+	// sorted above. The maps are loaded without an explicit ORDER BY, so their
+	// slice order is not stable across reads; folding them unsorted makes
+	// graph_content_checksum non-deterministic for a framework with >=2 maps.
+	sort.Slice(f.FrameworkMaps, func(i, j int) bool {
+		return f.FrameworkMaps[i].Mrn < f.FrameworkMaps[j].Mrn
+	})
+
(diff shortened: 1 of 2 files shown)


jaym:
Closing — not the right fix. The rebuild drift this targeted turned out to be a stale/unreconciled framework graph_content_checksum column (the recompute and the stored column stably disagree), not framework-map order non-determinism: telemetry showed identical expected/actual across multiple separate runs, and a manual framework re-heal reconciled it. The real fix is server-side self-heal (mark the framework invalidated on a framework-GCC persist drift so the next read recomputes+rewrites the column). The map-sort is valid hardening but not the cause here; closing to avoid implying it's the fix. Can revisit as standalone hardening later.

PR #3003, opened 2026-07-03
✨ azure: add terraform-hcl variants for 4 checks

tas50:
## Summary

Adds `-terraform-hcl` variants for four Azure security checks so they evaluate Terraform HCL assets in addition to the live Azure API. Each variant appends a `-terraform-hcl` entry to the parent's `variants:` list plus a new query block (filters + mql).

| Check | Target resource |
| --- | --- |
| `mondoo-azure-security-cosmosdb-private-endpoints` | `azurerm_private_endpoint` with a `private_service_connection` targeting a Cosmos DB subresource (Sql/MongoDB/Cassandra/Gremlin/Table) |
| `mondoo-azure-security-redis-cmk-encryption` | `azurerm_managed_redis` with a `customer_managed_key` block (CMK is Enterprise/Managed tier only) |
| `mondoo-azure-security-frontdoor-origin-https` | `azurerm_cdn_frontdoor_route` with `forwarding_protocol == "HttpsOnly"` |
| `mondoo-azure-security-iot-hub-diagnostic-settings-enabled` | `azurerm_monitor_diagnostic_setting` whose `target_resource_id` points at an `azurerm_iothub` |

## Validation

- `cnspec policy lint ./content/mondoo-azure-security.mql.yaml` → valid bundle.
- Each variant was validated with a local per-check pass/fail Terraform fixture harness (not committed).

### DIFF
--- a/content/mondoo-azure-security.mql.yaml
+++ b/content/mondoo-azure-security.mql.yaml
@@ -16557,8 +16557,6 @@ queries:
       bicep.resources.where(type == "Microsoft.DocumentDB/databaseAccounts").all(
         properties["ipRules"] != empty
       )
-  # No Terraform variants: Cosmos DB private-endpoint correlation across resources is not expressible cleanly
-  # No Terraform remediation: same reason
   - uid: mondoo-azure-security-cosmosdb-private-endpoints
     title: Ensure Cosmos DB accounts have private endpoints configured
     impact: 70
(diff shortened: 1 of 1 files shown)


tas50:
Thanks. Applied the same rule — a Terraform variant must match its cloud sibling or be removed:

- **cosmosdb-private-endpoints** — **removed.** The Azure check verifies each Cosmos DB account has a private endpoint connection. In HCL a private endpoint is a separate `azurerm_private_endpoint` resource tied to an account only by reference, so the variant could only assert that *some* Cosmos-targeted endpoint exists, not per-account coverage. Restored the pre-existing `# No Terraform variant ... not expressible cleanly` comment.
- **iot-hub-diagnostic-settings-enabled** — **removed.** As you noted, `.any()` is weaker than the native `iotHubs.all(diagnosticSettings != empty)`; per-hub correlation across a separate `azurerm_monitor_diagnostic_setting` cannot be expressed in HCL. Left an explanatory comment.
- **redis-cmk-encryption** — **kept.** `blocks.where(type == "customer_managed_key") != empty` is the presence idiom already used 60+ times in this file; `[] == empty` is true in MQL so `!= empty` correctly tests a non-empty list (null/empty-safe). It maps directly to the sibling and is harness-validated.
- **frontdoor-origin-https** — kept (direct field mapping, harness-validated).

PR #3007, opened 2026-07-03
✨ content: add IaC-variant test framework + Dockerfile & Kubernetes suites

tas50:
## What

Adds a **content IaC-variant test framework** that validates policy checks against per-check pass/fail fixtures, plus a dedicated GitHub Action that runs it. This first PR wires up the framework and ships full **Dockerfile** and **Kubernetes manifest** coverage; Terraform / CloudFormation / Bicep fixtures land in follow-up PRs.

## How it works

- `content/iac_variants_test.go` is behind a `//go:build iac_variants` tag, so it **never runs in the default `go test ./...`** (the main "Code Test" job is unchanged). It downloads extra providers and runs many provider-backed scans, so it is isolated by design.
- Fixtures live under `content/iac-variant-testdata/<policy>/<check-uid>/{pass,fail}/<scenario>/`. Each scenario is scanned in isolation; `pass` scenarios must score 100, `fail` scenarios must score < 100. A check that doesn't run against its fixture is a fixture bug (test fails).
- One runnable test per IaC type: `TestTerraformVariants`, `TestCloudFormationVariants`, `TestBicepVariants`, `TestDockerfileVariants`, `TestKubernetesManifestVariants`.
- Dockerfile fixtures scan via the `docker-file` connection (os provider); Kubernetes fixtures scan a manifest, and the check outcome is resolved across every discovered resource asset.

## Running locally

```
make test/go/content-iac                 # all types
make test/go/content-iac/dockerfile      # one type
make test/go/content-iac/kubernetes
```

## CI

`.github/workflows/content-iac-tests.yaml` runs a matrix job per IaC type on `content/**` changes (and manual dispatch), separate from the main Go test job. Terraform/CloudFormation/Bicep jobs skip until their fixtures land.

## Coverage in this PR

- **Dockerfile:** all 41 checks across `mondoo-dockerfile-security` + `mondoo-dockerfile-best-practices`.
- **Kubernetes:** 142 workload/manifest-testable checks in `mondoo-kubernetes-security` (pod/deployment/statefulset/daemonset/replicaset/job/cronjob/namespace/RBAC). The 41 node/control-plane CIS checks (kubelet/apiserver/etcd/file-permission) are intentionally not covered here — they require a live node/cluster, not a manifest.

## Note: policy bug surfaced by the harness

`mondoo-kubernetes-security-job-capability-net-raw` queries `k8s.deployment` under a `k8s-job` filter (its siblings correctly use their own kind), so it always errors on a Job asset. Left uncovered here; a policy fix will follow separately.

## Validation

`TestBundles` and the default `go test` are unchanged (verified `go test -list` shows only `TestBundles` without the tag). The Dockerfile and Kubernetes suites pass green at `-parallel 4`.


tas50:
Thanks — addressed:

- **pull_request trigger** — the workflow now triggers on `pull_request` (branches: `main`) with the same path filters, so content changes are validated pre-merge. I mirrored this repo's own `xgrep.yaml` pattern (`push` on `main` + `pull_request` on `main` + `workflow_dispatch`) rather than `push` on all branches, so PRs are gated but feature-branch pushes don't double-run the (provider-heavy) suite.
- **matrix enum comment** — updated to include `dockerfile` and `kubernetes`.
- **cockroachdb/errors** — switched `iac_variants_test.go` to `github.com/cockroachdb/errors` per the project convention.
- **-timeout** — added `-timeout 30m` to all `test/go/content-iac` targets so the provider-heavy suites don't hit the default 10m limit.

On the xgrep run: those findings are the intentionally-insecure patterns in the `fail` fixtures (hardcoded `ENV` secrets, `curl -k`, `--security=insecure`, host-path mounts, etc.) — they're the exact misconfigurations each check is meant to catch. xgrep is `fail-on: "off"` (report-only), so it doesn't block, and the findings are expected for a security-test fixture corpus.

PR #3028, opened 2026-07-09
🐛 Guard over-broad terraform filters in GCP/AWS security policies

vjeffrey:
## Problem

Sixteen check variants in `mondoo-gcp-security` and `mondoo-aws-security` filter on a bare
`asset.platform == "terraform-hcl"` (plus the `terraform-plan` / `terraform-state` siblings)
with **no guard on the cloud provider actually in use**.

That filter is true for *every* terraform asset. Scanning Azure Terraform therefore attaches the
Mondoo GCP Security and Mondoo AWS Security policies to it. The checks then evaluate against an
empty resource set and produce false findings:

| Policy | Check | Result on Azure Terraform |
|---|---|---|
| GCP | Ensure Cloud Logging has at least one sink configured (`… .length > 0`) | ✕ MEDIUM (60) |
| AWS | Ensure MFA is enabled for the "root user" account | ✕ CRITICAL (90) |

`mondoo-azure-security` was already fully guarded — it is the model, and is unchanged here.

## Fix

Guard each variant by the provider actually in use, mirroring the idiom already used in
`mondoo-aws-security` (L10827) and `mondoo-okta-security` (L284):

```yaml
# terraform-hcl
filters: asset.platform == "terraform-hcl" && terraform.providers.any(nameLabel == "aws")
# terraform-plan
filters: asset.platform == "terraform-plan" && terraform.plan.resourceChanges.contains(type == /^aws_/)
# terraform-state
filters: asset.platform == "terraform-state" && terraform.state.resources.contains(type == /^aws_/)
```

Guarding on each check's *own* resource type would make these checks vacuous — they assert either
that a resource is absent, or that a required one is present. So the guard is deliberately
provider-level rather than resource-level.

## Verification

Scanned real terraform fixtures with `cnspec scan terraform <dir> -f content/<bundle>.mql.yaml --incognito`.

**Before** — GCP policy against Azure-only HCL:
```
Passing:  ✓ Ensure user-managed service account keys are not created
Failing:  ✕ MEDIUM (60): Ensure Cloud Logging has at least one sink configured for log export
```

**After** — GCP policy against the same Azure HCL: no longer attaches
(`asset doesn't support any policies`, as this scan supplied only that one bundle).

**After** — GCP policy against GCP HCL, i.e. no coverage lost:
```
Passing:  ✓ Ensure user-managed service account keys are not created
Failing:  ✕ MEDIUM (60): Ensure Cloud Logging has at least one sink configured for log export
```
The sink check still legitimately fails when the GCP terraform declares no sink.

The same before/after holds for AWS: the CRITICAL root-MFA false positive disappears on Azure HCL,
and all three guarded AWS checks still evaluate on AWS HCL.

`cnspec policy lint` passes on both bundles with no new warnings (the two pre-existing
"standalone function in a WHERE clause" warnings are present on `main` as well).

## Note for reviewers

A naive grep over-counts the offenders. In a `filters: |` block, consecutive lines are ANDed, so
most `asset.platform == 'terraform-plan'` lines already carry their guard on the following line.
Only 16 filters are genuinely unguarded; those are the only ones touched.

### DIFF
--- a/content/mondoo-aws-security.mql.yaml
+++ b/content/mondoo-aws-security.mql.yaml
@@ -1080,7 +1080,7 @@ queries:
       eksPlan.all(_['endpoint_private_access'] == true)
       eksPlan.all(_['endpoint_public_access'] == false)
   - uid: mondoo-aws-security-eks-cluster-private-controlplane-terraform-state
-    filters: asset.platform == "terraform-state"
+    filters: asset.platform == "terraform-state" && terraform.state.resources.contains(type == /^aws_/)
     mql: |
       eksState = terraform.state.resources.where(type == "aws_eks_cluster").map(values.vpc_config).flat
       eksState.all(_['endpoint_private_access'] == true)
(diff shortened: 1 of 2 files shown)


tas50:
Let me make sure we have a broader check for issues like this with TF

tas50:
Audited the remainder. This is it

PR #3089, opened 2026-07-19
✨ cnspec vuln: --upload discovered vulnerabilities as VEX (hidden/experimental)

chris-rock:
## What

Phase 2 of native self-emit (ADR-062 WP-SE1): let `cnspec vuln` upload the vulnerabilities it discovers to Mondoo Platform as VEX findings.

```
cnspec vuln --upload <target>
```

`--upload` is **Hidden/experimental** until validated in production.

## How

Reuses the SBOM-scan capability (from xgrep, promoted to mql as `sbomscan` in #9150; wrapped as `upload.ScanSBOM` for the file path in #3088):

1. `cnspec vuln` already runs the SBOM query pack and builds a scan report.
2. With `--upload`, cnspec generates an **mql SBOM** from that same report and sends it to Mondoo Platform (`ExtendedVulnMgmt.ScanUploadedSbom`).
3. The **platform** computes the VEX and returns it; cnspec uploads it via the existing `UploadFindings`.

cnspec does no local vulnerability matching — the platform calculates everything. The existing display path (`AnalyseAsset` → printed report) is unchanged.

## Relationship to Phase 1

This is a **separate phase** from #3088 (the file-based `cnspec upload --format sbom` path). Phase 1 = upload an SBOM *file*; Phase 2 = a live `cnspec vuln` scan emits its own findings. Both reuse `upload.ScanSBOM`.

## Verification

- `go build ./apps/cnspec/...`; gofmt / golangci-lint clean
- `--upload` is hidden from `cnspec vuln --help` but parses and is wired

### DIFF
--- a/apps/cnspec/cmd/vuln.go
+++ b/apps/cnspec/cmd/vuln.go
@@ -155,8 +155,9 @@ var vulnCmdRun = func(cmd *cobra.Command, runtime *providers.Runtime, cliRes *pl
 		log.Fatal().Err(err).Msg("failed to print")
 	}
 
-	if upload, _ := cmd.Flags().GetBool("upload"); upload {
-		uploadVulnFindings(ctx, cnspecReport.ToCnqueryReport())
+	if doUpload, _ := cmd.Flags().GetBool("upload"); doUpload {
+		// Target the same space the scan resolved, honoring the user's config.
+		uploadVulnFindings(ctx, cnspecReport.ToCnqueryReport(), upstreamConf.SpaceMrn)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


chris-rock:
Superseded by #3998, which ports this onto today's `main` rather than rebasing it.

The feature is unchanged — a hidden `--upload` on `cnspec vuln` — but the implementation here no longer fits. This PR was written against the graphql `AnalyseAsset` display path, which never had VEX in hand, so `uploadVulnFindings` built a second SBOM and called `ScanUploadedSbom` again. `cnspec vuln` now scans the SBOM itself, so the port reuses the VEX already computed: one round-trip instead of two, and what gets uploaded is exactly what was printed rather than a second scan that may resolve differently.

One behavioural fix along the way: `upload.Opts{ScopeMrn: spaceMrn}` here carries no `ConfigPath`, which would re-introduce the bug where `--config` is silently ignored and the failure surfaces as a bad service-account key. The port reuses the scan's own `Opts`. (`upstreamConf` no longer exists in `vuln.go`; an empty `ScopeMrn` already falls back to the credentials' scope.)

Closing in favour of #3998.

PR #3090, opened 2026-07-19
♻️ cnspec upload --format sbom: use BulkUploadSbom (auto-enrich upstream)

chris-rock:
## What

Reworks the `cnspec upload --format sbom` path (Phase 1, from #3088) to use the **`BulkUploadSbom`** capability instead of the ephemeral scan round-trip.

Before: decode → `ScanUploadedSbom` (returns VEX) → `VexToDocuments` → `UploadFindings`.
After: decode → **`BulkUploadSbom`** — cnspec just hands the SBOM to Mondoo Platform, which **stores it and computes the vulnerabilities itself**. No VEX handling in cnspec.

This matches the intended model: *uploading an SBOM enriches automatically upstream; the client doesn't calculate VEX.*

## Contents

- **`upload.UploadSBOM`** — wraps the new mql `sbomupload` client (`Sbom.BulkUploadSbom`, from mql #9223); testable core (`uploadSBOM` + injected client) with unit tests.
- **`cnspec upload` command** — `--format cyclonedx/spdx/sbom` decodes locally then calls `UploadSBOM(create_assets=true)`. Pure converters (SARIF) are unchanged; `--dry-run` still parses locally with no network.
- Bumps `go.mondoo.com/mql/v13` to the version that includes `sbomupload`.

`upload.ScanSBOM` remains — it's still used by `cnspec vuln --upload` (#3089), which keeps the ephemeral-scan model on purpose.

## Two SBOM mechanisms, by design

| Path | Mechanism | Behavior |
|---|---|---|
| `cnspec upload --format sbom` (this PR) | `BulkUploadSbom` | upload SBOM → platform stores + auto-enriches |
| `cnspec vuln --upload` (#3089) | `ScanUploadedSbom` | live scan → platform returns VEX → cnspec uploads |

## Verification

- build, `go test ./upload/` (incl. new `uploadSBOM` tests)
- `--dry-run` parses locally, no network; real run reaches the upload
- gofmt / golangci-lint clean; go.mod diff is version bumps only


chris-rock:
Superseded by #3998, which ports this onto today's `main` rather than rebasing it.

The design carries over unchanged — `cnspec upload --format sbom` hands the decoded SBOM to `Sbom.BulkUploadSbom` and stops, letting the platform store it and compute the vulnerabilities, instead of round-tripping VEX through `ScanUploadedSbom`. `upload.UploadSBOM` and its injected-client core come across as written.

What differs:

- The `go.sum` churn and the `mql/v13` bump are dropped — the module path is no longer versioned and `sbomupload` is already in the pinned mql.
- Two tests added beyond the originals, for the two ways this wrapper could break without anyone noticing: that `create_assets=false` actually reaches the wire (forwarding a constant `true` would create assets for a caller that asked not to), and that a nil response with a nil error reads as `0` rather than panicking — which was the stated reason for using the generated getter, but was never asserted.

Closing in favour of #3998.

PR #3119, opened 2026-07-23
👷 ci: shard the Terraform IaC-variant tests across parallel runners

tas50:
## Problem

The Terraform IaC-variant suite (`TestTerraformVariants`) scans **~1,200 variant directories** — thousands of provider-backed scans — on a **single CI runner**, taking roughly **1.5 hours** of wall-clock. In-process concurrency is deliberately capped at a safe `-parallel` because higher concurrency deadlocks on provider-subprocess contention, so the way to go faster is to fan the work out across runners, not add threads.

## Change

Shard the suite across parallel runners:

- **`content/iac_variants_test.go`** — add `IAC_SHARD_INDEX`/`IAC_SHARD_TOTAL` support. `runFixtureSuite` now runs only the scenarios whose stable **FNV-1a** hash of `policyDir/uid/scenario` falls in this runner's shard (`iacShard`/`inShard`). The filter runs *before* `t.Run`, so off-shard scenarios cost nothing. Unset env (or total ≤ 1) means run everything, so local runs and the smaller suites are unaffected. Generic across suites; only Terraform sets many shards.
- **`.github/workflows/content-iac-tests.yaml`** — expand the single `terraform` matrix cell into **8 sharded runners**; `cloudformation`/`bicep`/`dockerfile`/`kubernetes` stay single-runner (already fast — 232s/72s/25s). One shared set of steps via a matrix `include`; `IAC_SHARD_*` are empty for the non-terraform suites.
- **`Makefile`** — document the shard env vars alongside `IAC_VARIANT_PARALLEL`.

## Why hash-partition

- Needs no knowledge of how many variants exist and self-balances across thousands of scenarios.
- Stable when fixtures are added — a new scenario lands in one shard rather than reshuffling everything.
- Deterministic across reruns, so a single failing shard can be reproduced locally with `IAC_SHARD_TOTAL=8 IAC_SHARD_INDEX=<n> make test/go/content-iac/terraform`.

## Verification

- `go vet -tags iac_variants ./content` clean; `gofmt` clean.
- Ran the mechanism end-to-end on the hetzner Terraform variants: full = 33 subtests; shard 0/2 = 11, shard 1/2 = 22 — **zero overlap, zero missing, union == full**.

## Expected effect

8-way fan-out turns the ~1.5h job into ~8 parallel runners of roughly `total/8` + fixed overhead (checkout + provider download ~2–3 min), landing each shard around **~10–12 minutes**.

To retune the fan-out: add/remove `terraform` matrix entries and keep every `shards:` value equal to the number of terraform entries.

### DIFF
--- a/.github/workflows/content-iac-tests.yaml
+++ b/.github/workflows/content-iac-tests.yaml
@@ -45,10 +45,13 @@ jobs:
       # number of terraform entries.
       matrix:
         include:
-          - { iac: cloudformation }
-          - { iac: bicep }
-          - { iac: dockerfile }
-          - { iac: kubernetes }
+          # shards: 1 = run everything on one runner; the harness treats total 1
+          # as "no sharding". Explicit values keep IAC_SHARD_* non-empty so
+          # Actions emits no empty-env annotations.
+          - { iac: cloudformation, shard: 0, shards: 1 }
+          - { iac: bicep, shard: 0, shards: 1 }
+          - { iac: dockerfile, shard: 0, shards: 1 }
+          - { iac: kubernetes, shard: 0, shards: 1 }
           - { iac: terraform, shard: 0, shards: 8 }
           - { iac: terraform, shard: 1, shards: 8 }
           - { iac: terraform, shard: 2, shards: 8 }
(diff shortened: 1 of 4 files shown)


tas50:
Addressed both review comments in 510934ea:

- **Silent shard-0 fallback on bad env** (`iac_variants_test.go`): `iacShard` now takes `*testing.T` and `t.Fatalf`s on a malformed or out-of-range `IAC_SHARD_INDEX`/`IAC_SHARD_TOTAL` rather than coercing to 0. Rationale: a typo'd shard would run the wrong slice and leave another shard's scenarios unrun *everywhere*, an invisible coverage gap in a green build. Verified: `IAC_SHARD_TOTAL=8x` and index `8/8` both hard-fail with a clear message; `shards=1` still runs everything.
- **Empty-env annotations** (`content-iac-tests.yaml`): non-terraform matrix entries now carry explicit `shard: 0, shards: 1`, so `IAC_SHARD_*` are never empty. `shards: 1` is the harness's "no sharding / run everything" case.

PR #3141, opened 2026-07-27
🐛 linux-security: stop pinning auditd checks to one literal audit key

syrull:
## Problem

The 16 auditd checks in **Mondoo Linux Security** each required an exact audit key — `-k scope`, `-k logins`, `-k time-change`, `-k perm_mod`, and so on.

The auditd key is only a label used to group and search audit events. The kernel emits the record either way. What actually matters is the watched path plus its permissions (watch rules), or the syscall list plus the arch/exit/auid filters (syscall rules).

Pinning the literal produces false negatives on correctly hardened hosts. Reported downstream against a RHEL 9.8 server (cnspec 13.29.1) whose rules are correct:

```
# auditctl -l | grep sudoers
-w /etc/sudoers -p wa -k sudoers
-w /etc/sudoers.d -p wa -k sudoers
```

The check failed purely because the key read `sudoers` rather than `scope`.

CIS agrees. Every generation of the CIS Linux benchmarks verifies these rules with

```
awk '/^ *-w/ && /\/etc\/sudoers/ && / +-p *wa/ && (/ key= *[!-~]* *$/ || / -k *[!-~]* *$/)'
```

`[!-~]*` accepts any run of printable characters, so CIS requires a **non-empty** key and never a specific one. The named key in the benchmark text is example output, not an assertion.

## Change

All 32 keyname constraints across the 16 checks now assert that a key is *present* instead of that it equals a literal.

| rule kind | before | after |
|---|---|---|
| watch (`-w`) | `f.keyname == "scope"` | `f.keyname != ""` |
| syscall (`-a`) | `keyname == "time-change"` | `action == "always"` + `fields.contains(key == "key" && value != "")` |

Two details worth review:

**Syscall rules use `fields`, not `keyname`.** `keyname` is populated only from `-k` (`providers/os/resources/auditd.go:349`). The provider deliberately normalizes `-k` into `fields` as `key=<name>` — its own comment says *"-k is shorthand for -F key=; normalize into fields so queries don't need to check both representations."* Reading the field therefore also matches rules written as `-F key=<name>`, which `keyname` alone silently misses.

**`action == "always"` is new and load-bearing.** No syscall selector constrained `action`, so `-a never,exit` suppression rules were already eligible to satisfy these checks; the literal key was only an accidental guard, since nobody tags a suppression rule `-k time-change`. Relaxing the key would have widened that hole, so the guard is now explicit. Verified below.

Untouched on purpose: every `uid`, `title`, `impact`, `filters`, `variants` and `compliance/*` tag, so framework mappings and customer exceptions are unaffected. Remediation snippets keep recommending the conventional key — that is still good practice.

Docs follow the code. Every `audit:` procedure that told the user to `grep -- '-k <key>'` now verifies what the check actually asserts, and no `desc:` still claims records carry a specific identifier. The auditd example in `skills/mql/mql-reference.md` taught the same anti-pattern and is updated with a note.

## Verification

`cnspec policy lint` clean; `go test ./content -run TestBundles` passes.

Behaviour was measured, not assumed — five mock RHEL 9.8 roots scanned with `cnspec run fs`:

| query | reported host<sup>1</sup> | CIS host<sup>2</sup> | no key at all<sup>3</sup> | no rules | `-a never`<sup>4</sup> |
|---|---|---|---|---|---|
| sudoers **before** | ❌ FAIL | ✅ PASS | ❌ FAIL | ❌ FAIL | – |
| sudoers **after** | ✅ **PASS** | ✅ PASS | ❌ FAIL | ❌ FAIL | – |
| time-change **before** | ❌ FAIL | – | – | ❌ FAIL | ⚠️ PASS |
| time-change **after** | ✅ **PASS** | – | – | ❌ FAIL | ✅ FAIL |
| time-change, key fix only | ✅ PASS | – | – | ❌ FAIL | ⚠️ PASS |

<sup>1</sup> `-k sudoers` watch rules + a syscall rule spelled `-F key=`, exactly as reported.
<sup>2</sup> conventional `-k scope`.
<sup>3</sup> `-w /etc/sudoers -p wa` with no key — **still fails**; the requirement was relaxed, not dropped.
<sup>4</sup> `-a never,exit … -k time-change`; the last row is the same fix without the `action` guard, showing the guard is what closes the hole.

Reported downstream: mondoo-community/workspace-telekom#203

### DIFF
--- a/content/mondoo-linux-security.mql.yaml
+++ b/content/mondoo-linux-security.mql.yaml
@@ -6804,7 +6804,7 @@ queries:
       paths.all(p:
         auditd.rules.files.contains( f:
           f.path == p && f.permissions == "wa"
-          && f.keyname != ""
+          && f.keyname != empty
         )
       )
     docs:
@@ -7125,7 +7125,7 @@ queries:
       paths.all(p:
         auditd.rules.files.contains( f:
           f.path == p && f.permissions == "wa"
-          && f.keyname != ""
+          && f.keyname != empty
         )
       )
   - uid: mondoo-linux-security-login-and-logout-events-are-collected-rhel
@@ -7135,7 +7135,7 @@ queries:
       paths.all(p:
         auditd.rules.files.contains( f:
           f.path == p && f.permissions == "wa"
-          && f.keyname != ""
+          && f.keyname != empty
         )
       )
   - uid: mondoo-linux-security-login-and-logout-events-are-collected-other
@@ -7145,7 +7145,7 @@ queries:
       paths.all(p:
         auditd.rules.files.contains( f:
           f.path == p && f.permissions == "wa"
-          && f.keyname != ""
+          && f.keyname != empty
         )
       )
   - uid: mondoo-linux-security-session-initiation-information-is-collected
@@ -7169,14 +7169,14 @@ queries:
       // Check utmp is watched with a non-empty audit key
       auditd.rules.files.contains( f:
         f.path == "/var/run/utmp" && f.permissions == "wa"
-        && f.keyname != ""
+        && f.keyname != empty
       )
       // Check wtmp and btmp are watched with a non-empty audit key
       sessionPaths = ["/var/log/wtmp", "/var/log/btmp"]
       sessionPaths.all(p:
         auditd.rules.files.contains( f:
           f.path == p && f.permissions == "wa"
-          && f.keyname != ""
+          && f.keyname != empty
         )
       )
     docs:
@@ -7322,7 +7322,7 @@ queries:
         syscalls.in(syscalls_64bit)
         && arch == "b64"
         && action == "always"
-        && fields.contains(key == "key" && value != "")
+        && fields.contains(key == "key" && value != empty)
       ).map(syscalls).flat
       desiredRules64.containsAll(syscalls_64bit)
 
@@ -7332,7 +7332,7 @@ queries:
         syscalls.in(syscalls_32bit)
         && arch == "b32"
         && action == "always"
-        && fields.contains(key == "key" && value != "")
+        && fields.contains(key == "key" && value != empty)
       ).map(syscalls).flat
       desiredRules32.containsAll(syscalls_32bit)
 
@@ -7340,7 +7340,7 @@ queries:
       auditd.rules.files.contains(
         path == "/etc/localtime"
         && permissions == "wa"
-        && keyname != ""
+        && keyname != empty
       )
     docs:
       desc: |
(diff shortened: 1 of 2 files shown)


syrull:
Adopted `!= empty` in `93075719` — but the stated rationale was wrong in a way worth recording, because one of the three comments found a real bug and the other two did not.

**`keyname` is never null.** It is declared non-nullable (`keyname string`, `providers/os/resources/os.lr:1379` and `:1410`) and the parser explicitly sets it to `""` when the rule carries no key (`auditd.go:375-377` for watch rules, `:466-468` for syscall rules). Measured on a mock RHEL 9.8 root whose only rules are `-w /etc/sudoers -p wa` and `-w /etc/sudoers.d -p wa`:

```
> auditd.rules.files { path keyname }
  0: { keyname: ""  path: "/etc/sudoers" }
  1: { keyname: ""  path: "/etc/sudoers.d" }

> auditd.rules.files.all(keyname == null)   → [failed]
> auditd.rules.files.all(keyname == "")     → [ok] value: true
```

That fixture was already in the PR as the `no key at all` column, and it **failed** the check under `keyname != ""`, which is the behaviour the review asked for. So keyless rules were never passing.

**The `fields` comment did find something, though not for the reason given.** A malformed `-F key` with no operator parses to `{key: "key"}` with no `value` key at all (`auditd.go:392-394`), and `null != ""` is true, so that rule satisfied the predicate:

```
-a always,exit -F arch=b64 -S adjtimex,settimeofday,clock_settime -F key
  fields: [ {key:"arch", op:"=", value:"b64"}, {key:"key"} ]     ← no value

  value != ""     → PASS   ← false positive
  value != empty  → FAIL   ← correct
```

`-F key=` with an explicit empty value was already rejected by both spellings. So `!= empty` is a strict improvement on exactly one malformed-input path, and identical everywhere else. Applied uniformly to both spellings and to the reference doc, since one predicate is easier to reason about than two.

Full matrix re-measured after the change (`cnspec run fs` against five mock RHEL 9.8 roots):

| query | reported host | CIS host | no key | bare `-F key` | no rules | `-a never` |
|---|---|---|---|---|---|---|
| sudoers **before** | ❌ FAIL | ✅ PASS | ❌ FAIL | – | ❌ FAIL | – |
| sudoers **after** | ✅ **PASS** | ✅ PASS | ❌ FAIL | – | ❌ FAIL | – |
| time-change **before** | ❌ FAIL | – | – | ❌ FAIL | ❌ FAIL | ⚠️ PASS |
| time-change **after** | ✅ **PASS** | – | – | ❌ FAIL | ❌ FAIL | ✅ FAIL |
| time-change, no `action` guard | ✅ PASS | – | – | ❌ FAIL | ❌ FAIL | ⚠️ PASS |

`cnspec policy lint` clean, `go test ./content -run TestBundles` passes.

PR #3148, opened 2026-07-29
feat(upload): report why a scan-result upload failed

vjeffrey:
When a scan-result upload fails, cnspec currently tells the platform almost nothing — so an upload that never completes is indistinguishable from one that was never attempted. This adds structured reporting for every upload failure a running client can observe.

## The gap

In `uploadScanDataStore`, only a **non-2xx PUT status** was reported. Everything else returned a bare error:

| failure | before | after |
|---|---|---|
| `GetUploadURL` fails | silent | `url_request_failed` |
| **PUT transport error** (timeout, reset, DNS, TLS, ctx cancel) | **silent** | classified `failureKind` |
| PUT non-2xx | reported | reported + structured fields |
| **`ReportUploadCompleted` fails** | **silent** | `report_rpc_failed` |

The silent cases turn out to be the common ones, and they are exactly the cases the server cannot diagnose on its own: from the receiving end, a client that stalled mid-upload and a client that never connected look identical.

## What this adds

**`bytes_sent`, via a counting `io.Reader`** — the useful field, because it separates causes the server cannot:

- `0` → connection never established (refused, DNS, TLS, blocked egress)
- `0 < n < total` → **stalled mid-body**, e.g. a proxy or load balancer request timeout on a multi-MB PUT
- `== total`, no response → the body landed and the response was dropped; **the object may exist and be recoverable**

`UploadFile` now returns a `Result` reporting `BytesSent` and `Duration` on **both** the success and error paths. The counter is `atomic.Int64` because `http.Client` may read the request body on a different goroutine than the one returning from `Do`.

**A closed `failureKind` taxonomy** (`upload/failurekind.go`), classified with `errors.Is`/`errors.As` against the error tree rather than string matching, so it survives stdlib message changes. An unrecognised error maps to `other` **and keeps its message** — a failure is never dropped because the classifier doesn't recognise it yet.

**Success baseline** — `cnspec.scan.upload_duration` and `cnspec.scan.upload_throughput_bps` on the existing `ScanStatistics` payload (already `repeated Metric`, so **no proto change**). Without a baseline you can't tell whether failing uploads are the tail of the normal latency distribution or a distinct mode.

Reporting goes through the **existing** `health.SendError` RPC with a `reportKind=upload_failure` discriminator tag — no new RPC, no proto change. It's best-effort by construction: `health.ReportError` swallows its own errors and returns nothing, so reporting can never fail a scan.

A matching platform-side change records these reports with their structured fields; until it ships, the reports degrade gracefully to the existing generic error-report path.

## Scope

This makes upload failures **diagnosable**. It does **not** change what happens on failure — a failed bucket upload still gives up, and the scan result is lost. Adding a retry or a fallback to the `StoreResults` path is deliberately left for separate work.

Also noticed and deliberately not changed here: `newHTTPClient()` returns a bare `&http.Client{}` with **no timeout** when no proxy is configured, so the scan-db PUT is unbounded — unlike the findings upload path, which sets an explicit 2-minute timeout. Worth fixing, but it changes behaviour rather than observability.

## Testing

`go build ./...`, `go vet`, and `go test -race` clean across `./upload/`, `./internal/datalakes/sqlite/`, `./policy/scanstats/`.

- `ClassifyFailure`: one case per taxonomy entry, plus a wrapped error and an unrecognised error → `other`
- counting reader: exact count on success; partial count when the server hijacks the connection and closes mid-body
- tag builder: `httpStatus` omitted when there was no response; `uploadSessionId` omitted for `url_request_failed`, since no session exists yet

### DIFF
--- a/internal/datalakes/sqlite/sqlite.go
+++ b/internal/datalakes/sqlite/sqlite.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/rs/zerolog/log"
@@ -170,17 +171,63 @@ func fileSizeBytes(path string) int64 {
 	return info.Size()
 }
 
+// uploadFailureReportKind is the value sent on the "reportKind" tag to mark a
+// report as an upload failure. The platform dispatches on it to record the
+// failure with its structured fields, so the literal is part of the wire
+// contract and must stay in sync with the server side.
+const uploadFailureReportKind = "upload_failure"
+
+// uploadFailureTags builds the tag set for an upload-failure report. Numeric
+// values are rendered as strings because health.SendError tags are
+// map[string]string; httpStatus and uploadSessionId are omitted rather than
+// sent as zero/empty so the server can tell "absent" from "actually zero".
+func uploadFailureTags(sessionID, assetMrn string, kind upload.FailureKind, httpStatus int, res upload.Result, bytesTotal int64) map[string]string {
+	tags := map[string]string{
+		"reportKind":  uploadFailureReportKind,
+		"failureKind": string(kind),
+		"bytesSent":   strconv.FormatInt(res.BytesSent, 10),
+		"bytesTotal":  strconv.FormatInt(bytesTotal, 10),
+		"durationMs":  strconv.FormatInt(res.Duration.Milliseconds(), 10),
+	}
+	if sessionID != "" {
+		tags["uploadSessionId"] = sessionID
+	}
+	if assetMrn != "" {
+		tags["assetMrn"] = assetMrn
+	}
+	if httpStatus != 0 {
+		tags["httpStatus"] = strconv.Itoa(httpStatus)
+	}
+	return tags
+}
+
+// reportUploadFailure sends one upload-failure report. Best-effort by
+// construction: health.ReportError swallows its own errors and returns
+// nothing, so a reporting problem can never fail the scan.
+func reportUploadFailure(msg string, tags map[string]string) {
+	health.ReportError("cnspec", cnspec.Version, cnspec.Build, msg, health.WithTags(tags))
+}
+
 func uploadScanDataStore(ctx context.Context, services *policy.Services, assetMrn string, scanDataPath string, stats *scanstats.Collector) error {
+	bytesTotal := fileSizeBytes(scanDataPath)
+
 	urlResp, err := services.GetUploadURL(ctx, &policy.GetUploadURLReq{
 		Kind:     policy.UploadURLKind_UPLOAD_URL_KIND_SCAN_DATABASE_V0,
 		ScopeMrn: assetMrn,
 	})
 	if err != nil {
+		// No upload session exists yet, so the platform has nothing to
+		// correlate this with — but a client that cannot even obtain a URL is
+		// still worth surfacing.
+		reportUploadFailure("upload failed: could not get upload URL: "+err.Error(),
+			uploadFailureTags("", assetMrn, upload.FailureURLRequest, 0, upload.Result{}, bytesTotal))
 		return err
 	}
 
 	uploadUrl := urlResp.UploadUrl
 	if uploadUrl == nil {
+		reportUploadFailure("upload failed: no upload URL for scan data store",
+			uploadFailureTags(urlResp.UploadSessionId, assetMrn, upload.FailureURLRequest, 0, upload.Result{}, bytesTotal))
 		return errors.New("no upload URL for scan data store")
 	}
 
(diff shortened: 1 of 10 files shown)


vjeffrey:
Thanks — all five addressed in ec049fc8. Four taken as suggested, one declined with reasoning.

**Note on the two marked "Resolved":** `upload.go:75` (ContentLength) and `failurekind.go:69` (ECONNREFUSED/EPIPE) were auto-resolved when I force-pushed a history rewrite, but that rewrite only touched comments — the code was unchanged and both issues were still live. Genuinely fixed now.

### 🟡 Raw `err.Error()` leaking presigned URLs — `sqlite.go:223`

Right, and worse than "could contain full URLs": `net/http` transport errors arrive wrapped in `*url.Error`, which embeds the request URL, and for a signed upload that query string carries `X-Goog-Signature` — a time-limited **write credential** for the object. That was being copied into health reporting and on into logs.

New `upload.RedactError`: unwraps `*url.Error` and keeps only op + underlying cause (dropping the URL entirely), regex-strips the query from any URL an intermediate layer formatted in, and bounds the result to 512 bytes — the same cap already applied to `responseBody`. Applied at all three sites.

Nothing is lost diagnostically — the upload session id and asset MRN already ride along as tags and identify the request far better than a single-use URL. Five tests, including one built from a realistic signed GCS URL asserting the signature cannot survive.

### 🔵 `ECONNREFUSED`/`EPIPE` classified as `connection_reset` — `failurekind.go:69`

Agreed, and I took the split rather than the rename, because the distinction turns out to be load-bearing for the investigation this PR came from.

`connection_refused` · `connection_reset` · `broken_pipe` are now separate kinds. There's an open question on INC-2026-07-29 about hosts that can reach `api.<instance>` but apparently not `ingest.<instance>` — the v2 path introduced a new egress destination. Those hosts present as **refused** or **dns**; folding them into "reset" would have buried the exact signal being hunted. (server#18058 fixes the other failure mode and explicitly leaves this one to the client side.)

Server-side record doc updated to match, since the taxonomy is a wire contract.

### 🔵 `ContentLength` fragility — `upload.go:75`

Comment added. One correction to the rationale, though: the assignment was **always** required, not only after `countingReader`. `net/http` infers `ContentLength` for `*bytes.Buffer`, `*bytes.Reader` and `*strings.Reader` only — `*os.File` isn't one of them either, so the original code needed the explicit set too. `countingReader` does additionally mask the concrete type. The comment says that, and says not to drop or move the line.

### 🔵 `Collector.add()` nil receiver — `sqlite.go:264`

Fixed in `add()` rather than at the call site, since `ToProto` already had the same guard and being nil-tolerant in one and panicking in the other is the actual inconsistency. Test asserts all four `Add*` helpers are safe on a nil collector.

### 🔵 nil → `FailureOther` — `failurekind.go:49`

**Documented, sentinel declined.** `FailureKind` is a closed wire vocabulary that downstream queries and dashboards match on. Adding `FailureNone` would oblige every consumer to handle a value that should never legitimately be emitted — nil only reaches `ClassifyFailure` via a caller bug, and callers only classify on a failure. `FailureOther` with an empty error message already makes such a bug conspicuous in the data. Godoc on both the constant and the function now states the mapping is intentional.

`go build ./...`, `go vet`, `go test -race` green across `./upload/`, `./internal/datalakes/sqlite/`, `./policy/scanstats/`.

vjeffrey:
## Re: code scanning — unanchored hostname regexp (`redact.go:21`, alert #1091)

**False positive, and I'd argue against the suggested remediation** — anchoring this pattern would reintroduce the exact credential leak the function exists to prevent.

The rule targets regexes that **validate** a URL as part of a security decision, where an unanchored match lets an attacker embed a trusted host inside a longer string and pass the check. `signedURLQueryRe` makes no security decision. It is a redaction sweep over free-form error text, and the URL it has to find is **always mid-string** — `net/http` renders a failed request as:

```
Put "https://host/obj?X-Goog-Signature=…": dial tcp 10.0.0.1:443: connect: connection refused
```

Verified rather than asserted — same input, both variants:

```
unanchored: Put "https://…/s1?<redacted>": dial tcp …            ← signature removed
anchored:   Put "https://…/s1?X-Goog-Signature=DEADBEEFCAFE": …  ← signature intact
```

With `^`/`$` the regex matches nothing, redaction silently becomes a no-op, and the signing credential ships to the platform and into logs.

There's also no untrusted input in the security-decision sense the rule cares about: the input is an error string produced locally by `net/http` for a URL **we** requested from our own API, and the output is only ever a telemetry message — nothing downstream branches on it.

Pushed in d8aaba64 (docs + test only, no behaviour change):
- comment on the regex stating it must stay unanchored, and why
- `TestRedactError_RedactsURLMidMessage`, built from the exact `net/http` error shape, whose failure message names anchoring as the likely cause

Note the regex is only the second layer anyway — the primary defence is unwrapping `*url.Error` and dropping the URL entirely.

Suggest dismissing alert #1091 as a false positive. Happy to do that, or leave it for someone with security-tab ownership to action.

PR #3154, opened 2026-07-30
🐛 scan: stop a failed upstream sync from segfaulting the scanner

preslavgerchev:
## What happened

A production Azure scan crashed with a nil pointer dereference:

```
panic: runtime error: invalid memory address or nil pointer dereference
go.mondoo.com/mql/v13/providers.(*Runtime).EnsureProvidersConnected(...)
	providers/runtime.go:326
go.mondoo.com/cnspec/v13/policy/scan.(*scanDispatcher).scanSingleAsset(...)
	policy/scan/scan_pipeline.go:231
go.mondoo.com/cnspec/v13/policy/scan.(*scanDispatcher).Submit.func1()
	policy/scan/scan_pipeline.go:216
```

The receiver itself was nil, not one of its fields. Working backwards from the logs:

1. `SynchronizeAssets` returned `Unauthenticated`. All three attempts failed, so `syncBatchWithUpstream` returned the error.
2. `syncBatcher.Flush` propagated it up through `scanSubtree`, and `distributeJob` returned.
3. Nothing drains the dispatcher on that path — `dispatcher.Wait()` only runs on the success path and at branch boundaries — so the deferred `explorer.Shutdown()` closed every asset and set each `Runtime` to nil while scan goroutines were still alive.
4. Assets submitted by earlier, successful batches were parked waiting for a worker slot (`connSem` is 50, parallelism is much lower, so queueing is the normal state). ~6s later a slot freed, one resumed into `scanSingleAsset`, and dereferenced its now-nil runtime.

## Changes

- **Drain in-flight scans before returning.** A cancellable `scanRunCtx` bounds every scan goroutine, and a defer registered after the `explorer.Shutdown` one (defers are LIFO, so it runs first) cancels it and waits for the dispatcher. `Submit` rechecks `ctx.Err()` after acquiring a slot, since `select` picks at random when both cases are ready. This is the fix for the crash.
- **Guard `scanSingleAsset` against a closed asset.** Defence in depth for the exact deref, and it also covers a second race: `AssetExplorer.dedup` evicts already-submitted assets whose platform IDs turn out to be a subset of a newly connected one, clearing their runtime the same way.
- **Make the per-asset panic handler actually recover.** It delegated to `health.ReportPanic`, whose `recover()` sat one frame too deep to catch anything — so panics killed the process *and* no report was ever sent. Panics are now recovered, reported with the same asset tags, recorded as a scan error, and contained to the one asset. This also makes `sendErrorToMondooPlatform` dead code (mql's `sendPanic` does the same work), so it's removed.

Retry behaviour for `SynchronizeAssets` is unchanged — every failure still retries, pending a proper assessment of which upstream codes are genuinely non-retryable.

## Testing

New `policy/scan/scan_pipeline_test.go` covers the nil-runtime path that segfaulted, and that the panic handler swallows the panic and records the asset as failed.

`go test -race ./policy/...` passes and `golangci-lint run ./policy/scan/...` is clean.

## Note

This fixes the crash, not the trigger. The scan still fails on `Unauthenticated` from `SynchronizeAssets` — now with a clean error instead of a SIGSEGV — so that credential rejection still needs looking at on the platform side.

### DIFF
--- a/policy/scan/local_scanner.go
+++ b/policy/scan/local_scanner.go
@@ -26,7 +26,6 @@ import (
 	"go.mondoo.com/cnspec/v13/policy"
 	"go.mondoo.com/cnspec/v13/policy/executor"
 	"go.mondoo.com/mql/v13"
-	"go.mondoo.com/mql/v13/cli/config"
 	"go.mondoo.com/mql/v13/cli/execruntime"
 	"go.mondoo.com/mql/v13/discovery"
 	"go.mondoo.com/mql/v13/llx"
@@ -41,7 +40,6 @@ import (
 	"go.mondoo.com/mql/v13/providers-sdk/v1/upstream/gql"
 	"go.mondoo.com/mql/v13/providers-sdk/v1/upstream/health"
 	"go.mondoo.com/mql/v13/utils/multierr"
-	ranger "go.mondoo.com/ranger-rpc"
 	"go.mondoo.com/ranger-rpc/codes"
 	"go.mondoo.com/ranger-rpc/status"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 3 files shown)


policy/scan/local_scanner.go
```diff
@@ -728,6 +741,21 @@ func (sc *scanContext) scanSubtree(ctx context.Context, node *discovery.TrackedA
 	return nil
 }
 
+// isPermanentUpstreamError reports whether an upstream RPC error will fail the
+// same way on every attempt — expired or rejected credentials, missing
+// permissions, or a request the server refuses outright — as opposed to a
+// transient network, timeout or backend fault that a retry can recover from.
+func isPermanentUpstreamError(err error) bool {
```
preslavgerchev: drop this, id rather us retry unconditionally for now. we need to properly assess  whats a permanetn err.

PR #3177, opened 2026-08-08
🐛 syslimits: don't set GOMEMLIMIT on very small containers

chris-rock:
**Stacked on #3176** (2 of 3).

## Problem
`GOMEMLIMIT` is a *soft* limit. When the live heap approaches it, the Go GC runs back-to-back trying to stay under it — a "GC death spiral" that burns CPU instead of doing work. On very small containers, setting the limit near the live heap is more likely to thrash than to help.

## Fix
Add a 64 MiB floor (`minMemoryLimitBytes`) below which `applyMemory` leaves the GC on its default behavior. Operators who still want a limit on a tiny container can set `GOMEMLIMIT` explicitly.

## Test
`TestApplyMemory` gains below-floor (skipped) and at-floor (applied) cases.


chris-rock:
Closing: the 64 MiB `GOMEMLIMIT` floor from this PR has been folded down into #3175, so this branch is now empty (`git rebase` reports `patch contents already upstream`).

The reason for moving it: the floor mitigates a risk that #3175 introduces. Stacked two PRs above it, any partial merge of the stack would have shipped `GOMEMLIMIT` without its safety floor — measured, a 128 MiB container gets `GOMEMLIMIT` 115 MiB, and if the live heap approaches that the GC runs back-to-back chasing a target it can't reach. The mitigation now ships in the same PR as the risk.

Verified against real cgroup v2 limits after the move: 48 MiB container → no `GOMEMLIMIT` set (floor applies); 128 MiB → 115 MiB; 512 MiB → 461 MiB; 2 GiB → 1843 MiB.

PR #3178, opened 2026-08-08
📚 docs: document resource-limit behavior and env vars

chris-rock:
**Stacked on #3177** (3 of 3).

Documents the resource-limit behavior introduced by this series so it's discoverable:

- `docs/resource-limits.mdx` — how cnspec constrains itself to cgroup CPU/memory limits (CPU via the Go runtime, `GOMEMLIMIT` with headroom and the small-container floor, memory-scaled provider concurrency), plus the background failure-reporting behavior, with a table of the new/relevant environment variables:
  - `MONDOO_DISABLE_AUTO_RESOURCE_LIMITS`
  - `MONDOO_MEMORY_LIMIT_HEADROOM`
  - `MONDOO_MAX_PROVIDER_CONNECTIONS`
  - `GOMEMLIMIT` / `GOMAXPROCS`
- A unit test for `memoryHeadroom()` env parsing (valid override, full-at-1.0, invalid values falling back to the default) — filling the one remaining test gap noted in review.


tas50:
@chris-rock we probably want to cleanup the old docs in this dir first

PR #3201, opened 2026-08-09
AWS security: adopt new MQL reachability, inline-policy, and exposure fields

tas50:
Improves four checks in `mondoo-aws-security` using AWS provider fields added to mql in the last two weeks.

- **secgroup internet-facing admin/database ports** — use NACL-aware `aws.ec2.networkinterface.effectiveIngress` (`.reachable`), so a port blocked by the subnet network ACL is no longer reported as internet-exposed. Rule is attributed to this security group to avoid a sibling SG on the same ENI causing a false positive.
- **eks-cluster-restrict-public-access** — also assert `aws.eks.cluster.exposure.internetReachable`, catching security-group-level exposure the CIDR allow-list alone misses.
- **ecs-container-no-public-ip** — evaluate `aws.ecs.clusters { services.exposure }` (public IP combined with task security groups) instead of a bare public IP; **retitled/rescoped** to internet reachability, with a null-guard for host/bridge networking.
- **iam-no-wildcards-policies** — extend coverage to inline policies via `inlinePolicyDetails.hasWildcardAllow()` on roles, users, and groups; **retitled/rescoped**.

Titles and descriptions were updated where the scope changed. Terraform/CloudFormation variants are untouched.

---
> **Heads-up on CI:** these checks reference AWS provider fields that landed in mql only recently. cnspec content-lint resolves against *released* provider schemas, so this PR's content-lint will stay red until the `aws` provider release ships these fields.

https://claude.ai/code/session_014b3xEcxLyvaRwmFzoazLuk

---
_Generated by [Claude Code](https://claude.ai/code/session_014b3xEcxLyvaRwmFzoazLuk)_

### DIFF
--- a/content/mondoo-aws-security.mql.yaml
+++ b/content/mondoo-aws-security.mql.yaml
@@ -284,6 +284,7 @@ policies:
           - uid: mondoo-aws-security-ecs-task-definition-no-plaintext-secrets
           - uid: mondoo-aws-security-ecs-init-process-enabled
           - uid: mondoo-aws-security-ecs-container-no-public-ip
+          - uid: mondoo-aws-security-ecs-service-not-internet-reachable
           - uid: mondoo-aws-security-ecs-container-no-latest-tag
           - uid: mondoo-aws-security-ecs-container-image-from-ecr
           - uid: mondoo-aws-security-ecs-container-supported-fargate-version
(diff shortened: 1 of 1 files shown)


content/mondoo-aws-security.mql.yaml
```diff
@@ -55910,11 +55915,11 @@ queries:
   # representation. Terraform remediation is still provided because the fix lives in
   # the ECS service network configuration.
   - uid: mondoo-aws-security-ecs-container-no-public-ip
-    title: Ensure ECS containers are not assigned public IP addresses
+    title: Ensure ECS service tasks are not reachable from the internet
     impact: 80
     filters: asset.platform == "aws"
     mql: |
-      aws.ecs.containers.all(publicIp == empty)
+      aws.ecs.clusters.all(services.all(exposure == null || exposure.internetReachable == false))
```
tas50, line 56148: Create an all new check for this instead of using the old uid

tas50: Split into a new check. `mondoo-aws-security-ecs-container-no-public-ip` is restored byte-for-byte to its original title, MQL (`aws.ecs.containers.all(publicIp == empty)`), and docs, and the reachability rule now lives under a new uid, `mondoo-aws-security-ecs-service-not-internet-reachable`, with its own docs, remediation, and compliance tags.

Also addressed the bot's note on the null guard with an inline comment explaining that `exposure` is null for host/bridge networking services, where the container instance rather than the task decides reachability.

fe9f8ff

PR #3233, opened 2026-08-11
🐛 embed a VERSIONINFO resource in Windows builds

### LINKED ISSUE #3228: Windows binaries ship without a VERSIONINFO resource
## Summary

The Windows binaries we ship do not embed a `VERSIONINFO` resource. `cnspec.exe` therefore has empty `CompanyName`, `ProductName`, `FileDescription`, `FileVersion`, and `OriginalFilename` fields in Explorer's Details tab and to anything else that reads PE file metadata.

## Details

Go does not emit a version resource by default — it has to be supplied as a `.syso` object linked into the binary. There is no `.syso`, and no `go-winres` / `goversioninfo` step anywhere in the build:

- `.goreleaser.yml` — `builds:` → `id: windows`
- `.github/.goreleaser-unstable.yml` — same
- `Makefile` — `cnspec/build/windows`

Authenticode signing is already in place (jsign via Azure Trusted Signing, post-build hook), so the publisher identity is present. Only the descriptive file metadata is missing.

## Why it matters

A stripped (`-s -w`), statically linked, `CGO_ENABLED=0` Go binary carrying no file metadata is a well-documented contributor to heuristic and ML-based AV/EDR misclassification. Endpoint products weight PE metadata completeness as a feature, and "no company, no product, no description" is over-represented in their malicious training data.

This is amplified for `cnspec serve`, which legitimately exhibits several behaviours those classifiers score on:

- runs as a SYSTEM service launched by `services.exe`
- fetches and executes provider plugin binaries at runtime (`updateProviders()` in `apps/cnspec/cmd/serve.go`)
- shells out to PowerShell for host inventory and configuration checks

Every release also lands as a previously unseen file hash with no cloud prevalence, so there is nothing to offset the missing metadata. Populating the version resource is a cheap, standard mitigation that gives these products something to attribute the file to.

Secondary benefit: `cnspec.exe` currently shows blank fields in Task Manager, Explorer, and software-inventory tooling, which is a poor look for a security product.

## Proposed fix

Add a [`go-winres`](https://github.com/tc-hib/go-winres) step that generates a `.syso` before the Windows build, and wire it into both goreleaser configs and the `Makefile` target. Fields to populate:

| Field | Value |
|---|---|
| `CompanyName` | Mondoo, Inc. |
| `ProductName` | cnspec |
| `FileDescription` | Cloud-Native Security and Policy Framework |
| `OriginalFilename` | cnspec.exe |
| `LegalCopyright` | Copyright Mondoo, Inc. |
| `FileVersion` / `ProductVersion` | injected from the release version |

The manifest should keep the default `asInvoker` execution level — `cnspec` must not request elevation.

Two things to get right:

1. The `.syso` must exist in the `main` package directory (`apps/cnspec/`) at compile time and be cleaned up afterwards, so it does not leak into non-Windows builds or the working tree.
2. Signing must remain the *last* step. The version resource has to be linked in before jsign runs, or the signature will not cover it.

M-gre:
Closes #3228.

`cnspec.exe` shipped with no PE file metadata — empty `CompanyName`, `ProductName`, `FileDescription`, `FileVersion` and `OriginalFilename`. Go does not emit a version resource on its own; it has to be linked in as a `.syso`, and nothing in the build supplied one.

## Changes

- **`apps/cnspec/winres/winres.json`** — new [go-winres](https://github.com/tc-hib/go-winres) config holding the version resource and application manifest. No icon is included; there is no icon asset in the repo and that can follow separately.
- **`.goreleaser.yml`, `.github/.goreleaser-unstable.yml`** — a global `before` hook generates `apps/cnspec/rsrc_windows_{amd64,arm64}.syso`, with the version fed from `{{ .Version }}`. Pinned to `go-winres@v0.3.3` and invoked via `go run`, so no new entry in `go.mod` and no extra CI install step. `.github/.goreleaser-edge.yml` is Linux-only and needs no change.
- **`Makefile`** — new `cnspec/winres` target, made a prerequisite of `cnspec/build/windows`. The tag is stripped of its leading `v` for the numeric PE fields.
- **`.gitignore`** — ignores the generated `.syso` files.

## The one non-obvious bit

The windows build target had to move from `./apps/cnspec/cnspec.go` to `./apps/cnspec`. Go only links `rsrc_windows_*.syso` when building a **package directory** — passing an explicit `.go` file makes it ignore the resource with no warning. I hit exactly this while testing: the first binary I built had the `.syso` sitting right next to it and still came out with no resource directory at all.

Both goreleaser configs and the Makefile carry a comment on this so it does not get "tidied" back.

The `linux` and `macos` builds still use the file path. They are unaffected either way, and I kept the diff narrow rather than churn all three.

## Verification

Built `windows/amd64` and read the resource back out of the linked binary with `go-winres extract`:

```
"file_version":     "13.32.1.0"
"product_version":  "13.32.1.0"
"CompanyName":      "Mondoo, Inc."
"ProductName":      "cnspec"
"FileDescription":  "Cloud-Native Security and Policy Framework"
"FileVersion":      "13.32.1"
"OriginalFilename": "cnspec.exe"
"LegalCopyright":   "Copyright Mondoo, Inc. All rights reserved."
```

Extracted manifest confirms the execution level is unchanged:

```xml
<requestedExecutionLevel level="asInvoker" uiAccess="false"/>
```

Also confirmed:

- `make cnspec/build/windows` produces the metadata end to end.
- `darwin/arm64` and `linux/amd64` builds are unaffected with the `.syso` files present — the `_windows_<arch>` filename suffix keeps the Go toolchain from linking them.
- `go.mod` / `go.sum` are untouched by the `go run` invocation.
- Both goreleaser configs parse; `goreleaser check` runs on them in CI via the existing GoReleaser Check workflow.

Signing is unaffected. It remains a post-build hook and still runs last, so the Authenticode signature covers the new resource.

## Why bother

A stripped (`-s -w`), statically linked, `CGO_ENABLED=0` Go binary with no file metadata is a well-documented contributor to heuristic and ML-based AV/EDR misclassification — endpoint products weight PE metadata completeness as a feature. `cnspec serve` compounds it by legitimately doing several things those classifiers score on: running as a SYSTEM service under `services.exe`, fetching and executing provider plugin binaries at runtime, and shelling out to PowerShell for host inventory checks. Every release also lands as a previously unseen hash with no cloud prevalence.

This does not make the problem go away on its own, but it is the cheapest available mitigation and removes an easy strike against the binary.

### DIFF
--- a/.github/.goreleaser-unstable.yml
+++ b/.github/.goreleaser-unstable.yml
@@ -6,6 +6,12 @@ version: 2
 project_name: cnspec
 env:
   - CGO_ENABLED=0
+# Generate the Windows version resource (VERSIONINFO) and application manifest.
+# Without it cnspec.exe reports no CompanyName/ProductName/FileDescription, which
+# contributes to heuristic AV/EDR misclassification of the binary.
+before:
+  hooks:
+    - go run github.com/tc-hib/go-winres@v0.3.3 make --in apps/cnspec/winres/winres.json --out apps/cnspec/rsrc --arch amd64,arm64 --file-version {{ .Version }} --product-version {{ .Version }}
 builds:
   - id: linux
     main: ./apps/cnspec/cnspec.go
@@ -49,7 +55,10 @@ builds:
           env:
             - QUILL_LOG_FILE=/tmp/quill-{{ .Target }}.log
   - id: windows
-    main: ./apps/cnspec/cnspec.go
+    # NOTE: this must stay a package path, not ./apps/cnspec/cnspec.go. Go only
+    # links rsrc_windows_*.syso when building a package directory; passing an
+    # explicit .go file silently drops the version resource and manifest.
+    main: ./apps/cnspec
     binary: cnspec
     goos:
       - windows
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,9 @@ gha-creds-*.json
 # goreleaser build folder
 dist/
 
+# generated Windows version resource (see `make cnspec/winres`)
+apps/cnspec/rsrc_windows_*.syso
+
 # claude local config. .claude/settings.json is shared team config and is
 # tracked; everything else under .claude/ is personal. Note this must be
 # ".claude/*" and not ".claude/": git does not descend into an ignored
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -6,6 +6,12 @@ version: 2
 project_name: cnspec
 env:
   - CGO_ENABLED=0
+# Generate the Windows version resource (VERSIONINFO) and application manifest.
+# Without it cnspec.exe reports no CompanyName/ProductName/FileDescription, which
+# contributes to heuristic AV/EDR misclassification of the binary.
+before:
+  hooks:
+    - go run github.com/tc-hib/go-winres@v0.3.3 make --in apps/cnspec/winres/winres.json --out apps/cnspec/rsrc --arch amd64,arm64 --file-version {{ .Version }} --product-version {{ .Version }}
 builds:
   - id: linux
     main: ./apps/cnspec/cnspec.go
@@ -51,7 +57,10 @@ builds:
           env:
             - QUILL_LOG_FILE=/tmp/quill-{{ .Target }}.log
   - id: windows
-    main: ./apps/cnspec/cnspec.go
+    # NOTE: this must stay a package path, not ./apps/cnspec/cnspec.go. Go only
+    # links rsrc_windows_*.syso when building a package directory; passing an
+    # explicit .go file silently drops the version resource and manifest.
+    main: ./apps/cnspec
     binary: cnspec
     goos:
       - windows
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,12 @@ MAJOR_VERSION=v13
 LDFLAGS="-s -w -X go.mondoo.com/mql/${MAJOR_VERSION}.Version=${LATEST_VERSION_TAG} -X go.mondoo.com/cnspec/${MAJOR_VERSION}.Version=${LATEST_VERSION_TAG} ${LDFLAGSEXTRA}"
 LDFLAGSDIST=-tags production -ldflags ${LDFLAGS}
 
+# Windows version resource generator. The PE version fields are numeric, so strip
+# the leading "v" from the tag (v13.32.1 -> 13.32.1) to match release builds,
+# which use goreleaser's already-stripped {{ .Version }}.
+GO_WINRES=github.com/tc-hib/go-winres@v0.3.3
+WINRES_VERSION=$(LATEST_VERSION_TAG:v%=%)
+
 .PHONY: info/ldflags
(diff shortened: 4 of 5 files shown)


chris-rock (commented):
Rebased onto `main` (it had gone conflicting on `.gitignore`, which grew a `.claude/*` comment block under the same lines) and pushed 17309aa77 with the same template trim we just made to the provider template in mondoohq/mql#9800. Your commit is otherwise untouched.

## What changed

Eleven of the manifest settings were go-winres defaults. Built the resource from both configs and diffed the manifest out of the `.syso`:

```
$ diff before.manifest after.manifest
(identical)
```

`dpi-awareness` stays — it is the one that is not a default (omitted, you get `dpiAware=true`/`system`) and is now the only manifest line in the file doing something.

On the version block, `FileVersion` and `ProductVersion` are filled from `--file-version` / `--product-version`, so the empty placeholders went, along with three fields that were shipping as entries with empty values:

```
version strings only in the old config: ['LegalTrademarks', 'PrivateBuild', 'SpecialBuild']
version strings only in the new one:    []
```

53 lines to 34, nine populated strings unchanged.

## Verified end to end

`make cnspec/build/windows` on the rebased branch, then read back out of the linked `cnspec.exe`:

```
VS_VERSION_INFO | https://mondoo.com/docs/cnspec | Mondoo, Inc.
Cloud-Native Security and Policy Framework | 13.33.2 | cnspec
Copyright Mondoo, Inc. All rights reserved. | cnspec.exe
manifest: present | execution level: asInvoker | longPathAware: true
```

| build | manifest | version strings |
| --- | --- | --- |
| `make cnspec/build/windows` (`./apps/cnspec`) | present, `asInvoker` | all nine |
| `go build apps/cnspec/cnspec.go` | **absent** | absent |
| `darwin/arm64` with both `.syso` present | absent (clean Mach-O) | absent |

The middle row is the pitfall your description calls out, confirmed rather than taken on trust.

## Two things I checked and cleared

- **Version shapes.** A `before` hook failing would break a release, so I fed go-winres the three forms this can produce — `13.33.2`, a `git describe` string `13.33.2-81-ge5c6a40c6`, and a snapshot `13.33.2-next`. All accepted, and the string lands verbatim in `FileVersion`. The Makefile uses `git describe --abbrev=0`, so it is always a clean tag anyway.
- **`goreleaser check`** reports "valid, but uses deprecated properties" on both configs — but it reports the same on `main` (`archives.format_overrides.format`), so it is pre-existing and nothing to do with this change. Worth knowing separately that `check` exits non-zero on it.

Confirmed your note about the edge config too: `.github/.goreleaser-edge.yml` mentions windows zero times.

## One thing I deliberately did not add

mondoohq/mql#9800 grew a cleanup trap for the generated objects, because there the bundler generates into a shared provider source tree during every release build. Here `cnspec/winres` is a `.PHONY` prerequisite of the windows build, so the objects are regenerated on every invocation and cannot go stale, and they sit in a gitignored path. Adding cleanup would be machinery without a failure mode behind it.

PR #3239, opened 2026-08-13
✨ sbom: include Windows print drivers

chris-rock:
A generated SBOM listed no print drivers at all, so nothing downstream could reason about them. This is the client half of print-driver vulnerability matching — the server side landed in mondoohq/server#18477, which can now match `pkg:windows-driver/<vendor>/<name>` but has had nothing emitting those purls.

Drivers have to be read from the print spooler rather than from installed software: a driver delivered as an INF/CAB package lands in the driver store and creates **no Add/Remove Programs entry**, so a software inventory cannot see it while the spooler is still loading it. The SBOM is the only place they appear.

Three parts: the pack query (windows-filtered), the `BomFields` entry, and the generator mapping that types them `windows-driver`.

## The load-bearing detail is the JSON key

`BomFields` is populated from the *compiled datapoint label*, and for a list resource that label is `<resource>.list` — so `windows.printerDrivers.list`, matching `packages.list` and `npm.packages.list`.

Verified by compiling the query rather than by reading the convention, because the convention has an exception sitting right next to it (`python.packages` has no `.list`). Getting it wrong **fails silently**: the field stays empty, no error is raised, the drivers are just missing. The fixture therefore carries the verbatim key, and I confirmed the test catches a mismatch by breaking the tag and watching it go red.

## Note on the two generators

`internal/sbom/generator` is a separate implementation from mql's `sbom/generator`, and `cnspec sbom` / `cnspec vuln` use **this** one. I nearly wired only the mql side, which would have looked correct and changed nothing here. The matching mql change is mondoohq/mql#9816; both are needed, for their own SBOM paths.

## Drivers without a purl

The resource returns an empty purl when the spooler reports no manufacturer or no name — a driver name alone is not an identity, since `PCL 6 Driver` names a page description language many vendors ship a driver for. Those are still listed, just without a purl, and the test covers it.

`internal/sbom/generator` had no tests; this adds the first.

## Not verified yet

**Whether the driver *name* actually matches the advisory side.** `Get-PrinterDriver` reports e.g. `RICOH PCL6 UniversalDriver V4.x` where Ricoh's advisory says `PCL6 Driver for Universal Print`. If those don't meet, the whole path matches nothing in production. That needs a real Windows host with a vendor driver installed, which I don't have — flagging it rather than letting it look finished.

## Test plan

```
go test ./internal/sbom/...
go build ./...
```

Green, including the new `TestPrinterDriversInSbom`.

### DIFF
--- a/internal/sbom/generator/generator.go
+++ b/internal/sbom/generator/generator.go
@@ -105,6 +105,22 @@ func GenerateBom(r *cr.Report) []*sbom.Sbom {
 				}
 			}
 
+			// Windows print drivers. Their purl comes from the resource and is
+			// empty when the spooler reports no manufacturer or no name -- a
+			// driver name alone does not identify a driver, since "PCL 6
+			// Driver" names a page description language that every printer
+			// vendor ships one for. Such a driver is still listed, without a
+			// purl.
+			for _, drv := range rb.PrinterDrivers {
+				bom.Packages = append(bom.Packages, &sbom.Package{
+					Name:    drv.Name,
+					Version: drv.Version,
+					Purl:    drv.Purl,
+					Cpes:    drv.CPEs,
+					Type:    "windows-driver",
+				})
+			}
+
 			if rb.Packages != nil {
 				for _, pkg := range rb.Packages {
 					bomPkg := &sbom.Package{
--- a/internal/sbom/generator/printer_drivers_test.go
+++ b/internal/sbom/generator/printer_drivers_test.go
@@ -22,6 +22,11 @@ func findPkg(pkgs []*sbom.Package, name string) *sbom.Package {
 
 // TestPrinterDriversInSbom covers Windows print drivers reaching the SBOM.
 //
+// The fixture is captured verbatim from a real scan of a Windows 11 Pro host
+// (platform windows 26200), not hand-written -- including the packed
+// DriverVersion unpacked to a dotted quad, which is how 2814751477605035
+// becomes 10.0.26100.8875.
+//
 // The load-bearing detail is the JSON key. BomFields is populated from the
 // compiled datapoint label, and for a list resource that label is
 // "<resource>.list" -- windows.printerDrivers.list, the same shape as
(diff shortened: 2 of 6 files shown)


chris-rock:
## Tested with real scans — and it found a release-ordering problem

I had only tested the generator against a fixture, so I ran actual `cnspec sbom` scans. The SBOM **never fails**, but with a provider that predates the resource it logs a compile error on **every scan, on every platform**:

```
! failed to compile error="failed to compile query 'windows.printerDrivers { name version purl }':
  cannot find field 'printerDrivers' in windows"
```

The platform filter gates **execution, not compilation**, so this fires on Linux and macOS too — not just Windows.

| os provider | target | result |
|---|---|---|
| released (no resource) | macOS local | exit 0, 1385 pkgs, status OK — **2 compile errors** |
| released (no resource) | alpine:3.19 container | exit 0, 15 pkgs, status OK — **2 compile errors** |
| built from mql#9816 branch | macOS local | exit 0, 1385 pkgs — **0 errors** |
| built from mql#9816 branch | alpine:3.19 container | exit 0, 15 pkgs — **0 errors** |

Verified by building and installing the os provider from the mql branch, re-running, and watching the errors disappear — then restoring the released provider and watching them come back.

Good news: it degrades gracefully. `status: 1`, no `error_message`, identical package counts, and zero `windows-driver` entries on non-Windows — the filter does its job once the query compiles.

### So this PR is release-coupled

`windows.printerDrivers` is in **no released mql tag** (latest is v13.32.2). Merging this before the os provider ships the resource would put two compile errors into every customer's scan output for no benefit.

**I've moved it to draft** so it doesn't merge early. The order is:

1. mondoohq/mql#9816 merges
2. an os provider release ships `windows.printerDrivers`
3. this PR comes out of draft

### Still not verified

**Windows itself.** I have no Windows host, so the resource has never run against a live spooler in this flow — and the driver *name* question stands (`Get-PrinterDriver` reports `RICOH PCL6 UniversalDriver V4.x`, the advisory says `PCL6 Driver for Universal Print`). Those are the two things a Windows box would settle.

chris-rock:
## Verified end to end on a real Windows host

Ran `cnspec sbom ssh` against a Windows 11 Pro machine (platform `windows 26200`). The full chain works:

```
asset: win11devbox | platform: windows 26200 | total packages: 171
windows-driver packages: 4
  Universal Print Class Driver          10.0.26100.8875  pkg:windows-driver/microsoft/universal-print-class-driver@10.0.26100.8875
  Microsoft Virtual Print Class Driver  10.0.26100.8875  pkg:windows-driver/microsoft/microsoft-virtual-print-class-driver@10.0.26100.8875
  Microsoft Print To PDF                10.0.26100.4484  pkg:windows-driver/microsoft/microsoft-print-to-pdf@10.0.26100.4484
  Microsoft IPP Class Driver            10.0.26100.8875  pkg:windows-driver/microsoft/microsoft-ipp-class-driver@10.0.26100.8875
```

Three things this confirms that were previously only assumed:

1. **The datapoint key.** `windows.printerDrivers.list` is right in production, not just in a fixture.
2. **The packed version unpacks correctly.** The spooler reports `DriverVersion: 2814751477605035`; that appears as `10.0.26100.8875`, matching the four-16-bit-field layout.
3. **The server side accepts these purls.** Round-tripped the captured packages through `PreProcessPackage` + `PurlToOsvEcosystems` in the server repo — the vendor namespace and name survive, and they resolve to `Ecosystem_WindowsDriver`.

The fixture is now the captured data rather than my hand-written one.

### A factual error in my own fixture, corrected

I had used **Microsoft Print To PDF as the example of a driver with no purl**. It reports `Manufacturer: Microsoft` and does carry one — every driver on the captured host did. The empty-purl path still needs covering, so it keeps a row, now marked synthetic rather than attributed to a real driver that behaves differently.

### Still open

The host has **only Microsoft built-in drivers**, so the vendor-name question is still unanswered: whether `Get-PrinterDriver`'s `RICOH PCL6 UniversalDriver V4.x` meets the advisory's `PCL6 Driver for Universal Print`. That needs a machine with a vendor print driver installed. Everything *structural* is now verified; what remains is purely whether the two naming conventions line up.

This PR stays in draft pending the os provider release, as above.

chris-rock:
Un-drafted: the blocker has shipped.

This was held because `windows.printerDrivers` and its `hardwareId` field did not exist in any **released** os provider — only in the module source. Provider releases are cut independently of the module tags, which is the distinction that made the original diagnosis take two runs.

**os provider 13.40.6** (published 2026-08-13 20:06 UTC) carries both. Verified on the released artifact rather than a branch build:

```
windows.printerDrivers:  list
windows.printerDriver:   configFile, dataFile, driverPath, environment, hardwareId,
                         infPath, manufacturer, modelVersion, name, printProcessor,
                         purl, version
```

Rebased onto main, which moves `go.mondoo.com/mql/v13` from `v13.33.0` to `v13.33.2`.

## Verification

The pack query compiles against the released provider (installed locally at exactly 13.40.6):

```
$ cnspec run local -c "windows.printerDrivers { name version purl }"
failed to retrieve printer drivers: sh: powershell.exe: command not found
windows.printerDrivers.list: no data available
```

It reaches execution and fails only at the platform step, which is the expected result on macOS. The control confirms a compile failure would look different:

```
$ cnspec run local -c "windows.notARealResource { name }"
failed to compile: cannot find field 'notARealResource' in windows
```

`go build ./internal/sbom/...` and `go test ./internal/sbom/...` are green.

## What is not covered

`pack_test.go` asserts the pack's query UIDs; it does not compile the MQL against a provider schema, so the manual run above is what establishes that the query resolves — not the test suite.

A live Windows run against 13.40.6 has not been repeated since the release (the test host is currently off-network). The committed `testdata/windows-print-drivers.json` is a real capture and the generator path is tested against it, so what is unconfirmed is specifically the provider→pack step on Windows, not the conversion.

PR #3243, opened 2026-08-13
✨ content: assess running PostgreSQL, MySQL, and Grafana servers

chris-rock:
## What this does

Three existing policies gain checks that read **live server state** through a connected provider, alongside the on-disk configuration they already assess.

### `mondoo-postgresql-security` — addresses @tas50's review

#3150 landed the OS-side PostgreSQL checks. Rather than ship a competing `mondoo-postgres-security` policy, the live-server checks are now a **Running Server Configuration** group in that policy, behind the `postgresdb` provider:

| Check | Reads | Category |
|---|---|---|
| TLS enabled for client connections | `pg_settings` | A04 |
| Passwords stored with SCRAM-SHA-256 | `pg_settings` | A04 |
| No `trust` rule in host-based auth | `pg_hba_file_rules` | A07 |
| Connections and disconnections logged | `pg_settings` | A09 |

The two halves are complementary, and the policy docs now say so: the host-based checks see the configuration the server *would* load on its next reload; the running-server checks see what it *is* enforcing, including anything set with `ALTER SYSTEM` (which lands in `postgresql.auto.conf` and is invisible to the on-disk checks).

### `mondoo-mysql-security`

Same treatment behind `mysqldb`: require secure transport (A04), accounts have a password or are locked (A07), no anonymous accounts (A07), `LOAD DATA LOCAL INFILE` disabled (A02). MariaDB's on-disk configuration stays with the MariaDB policy, but these read state MariaDB and Percona expose the same way MySQL does, so they apply to all three.

### `mondoo-grafana-security`

New **Authentication Security** group: per-user multi-factor enrollment (A07) and SAML single sign-on (A07). The existing policy covered service accounts, datasources, and alerting but had no user-authentication checks.

### Docs

Documents the OWASP Top 10 mapping parity invariant in `content/CLAUDE.md`, and records the deliberate LLM03/07/09/10 no-coverage cells next to the test that guards the mapping.

## Scope

This PR was split out of a larger branch. The 15 new policies that were on it now ship separately:

- 3337 — self-hosted databases (Cassandra, ClickHouse, Elasticsearch, OpenSearch, Redis, MongoDB, SQL Server)
- 3338 — SaaS and cloud platforms (ClickHouse Cloud, Neon, Netlify, Equinix Metal, HCP)
- 3339 — Active Directory, Jamf, Weaviate

Each branches off `main` independently — no stacking.

## Validation

- `cnspec policy lint ./content` — no new findings against a pristine-`main` baseline.
- `go test ./content/validation/...` — passes, including the OWASP parity test.

### DIFF
--- a/content/mondoo-postgresql-security.mql.yaml
+++ b/content/mondoo-postgresql-security.mql.yaml
@@ -3,33 +3,37 @@
 policies:
   - uid: mondoo-postgresql-security
     name: Mondoo PostgreSQL Security
-    version: 1.0.1
+    version: 1.1.0
     license: BUSL-1.1
     tags:
       mondoo.com/category: security
-      mondoo.com/platform: linux,postgresql
+      mondoo.com/platform: linux,postgresdb
     require:
       - provider: os
+      - provider: postgresdb
     authors:
       - name: Mondoo, Inc.
         email: hello@mondoo.com
     summary: Harden PostgreSQL server transport, host-based authentication, and audit logging
     docs:
       desc: |
-        The Mondoo PostgreSQL Security policy assesses the on-disk configuration of a self-managed PostgreSQL server. It inspects `postgresql.conf` together with the fragments pulled in by its `include`, `include_if_exists`, and `include_dir` directives, plus the `pg_hba.conf` host-based authentication rules and the `pg_ident.conf` identity mappings.
+        The Mondoo PostgreSQL Security policy assesses a self-managed PostgreSQL server from two directions. On a Linux host it inspects the on-disk configuration: `postgresql.conf` together with the fragments pulled in by its `include`, `include_if_exists`, and `include_dir` directives, plus the `pg_hba.conf` host-based authentication rules and the `pg_ident.conf` identity mappings. Against a running server it connects over SQL and reads the effective runtime state from `pg_settings` and `pg_hba_file_rules`.
 
-        This policy provides security checks across four areas:
+        This policy provides security checks across five areas:
 
         - Connection surface and TLS transport settings
         - Host-based authentication rules and identity mapping
         - Audit logging coverage and log file protection
         - Ownership and permissions on the configuration files themselves
+        - The effective settings a running server is enforcing right now
 
         **Scope and limitations**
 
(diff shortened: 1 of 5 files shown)


tas50:
https://github.com/mondoohq/cnspec/pull/3150 merged in the postgresql OS side checks. This will need to integrate into that

content/mondoo-postgresql-security.mql.yaml
```diff
@@ -3,33 +3,37 @@
 policies:
   - uid: mondoo-postgresql-security
     name: Mondoo PostgreSQL Security
-    version: 1.0.1
+    version: 1.1.0
     license: BUSL-1.1
     tags:
       mondoo.com/category: security
-      mondoo.com/platform: linux,postgresql
+      mondoo.com/platform: linux,postgresql,postgresdb
```
tas50:
```suggestion
      mondoo.com/platform: linux,postgresdb
```

tas50 (changes requested):

chris-rock:
Applied — `mondoo.com/platform: linux,postgresdb`. Also applied the same correction to `mondoo-mysql-security` (`linux,mysqldb`), where I'd introduced the identical mistake. `postgresql` and `mysql` were not real platform identifiers; each appeared exactly once in all of `content/`, in those two tags.

Also worth flagging: this PR was rebased onto `main` and **split into four**, since it had grown to 4,600 lines across 19 files. This one now contains only the running-server checks added to the existing PostgreSQL, MySQL, and Grafana policies — the PostgreSQL half is integrated into `mondoo-postgresql-security` as you asked, rather than the separate `mondoo-postgres-security` policy it used to add. The 15 new policies moved to #3337, #3338, and #3339, each branching off `main` independently.

Two policies from the original branch were dropped entirely: `mondoo-datadog-security` (all four checks read `datadog.org.*`, which no provider exposes) and a GitLab audit-event-streaming check (`gitlab.group.auditEventStreamingEnabled` does not exist). Both failed `cnspec policy lint`; neither field is in the installed providers or in mql v13.33.2.

tas50 (approved):

PR #3264, opened 2026-08-14
🧹 lint: report deprecated symbols used in query filters

### LINKED ISSUE #3262: Lint: query-deprecated-symbol never inspects filters:, so deprecated symbols in asset filters go unreported
## Summary

The `query-deprecated-symbol` lint rule only analyzes a query's `mql:` block. It never inspects `filters:`, so deprecated resources and fields used in asset-selection filters are silently unreported. The reported warning count is a **floor**, not a total, and a policy can lint clean while still depending on symbols that are on their way out.

This was noticed while migrating the GCP firewall checks off the deprecated `gcp.project.computeService.firewall.allowed` field in #3253: lint flagged 3 queries, but the same deprecated field is used in 5 more places in the same file that lint says nothing about.

## Mechanism

`walkQueryForDeprecatedSymbols` reads `q.Mql` and nothing else:

https://github.com/mondoohq/cnspec/blob/main/internal/bundle/lint_rules_deprecated_symbols.go#L55-L63

```go
func walkQueryForDeprecatedSymbols(schema resources.ResourcesSchema, conf mqlc.CompilerConfig, filename string, q *Mquery) []*Entry {
	if q == nil || q.Mql == "" {
		return nil
	}

	usage, _, err := mqlc.AnalyzeQuery(q.Mql, mqlc.EmptyPropsHandler, conf)
	...
}
```

`q.Filters` is never passed to `AnalyzeQuery`. Two distinct consequences:

1. **Queries with both `mql:` and `filters:`** get their `mql:` analyzed and their `filters:` skipped.
2. **Variant parent queries** — which carry `filters:` but no `mql:` — are dropped entirely by the `q.Mql == ""` early return before anything is analyzed.

The data is already in the right shape to fix this. A `filters: |` block unmarshals into a fully-formed `Mquery` with `.Mql` populated (`internal/bundle/bundle_ext.go:107-118`):

```go
var str string
err := node.Decode(&str)
if err == nil {
	x.Items = map[string]*Mquery{}
	x.Items[""] = &Mquery{
		Mql: str,
	}
	return nil
}
```

So each `Filters.Items` entry can be fed to the existing analysis path unchanged.

## Reproduction

On `main` at 843be46e:

```
$ cnspec policy lint content/mondoo-gcp-security.mql.yaml
 query-deprecated-symbol  warning  mondoo-gcp-security.mql.yaml  8307   query 'mondoo-gcp-security-firewall-no-overly-permissive-ingress-gcp' uses deprecated field 'gcp.project.computeService.firewall.allowed'
 query-deprecated-symbol  warning  mondoo-gcp-security.mql.yaml  23340  query 'mondoo-gcp-security-firewall-no-egress-all-ports-gcp' uses deprecated field 'gcp.project.computeService.firewall.allowed'
 query-deprecated-symbol  warning  mondoo-gcp-security.mql.yaml  25138  query 'mondoo-gcp-security-firewall-no-large-port-range-gcp' uses deprecated field 'gcp.project.computeService.firewall.allowed'
→ valid policy bundle(s)
```

Three warnings. But the same deprecated field appears in five more queries, each time inside `filters:`, and none are reported:

| line | query uid |
|---|---|
| 7640 | `mondoo-gcp-security-firewall-no-ingress-ssh-from-internet-gcp` |
| 7810 | `mondoo-gcp-security-firewall-no-ingress-rdp-from-internet-gcp` |
| 7978 | `mondoo-gcp-security-firewall-no-unrestricted-ingress-all-protocols-gcp` |
| 8144 | `mondoo-gcp-security-firewall-no-ingress-database-ports-from-internet-gcp` |
| 23176 | `mondoo-gcp-security-firewall-no-unrestricted-egress-gcp` |

For example, `content/mondoo-gcp-security.mql.yaml:7636-7640`:

```yaml
  - uid: mondoo-gcp-security-firewall-no-ingress-ssh-from-internet-gcp
    filters: |
      asset.platform == 'gcp-compute-firewall'
      gcp.compute.firewall.direction == "INGRESS"
      gcp.compute.firewall.allowed.any(ipProtocol == "tcp" && _['ports'] == empty || ...)
```

`gcp.compute.firewall.allowed` is the same deprecated field the rule flags three lines elsewhere.

This is one deprecated field in one file. Nothing currently measures the gap across `content/`, which is part of the problem.

## Why it matters

A deprecated symbol in `filters:` is arguably worse than one in `mql:`. `filters:` decides which assets a check applies to, so when the field is eventually removed the filter stops matching and the check silently **stops scoring assets** rather than failing loudly. Deprecation warnings exist to give us a window to migrate before that happens, and right now that window is invisible for every filter in the repo.

## Suggested fix

Visit `q.Filters.Items` alongside `q.Mql` in `walkQueryForDeprecatedSymbols`, and move the `q.Mql == ""` early return so it guards only the `mql:` analysis rather than the whole function.

Two details worth getting right:

- **Location and display ID.** Filter `Mquery` values built by `Filters.UnmarshalYAML` carry only `.Mql` — no `Uid` and no `FileContext`. Using them directly would emit `query 'at line 0' uses deprecated field ...`. The parent query's `FileContext` and `queryDisplayID` should be used instead, ideally with wording that distinguishes the two sites, e.g. `query 'X' uses deprecated field 'Y' in its filters`.
- **Deduplication.** A symbol deprecated in both `mql:` and `filters:` of the same query shouldn't produce two near-identical warnings pointing at the same line.

Test coverage for this lives in `internal/bundle/lint_rules_deprecated_symbols_test.go`; it currently exercises `walkQueryForDeprecatedSymbols` only through `q.Mql`. Cases worth adding: deprecated symbol in `filters:` only, in both `mql:` and `filters:`, and on a variant parent query that has `filters:` with an empty `mql:`.

## Note on fixing the GCP content

The five GCP usages above are deliberately not migrated in #3253. The replacement `allowedProtocols` is a `map[string][]string`, and the naive port translation is not safe in a filter: `allowedProtocols['tcp'] == empty` is true both when tcp allows every port **and** when tcp is absent entirely. A filter that is subtly wrong drops assets from scoring silently, so those need a `keys.contains("tcp")` presence guard and their own fixtures. Worth doing as a follow-up once this lint gap is closed and the warnings actually surface.

tas50:
Fixes #3262.

## What

`query-deprecated-symbol` only ever analyzed a query's `mql:`. Deprecated resources and fields used in `filters:` were silently unreported, so the warning count was a floor rather than a total and a policy could lint clean while still depending on symbols that are on their way out.

`walkQueryForDeprecatedSymbols` now analyzes `q.Filters.Items` alongside `q.Mql`. The `q.Mql == ""` early return no longer short-circuits the whole function — that return was separately dropping **variant parent queries**, which carry `filters:` but no `mql:` of their own, so those were never inspected at all.

The per-snippet analysis moved to a small `deprecatedSymbolsIn` helper. Symbol descriptions are built as strings and the entry construction is shared, so the existing warning text is byte-for-byte unchanged.

## Why filters matter as much as mql

`filters:` is asset selection. When a deprecated symbol is finally removed, a filter that references it stops matching and the check **silently stops scoring assets** rather than failing loudly. Deprecation warnings exist to give us a migration window before that happens, and until now that window was invisible for every filter in the repo.

## Reporting decisions

Two judgement calls worth flagging for review:

- **A symbol used in both `mql:` and `filters:` is reported once**, not twice. It is one migration to make, and both warnings would point at the same line.
- **The `in its filters` hint is added only for symbols that appear nowhere else.** Filter `Mquery` values built by `Filters.UnmarshalYAML` carry no `Uid` and no `FileContext`, so warnings necessarily borrow the parent query's line. Without the hint, a reader given a query's start line would search its `mql:` and find nothing. Where the symbol *is* reachable from `mql:`, the hint would be misleading, so it is omitted.

## Effect on ./content

`cnspec policy lint ./content` goes from 11 to 16 `query-deprecated-symbol` warnings. All 5 new ones are the deprecated `gcp.project.computeService.firewall.allowed` field used in filters, and **every pre-existing warning line is byte-identical**:

```
 query-deprecated-symbol  warning  mondoo-gcp-security.mql.yaml  7636   query 'mondoo-gcp-security-firewall-no-ingress-ssh-from-internet-gcp' uses deprecated field 'gcp.project.computeService.firewall.allowed' in its filters
 query-deprecated-symbol  warning  mondoo-gcp-security.mql.yaml  7806   query 'mondoo-gcp-security-firewall-no-ingress-rdp-from-internet-gcp' uses deprecated field 'gcp.project.computeService.firewall.allowed' in its filters
 query-deprecated-symbol  warning  mondoo-gcp-security.mql.yaml  7974   query 'mondoo-gcp-security-firewall-no-unrestricted-ingress-all-protocols-gcp' uses deprecated field 'gcp.project.computeService.firewall.allowed' in its filters
 query-deprecated-symbol  warning  mondoo-gcp-security.mql.yaml  8140   query 'mondoo-gcp-security-firewall-no-ingress-database-ports-from-internet-gcp' uses deprecated field 'gcp.project.computeService.firewall.allowed' in its filters
 query-deprecated-symbol  warning  mondoo-gcp-security.mql.yaml  23172  query 'mondoo-gcp-security-firewall-no-unrestricted-egress-gcp' uses deprecated field 'gcp.project.computeService.firewall.allowed' in its filters
```

**No content changes are needed to land this.** These are warnings, `make test/lint/content` passes no `--strict-rule`, and `policies_lint.yaml` only fails the build on SARIF `level == "error"`. Verified: `cnspec policy lint ./content` still exits 0 with 0 errors.

Migrating those 5 GCP filters is deliberately left out of this PR. The replacement `allowedProtocols` is a `map[string][]string`, and the naive translation is unsafe in a filter — `allowedProtocols['tcp'] == empty` is true both when tcp allows every port **and** when tcp is absent entirely. They need a `keys.contains("tcp")` presence guard and their own fixtures, and a subtly wrong filter drops assets from scoring silently. That belongs in its own PR now that the warnings are visible.

## Tests

Written first, and each was watched failing for the right reason before the fix (the three that exercise new detection reported 0 or 1 entries where 1 or 2 were expected):

| test | covers |
|---|---|
| `TestDeprecatedSymbol_FiltersOnly` | variant parent — filters, no mql; also asserts the parent's line is used |
| `TestDeprecatedSymbol_MqlAndFiltersReportSeparateSymbols` | distinct symbols in each site both reported |
| `TestDeprecatedSymbol_SameSymbolInMqlAndFiltersReportedOnce` | dedup, and no misleading `in its filters` hint |
| `TestDeprecatedSymbol_MultipleFilterItems` | list form of `filters:` — every item analyzed, not just the first |
| `TestDeprecatedSymbol_NoFiltersIsSafe` | nil `Filters`, and a query with neither mql nor filters |

`go test ./internal/bundle/...` passes; `gofmt` and `go vet` clean.

## Known remaining gap

Policy- and group-level `filters:` are still not inspected — only query filters are. Those are not `Mquery` values owned by a query, so they need a different message shape and display ID than `query 'X' uses ...`. Worth a follow-up issue rather than widening this change.

### DIFF
--- a/internal/bundle/lint_rules_deprecated_symbols.go
+++ b/internal/bundle/lint_rules_deprecated_symbols.go
@@ -52,18 +52,90 @@ func lintDeprecatedSymbols(schema resources.ResourcesSchema, conf mqlc.CompilerC
 	return entries
 }
 
+// walkQueryForDeprecatedSymbols reports deprecated symbols used by a query,
+// looking at both its mql and its filters. Filters matter as much as mql here:
+// when a deprecated symbol is finally removed, a filter that references it stops
+// matching and the check silently stops scoring assets instead of failing loudly.
+//
+// A symbol used in both places is reported once. The mql site wins, because that
+// is where a reader will find it; the "in its filters" hint is only added for
+// symbols that appear nowhere else, since every warning points at the query's own
+// line and filter Mqueries carry no file context of their own.
 func walkQueryForDeprecatedSymbols(schema resources.ResourcesSchema, conf mqlc.CompilerConfig, filename string, q *Mquery) []*Entry {
-	if q == nil || q.Mql == "" {
+	if q == nil {
 		return nil
 	}
 
-	usage, _, err := mqlc.AnalyzeQuery(q.Mql, mqlc.EmptyPropsHandler, conf)
-	if err != nil || usage == nil {
+	fromMql := deprecatedSymbolsIn(q.Mql, conf)
+
+	seen := make(map[string]struct{}, len(fromMql))
+	for _, symbol := range fromMql {
+		seen[symbol] = struct{}{}
+	}
+
+	var fromFilters []string
+	if q.Filters != nil {
+		// Sort for stable output across runs.
+		filterKeys := make([]string, 0, len(q.Filters.Items))
+		for key := range q.Filters.Items {
+			filterKeys = append(filterKeys, key)
+		}
+		sort.Strings(filterKeys)
+
+		for _, key := range filterKeys {
+			filter := q.Filters.Items[key]
+			if filter == nil {
+				continue
+			}
+			for _, symbol := range deprecatedSymbolsIn(filter.Mql, conf) {
+				if _, ok := seen[symbol]; ok {
+					continue
+				}
+				seen[symbol] = struct{}{}
+				fromFilters = append(fromFilters, symbol)
+			}
+		}
+	}
+
+	if len(fromMql) == 0 && len(fromFilters) == 0 {
 		return nil
 	}
 
 	loc := []Location{{File: filename, Line: q.FileContext.Line, Column: q.FileContext.Column}}
 	display := queryDisplayID(q)
+
+	entries := make([]*Entry, 0, len(fromMql)+len(fromFilters))
+	newEntry := func(symbol, where string) *Entry {
+		return &Entry{
+			RuleID:   QueryDeprecatedSymbolRuleID,
+			Level:    LevelWarning,
+			Message:  fmt.Sprintf("query '%s' uses %s%s", display, symbol, where),
+			Location: loc,
+		}
+	}
+	for _, symbol := range fromMql {
+		entries = append(entries, newEntry(symbol, ""))
+	}
+	for _, symbol := range fromFilters {
+		entries = append(entries, newEntry(symbol, " in its filters"))
+	}
+
+	return entries
+}
+
+// deprecatedSymbolsIn compiles a single MQL snippet and returns the deprecated
+// symbols it uses, as stable-sorted descriptions such as "deprecated resource
+// 'processes'" or "deprecated field 'file.basename'". Compile errors are
+// intentionally ignored — bundle-compile-error already surfaces them.
+func deprecatedSymbolsIn(query string, conf mqlc.CompilerConfig) []string {
+	if query == "" {
+		return nil
+	}
+
+	usage, _, err := mqlc.AnalyzeQuery(query, mqlc.EmptyPropsHandler, conf)
+	if err != nil || usage == nil {
+		return nil
+	}
 
 	// Sort for stable output across runs.
 	providerIDs := make([]string, 0, len(usage.Providers))
@@ -72,7 +144,7 @@ func walkQueryForDeprecatedSymbols(schema resources.ResourcesSchema, conf mqlc.C
 	}
 	sort.Strings(providerIDs)
 
-	var entries []*Entry
+	var symbols []string
 	for _, pid := range providerIDs {
 		pu := usage.Providers[pid]
 		resourceNames := make([]string, 0, len(pu.Resources))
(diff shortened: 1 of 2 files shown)


tas50:
Filed the remaining gap as #3265 — policy-group and query-pack `filters:` are still unanalyzed (245 + 207 blocks across 83 files, 51 of which reference real provider resources). Kept out of this PR because those aren't queries and need a different message shape and rule-ID decision.

PR #3299, opened 2026-08-15
⚡ perf(scan): reuse the compiled bundle across assets

MaxRink:
## What allocates

`localAssetScanner.prepareAsset` calls `Bundle.CompileExt` once per asset. A scan of N assets compiles the same bundle N times. Every compile reparses and recompiles every query in the bundle.

A heap profile of a 50 asset Kubernetes scan shows the cost. `scanDispatcher.scanSingleAsset` accounts for 420 MB of the 580 MB allocated at the sample point, and `bundleCache.compileQueries` alone accounts for 401 MB.

The compiled result depends only on the bundle and on the resource schema. `BundleCompileConf.Library` is not read during compilation, it is only attached to the returned map. Assets that share a schema can therefore reuse one compile.

## Change

Add a single entry cache, keyed by the bundle and by the resource schema. A miss falls back to a full compile, so a scan that mixes schemas stays correct.

Each caller gets its own map headers and its own `Library`. The values stay shared, because they point into the bundle that every asset already shares. The headers must not be shared: `SetBundleMap` and the policy hub insert into `Queries`, `Policies`, `Props` and `Frameworks` while a scan runs.

The cache is owned by the scan dispatcher, so it lives for one scan. `AssetJob.BundleCompileCache` is optional. When it is nil the bundle is compiled per asset as before.

Compiling under the cache lock also serialises the compiles that already mutate the shared bundle in place.

## Before and after

Command:

```
cnspec scan k8s pods50.yaml --discover pods --incognito \
  -f content/mondoo-kubernetes-security.mql.yaml --report-type none
```

50 pod assets. Total allocation and peak footprint read from `runtime/metrics` (`/gc/heap/allocs:bytes`, and `/memory/classes/total:bytes` minus `/memory/classes/heap/released:bytes`).

| Measurement | Before | After | Change |
| --- | --- | --- | --- |
| Total allocated | 2.63 GB | 0.75 GB | -71.4% |
| Peak Go footprint | 81.78 MB | 74.61 MB | -8.8% |

Peak footprint moves much less than total allocation because the live heap is small, about 30 MB, and the peak is set by garbage collector headroom. The gain is in churn, which is what costs time and garbage collector work under a memory limit.

Wall clock is not reported. The host carried competing load during these runs, so the timings were not reproducible enough to quote.

## Correctness

Unit tests in `policy/scan/bundle_compile_cache_test.go` cover reuse for one schema, per caller `Library`, per caller map headers, concurrent callers, recompilation for a new bundle, and the schema key. They pass under `-race`.

End to end, a 50 asset scan was run with and without the change and the JSON reports were compared. After normalising the asset MRNs, which are generated per run, all 50 assets and all 1100 score entries are identical.

`go test ./policy/...` passes.

## Known limit

`resources.ProviderInfo` carries only an id and a name, so the schema key cannot include provider versions. It uses the provider id set and the resource count. That is sufficient here, because the cache lives for one scan inside one process and a provider cannot change version while the process runs.

### DIFF
--- a/policy/scan/bundle_compile_cache.go
+++ b/policy/scan/bundle_compile_cache.go
@@ -1,0 +1,84 @@
+// Copyright Mondoo, Inc. 2024, 2026
+// SPDX-License-Identifier: BUSL-1.1
+
+package scan
+
+import (
+	"context"
+	"sort"
+	"strings"
+	"sync"
+
+	"go.mondoo.com/cnspec/v13/policy"
+	"go.mondoo.com/mql/v13/providers-sdk/v1/resources"
+)
+
+// bundleCompileCache reuses the compiled form of a policy bundle across the
+// assets of one scan.
+//
+// Bundle.CompileExt compiles every query in the bundle. The scanner calls it
+// once per asset, so a scan of N assets compiles the same bundle N times. The
+// result only depends on the bundle and on the resource schema, so a scan of
+// assets that share a schema can compile once and reuse the result.
+//
+// The cache holds a single entry. Scans are normally homogeneous, and a scan
+// that alternates between schemas still stays correct because a miss falls
+// back to a full compile.
+type bundleCompileCache struct {
+	mu     sync.Mutex
+	bundle *policy.Bundle
+	key    string
+	result *policy.PolicyBundleMap
+}
+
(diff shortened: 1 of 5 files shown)


tas50:
@arlimus

PR #3301, opened 2026-08-15
✨ scan: report CPU consumption alongside memory

vjeffrey:
Extends the scan telemetry from #3274 with CPU usage. Read from the same `runtime/metrics` call the memory sampler already makes, so the marginal cost is a handful of float reads once a second.

Renames `MemTracker` → `ResourceTracker` (and `MemSnapshot` → `ResourceSnapshot`), since the type now carries both, and splits the file it lived in — which was about to outgrow itself — into `tracker.go`, `sample.go`, `cgroup.go`.

## New metrics

`cnspec.scan.cpu.{busy,available,user,gc,scavenge}_seconds`, `cnspec.scan.cpu.{gc_fraction,utilization,gomaxprocs}`. No proto change; `policy.Metric` is namespace-based.

## Three ways CPU differs from memory

These drove the design and are where the bugs would have been.

**Cumulative, not instantaneous.** Memory is a level, so the tracker keeps peaks. The `/cpu/classes/*` counters accumulate since *process* start, so the tracker takes a baseline at the run's first observation and reports the difference. Without that, a scan starting in an already-warm process is charged for CPU burned before it began.

**`total` is CPU available, not CPU consumed** — it includes idle. Burned CPU is `total − idle`, which is what `busy_seconds` reports. Reporting `total` as usage would overstate every scan by its idle share, which for an I/O-bound scan is nearly all of it.

**The presence rule inverts.** A live process always has a non-zero memory footprint, so zero there means "did not resolve" and is omitted. But a scan short enough legitimately burns ~0 CPU-seconds, so **zero CPU is a real measurement** and must be reported. Presence tracks whether the metric names *resolved*, not whether the value is non-zero — keying on value would silently drop every fast scan. There's a test named for exactly this.

Ratios are omitted when their denominator is zero rather than reported as `0.0`, which would read as "no time in GC" when the truth is "not enough CPU to say".

## Verified end to end

A local scan, as the values landed upstream:

```
duration           15173 ms
gomaxprocs         12
available_seconds  181.18    (~= 15.17s x 12 cores)
busy_seconds       1.192     (= user 1.016 + gc 0.176 + scavenge 0)
gc_fraction        0.148
utilization        0.0066
```

The arithmetic cross-checks three independent ways: `available` matches wall-time × GOMAXPROCS, the three classes sum exactly to `busy`, and `busy` matches the local `DEBUG_PROVIDER_MEMORY` line to the digit.

It's also immediately informative — **utilization 0.66%** says scans are overwhelmingly waiting rather than computing, while **~15% of the CPU they do burn goes to GC**.

## Testing

12 new tests. `make test/lint` 0 issues; `policy/scanstats`, `policy/scan`, `internal/datalakes/sqlite`, `policy` all pass under `-race`.

## Notes for review

- `DEBUG_PROVIDER_MEMORY` keeps its name (renaming an operator-facing env var isn't worth it) but now also logs `cpu_busy_seconds`; the log message changed to "resource usage after asset scan".
- No ADR this time, by request. The design rationale lives in the code comments and this description.

### DIFF
--- a/docs/adr/0004-scan-memory-telemetry.md
+++ b/docs/adr/0004-scan-memory-telemetry.md
@@ -31,7 +31,7 @@ This ADR accepts that boundary rather than building a second transport. Adding m
 
 ### 2. Sampling: `runtime/metrics`, not `MemStats.Alloc`
 
-A new `scanstats.MemTracker` owns a process-wide high-water mark, updated by a single sampler goroutine started in `RunLocalScan` and stopped when it returns.
+A new `scanstats.MemTracker` owns a process-wide high-water mark, updated by a single sampler goroutine started in `RunLocalScan` and stopped when it returns. (Renamed to `ResourceTracker` when scan CPU telemetry was added alongside it — see #3301 — since the type now carries both.)
 
 The sampled quantity is the Go runtime footprint:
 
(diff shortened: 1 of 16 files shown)


vjeffrey:
Both findings fixed in the 2nd commit:

- **`gomaxprcs` typo** — corrected throughout `tracker.go` to `gomaxprocs`, matching the `GOMAXPROCS` field and `MetricCPUGOMAXPROCS` constant it's read from. The longer name also broke struct-literal alignment, which `gofmt -l` caught but `go build`/`go vet` didn't — fixed with the same `gofmt -w` pass.
- **Stale ADR-0004 references** — `scanstats.MemTracker`, `policy/scanstats/mem.go`, and `logMemoryStats` were all renamed or split by this PR. Annotated each in place with a pointer to this PR rather than rewriting the ADR's history, since it was accurate when written.

PR #3302, opened 2026-08-15
👷 validation: open weekly PRs for the validators' pinned upstreams

tas50:
## What

Adds a weekly workflow that opens a pull request for every upstream the variant and remediation validators are pinned to, plus the shared pin registry and bumper behind it.

## Why

These validators are only as good as what they check against, and every one of those is pinned somewhere:

| Pin | Declared in |
|-----|-------------|
| linter releases (`cfn-lint`, `ansible-lint`, `cookstyle`, `tflint`) | `.github/workflows/validate-remediation.yaml` |
| CLI release artifacts (`bicep`, `doctl`, `glab`, `hcloud`, `databricks`) — version **and** SHA-256 | same |
| tflint ruleset plugins, Terraform provider `~>` constraints | `content/validation/validate_terraform_remediation.py` |
| OpenAPI specs pinned to a commit | `content/validation/validators/openapi.py` |
| checked-in CLI grammars | the `_meta` block of each `cmd_data/*.json` |

Dependabot covers `gomod` and `github-actions`. Nothing watches any of the above. Left alone they rot quietly: the validators keep passing while checking remediation against a cloud API that has moved on.

`validation-upstream-drift.yaml` already *reports* this into one long-lived issue, but a report is easy to stop reading — and it was blind to the Terraform side entirely. Adding it turned up **eight providers a full major behind**:

```
terraform-provider-aws        ~> 5.0   ->  ~> 6.0   (hashicorp/aws 6.60.0)
terraform-provider-azurerm    ~> 4.0   ->  ~> 5.0
terraform-provider-google     ~> 6.0   ->  ~> 7.0
terraform-provider-oci        ~> 6.0   ->  ~> 8.0
terraform-provider-gitlab     ~> 17.0  ->  ~> 19.0
terraform-provider-okta       ~> 4.0   ->  ~> 6.0
terraform-provider-time       ~> 0.12  ->  ~> 0.14
terraform-provider-alicloud   ~> 1.0   ->  ~> 2.0
```

plus `cookstyle`, three OpenAPI specs, and the vercel CLI grammar (56.4.1 against 59.1.3). 13 of 43 pins behind.

## How

**`content/validation/upstream_pins.py`** — one registry, shared by the reporter and the bumper. It reads every pin out of the file that declares it and can write it back, so there is no second copy of a version anywhere and the registry cannot itself go stale. Terraform providers and tflint rulesets are discovered by *parsing* `PROVIDER_MAP` / `TFLINT_PLUGIN_MAP`, so adding a provider there puts it under watch with no second list to update.

**`content/validation/bump_upstream_versions.py`** — applies a bump, re-checksums an artifact, renders the PR body.

**`content/validation/check_upstream_versions.py`** — now a thin renderer over the registry. CLI is unchanged apart from a new `--format json`.

**`.github/workflows/validation-dependency-updates.yaml`** — Mondays 05:00 UTC (an hour before the drift report) and `workflow_dispatch`. A `discover` job emits a matrix of pins that are behind; one `bump` job each opens a PR. Grammars get dedicated jobs that install the tool, re-run the dump script, and run the matching validator so the PR body says up front whether the new CLI broke anything.

### Three things worth a look in review

1. **A CLI's version and its SHA-256 cannot move apart.** The bumper reads the download URL back out of the workflow's own `curl` line, fetches *that*, and hashes it — so it digests exactly what CI fetches, even if a tool later switches asset name or host. `--verify-checksums` re-derives all five at their current pins:

   ```
   [PASS] bicep        v0.46.1   …/bicep-linux-x64
   [PASS] doctl        v1.166.0  …/doctl-1.166.0-linux-amd64.tar.gz
   [PASS] glab         v1.113.0  …/glab_1.113.0_linux_amd64.tar.gz
   [PASS] hcloud       v1.67.0   …/hcloud-linux-amd64.tar.gz
   [PASS] databricks   v1.12.1   …/databricks_cli_1.12.1_linux_amd64.zip
   ```

   Every digest reproduces the one already pinned, which is what makes the machinery trustworthy.

2. **Branches carry the pin, not the version** (`deps/validation/<slug>`). A version in the branch name would abandon last week's branch and open a second PR on every upstream release; as written, `peter-evans/create-pull-request` updates the one already open. This is the property that makes a weekly cadence safe.

3. **`validate-remediation.yaml` only triggered on `content/**`.** A PR bumping a pin *inside that file* would never run the validators it configures — so the auto-bumps would have merged unvalidated, which is the one thing opening them as PRs is meant to prevent. Its `paths:` filter now includes itself.

Bump PRs use the `mondoo-mergebot` App token for the same reason `cnquery-update.yml` does: PRs opened with `GITHUB_TOKEN` do not trigger workflows, and CI's verdict is the entire point of opening one.

## Deliberately not a gate

A bump is a judgement call. A new `cfn-lint` or tflint ruleset can legitimately start failing snippets that were fine, and a Terraform provider major will rename resources out from under a lot of them. The tooling does the mechanical half and stops; each PR body carries a kind-specific note on what a red run most likely means, so triage doesn't start from scratch. Closing a bump PR is a valid answer — the branch reopens when upstream moves again.

## Testing

- `--verify-checksums` passes for all five artifact CLIs (above).
- Applied and reverted one bump of each shape — plain version string (`cookstyle`), Terraform constraint (`okta`), commit SHA (`grafana spec`) — and confirmed each produced a correct one-line diff in the right file.
- Exercised `--list`, `--all --dry-run`, `--only`, `--json`, `--pr-body`, `--github-output`, `--sync-dump-pins`.
- Ran the `discover` job's `jq` filter against real checker output for both the all-pins and single-pin cases.
- All three workflow files parse; `typos` clean.

The schedule only starts once this is on `main` — scheduled workflows run from the default branch. First run will open ~12 PRs clearing the backlog; steady state is 0–3 a week.


tas50:
Review comments addressed in 406ccb31.

**Token scoping by hostname** (CodeQL, `upstream_pins.py:76`) — fixed. `"api.github.com" in url` also matches a host that merely mentions it in a path or query, so the token would go to that host. Now compares `urlparse(url).hostname`. Every URL here is built from constants in the file, so this was hardening rather than a live hole — but a credential-scoping check shouldn't depend on that staying true.

```
OK   authed=True  https://api.github.com/repos/x/releases/latest
OK   authed=False https://evil.com/?x=api.github.com
OK   authed=False https://api.github.com.evil.com/x
OK   authed=False https://pypi.org/pypi/cfn-lint/json
```

**`$GITHUB_OUTPUT` injection** (`bump_upstream_versions.py:229`) — fixed, and at the root rather than only at the write. The real problem is that an upstream string was trusted as a single-line token by everything downstream: `$GITHUB_OUTPUT`, the markdown table, the PR body. Every resolver now passes its value through `as_token()`, which rejects anything that isn't version-shaped as `"unknown"` — the fail-safe reading, since an upstream we can't parse should stop the bump, not feed a controlled string into a PR title. The heredoc delimiter form went in too, as the layer that stays correct if a future resolver returns something richer.

```
OK       '1.2.3-rc.1'         -> '1.2.3-rc.1'          # real values survive
OK       '2.0.0+build5'       -> '2.0.0+build5'
OK       'a0e7cfa11b0d0…'     -> 'a0e7cfa11b0d0…'      # commit SHAs too
REJECTED '1.2.3\ntitle=pwned' -> 'unknown'
REJECTED '1.2.3\r\nbranch=…'  -> 'unknown'
REJECTED '$(whoami)'          -> 'unknown'
```

All 43 pins still resolve identically after the change, and `--verify-checksums` still reproduces all five artifact digests.

**`max()` on an empty list** (`check_upstream_versions.py:43`) — fixed with `default=0`. Worth doing beyond the trivial guard: every discoverer returns `[]` when the file it reads is absent, so a repo layout change would have surfaced as a `ValueError` rather than an empty report.

**Workflow permissions** — confirming it's intentional, as you read it. `contents: read` throughout; the mergebot App token carries the write scope, matching `cnquery-update.yml`. The App token is required rather than incidental: PRs opened with `GITHUB_TOKEN` don't trigger workflows, and CI's verdict on a bump is the entire point of opening one.

PR #3319, opened 2026-08-16
✨ alibaba: use the provider's typed verdicts and fix range-blind port matching

tas50:
The alicloud provider has grown computed verdicts (`isPublic`, `policyAllowsPublicAccess`, `allowsAdminAccess`) and typed references (prefix lists, KMS keys, resource groups) that these checks were working around. This reworks the existing checks onto them, and fixes three cases where the workaround was returning the wrong answer.

No new checks here. New service coverage is in a follow-up stacked on this branch.

## Per-resource asset platforms

21 checks iterated an account-wide collection with `.all()`, so one bad bucket failed the whole account as a single finding. All 15 alicloud per-resource platforms are in the auto-discovery set (`providers/alicloud/resources/discovery.go`), so OSS, RDS, PolarDB, Redis, MongoDB and SLB checks now bind to the resource they are about and score per asset. Same rework as PRs #3034–#3038 did for DO/OCI/Azure/GCP/AWS.

## OSS bucket policy

`policy-no-anonymous` parsed the policy document in MQL and tested `Principal.contains("*")`, which reads a string and a list differently. Object Storage Service publishes its own verdict as `policyAllowsPublicAccess`.

## Security-group port ranges — correctness fix

The four ingress checks compared `portRange` to the literals `"22/22"`, `"-1/-1"` and `"1/65535"`, so a rule of `20/25` or `1/1024` opened SSH and passed clean.

MQL has no string-to-int conversion, so containment is encoded as a regex over the `from/to` string: one alternation for "lower bound ≤ port", one for "upper bound ≥ port". Each half is verified exhaustively over 0–65535 in Python and re-checked in `mql run`:

```
20/25     [ok]      # now caught, previously passed
1/1024    [ok]      # now caught, previously passed
23/23     [failed]  # correctly not flagged
```

The same checks now resolve `sourcePrefixList`, so a rule sourced from a prefix list holding `0.0.0.0/0` is no longer invisible.

## KMS rotation interval — correctness fix

Inferring the interval from `nextRotationDate` passed a key with a 400-day interval that had rotated 50 days ago. Now reads the configured `rotationInterval`, bounded by a regex for the same string-to-int reason (verified to match exactly `0s`–`31536000s`).

## ActionTrail and VPC flow logs

Both asserted only that the thing was switched on. A trail with `EventRW` of `Write` records the change but not the reconnaissance that preceded it; a flow log with `TrafficType` of `Drop` records the connections that failed and not the ones that succeeded. Both now assert full capture, and both titles say so.

## Test evidence

Five new fail fixtures cover exactly what the old queries missed, and each fails only because of this change:

- `ecs-security-group-no-public-ssh` → `fail/world-open-port-span` (`20/25`)
- `ecs-security-group-no-public-rdp` → `fail/world-open-port-span` (`3300/3400`)
- `ecs-security-group-no-public-ipv6-admin` → `fail/world-open-ipv6-port-span` (`1/1024`)
- `actiontrail-trail-enabled` → `fail/write-events-only`
- `vpc-flow-logs-enabled` → `fail/rejected-traffic-only`

- `cnspec policy lint ./content` — valid, no new warnings
- `TestTerraformVariants` (alibaba) — all pass
- `TestRemediationSatisfiesCheck` (alibaba) — all pass, no budget entries added
- `validate_terraform_remediation.py alibaba` — 33 passed, 0 failed

### DIFF
--- a/content/iac-variant-testdata/mondoo-alibaba-security/mondoo-alibaba-security-actiontrail-trail-enabled-terraform-hcl/fail/write-events-only/main.tf
+++ b/content/iac-variant-testdata/mondoo-alibaba-security/mondoo-alibaba-security-actiontrail-trail-enabled-terraform-hcl/fail/write-events-only/main.tf
@@ -1,0 +1,10 @@
+# The trail is enabled, but it records only write operations. Read operations,
+# which is where reconnaissance and data access show up, go unrecorded.
+resource "alicloud_actiontrail_trail" "write_only" {
+  trail_name      = "write-only-audit"
+  oss_bucket_name = alicloud_oss_bucket.audit.bucket
+  oss_key_prefix  = "actiontrail/"
+  event_rw        = "Write"
+  trail_region    = "All"
+  status          = "Enable"
+}
--- a/content/iac-variant-testdata/mondoo-alibaba-security/mondoo-alibaba-security-ecs-security-group-no-public-ipv6-admin-terraform-hcl/fail/world-open-ipv6-port-span/main.tf
+++ b/content/iac-variant-testdata/mondoo-alibaba-security/mondoo-alibaba-security-ecs-security-group-no-public-ipv6-admin-terraform-hcl/fail/world-open-ipv6-port-span/main.tf
@@ -1,0 +1,10 @@
+# An IPv6 rule whose range spans port 22 opens SSH to the whole IPv6 internet.
+resource "alicloud_security_group_rule" "ipv6_span" {
+  security_group_id = alicloud_security_group.app.id
+  type              = "ingress"
+  ip_protocol       = "tcp"
+  port_range        = "1/1024"
+  ipv6_cidr_ip      = "::/0"
+  policy            = "accept"
+  priority          = 1
+}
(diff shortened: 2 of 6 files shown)


tas50:
Added 4166ec9f, prompted by a review on #3322 that found a Terraform plan/state variant asserting less than its `-terraform-hcl` sibling. I wrote the audit that would have caught it and ran it over this policy too; it found two long-standing instances here, disagreeing in opposite directions:

- **`oss-bucket-encryption-enabled`** — the HCL variant requires an `sse_algorithm` inside the `server_side_encryption_rule` block. Plan and state accepted the bare block, so a rule declaring no algorithm passed there and failed in HCL.
- **`kms-key-rotation-enabled`** — the HCL variant evaluates only encrypt/decrypt keys, since a signing key has no rotation setting to enable. Plan and state evaluated every key, so a signing key was reported as a finding an operator cannot fix.

The second is a false positive rather than a missed finding, which is the more expensive kind to ship.

The audit now reports no remaining gap in either policy.

PR #3343, opened 2026-08-16
🧹 azure: remove the checks for Azure services that no longer exist

tas50:
Nineteen checks in `mondoo-azure-security` target Azure services that are past retirement. They cannot fail, and their Terraform remediation cannot be applied.

| Service | Retired | Checks |
|---|---|---|
| Log Profiles | 2023-09-01 | 3 |
| MySQL Single Server | 2024-09-16 | 7 |
| PostgreSQL Single Server | 2025-03-28 | 9 |

## Why removal rather than repair

**The runtime variants report compliant on every scan.** They read `azure.subscription.monitor.logProfiles`, `azure.subscription.mySql.server` and `azure.subscription.postgreSql.server`. Those collections are empty on any current subscription, and `.all()` over an empty list is a vacuous pass. A check that cannot fail is worse than no check: it puts a green tick against a control nobody is actually enforcing.

**The Terraform variants cannot match anything.** azurerm removed the resource types when the services retired, verified against the pinned constraint:

```
$ terraform providers schema -json    # hashicorp/azurerm ~> 5.0 -> v5.1.0
azurerm_monitor_log_profile         ** MISSING **
azurerm_mysql_server                ** MISSING **
azurerm_mysql_firewall_rule         ** MISSING **
azurerm_postgresql_server           ** MISSING **
azurerm_postgresql_configuration    ** MISSING **
azurerm_postgresql_firewall_rule    ** MISSING **
```

A filter that matches nothing is indistinguishable from a passing check in a report, and the remediation told readers to write HCL that no longer plans.

There is no version of these checks worth keeping: the resource they describe cannot be running.

## What is kept

The **Flexible Server** equivalents cover the same controls on the live services and are untouched — `ensure-that-ssl-enabled-latest-version-mysql` and `ensure-that-ssl-enabled-postgresql`.

The MySQL TLS check carried a Single Server datapoint *alongside* its Flexible Server one in the Terraform and Bicep variants. That half is dropped, along with its `mysql-server-*` / `single-server-*` fixtures, so the check no longer asserts against a retired resource type. Its Flexible Server fixtures stay.

**`defender-auto-provisioning-enabled` is deliberately kept.** Microsoft retired the Log Analytics agent it provisioned and azurerm dropped `azurerm_security_center_auto_provisioning`, but `Microsoft.Security/autoProvisioningSettings` still answers, so the runtime and Bicep variants can still evaluate it. Only its three dead Terraform variants and their remediation are removed. Worth a second opinion — if you consider the setting meaningless now that the agent is gone, it should follow the other nineteen.

## Follow-up on the provider

Filed against `mql` to drop the now-unreachable provider resources: `azure.subscription.mySql.server`, `azure.subscription.postgreSql.server` and `azure.subscription.monitor.logProfiles`. Until that lands the fields simply return empty, so this PR stands on its own.

## Why nothing caught it

Found by comparing every `resource "..."` in `content/` against `terraform providers schema -json` at each pinned constraint. tflint ships rulesets only for aws, azurerm and google, and none of them flags an unknown resource type, so all of this passed the terraform validator. The same sweep found 26 occurrences across 4 policies; the others are #3340 (dns), #3341 (aws) and #3342 (gcp). A follow-up will teach `remediation/code/terraform.py` to assert resource-type existence.

## Verification

| Gate | Result |
|---|---|
| `cnspec policy lint` | valid |
| Azure Terraform + Bicep fixtures | `ok` (314s) |
| Fixture coverage gate | 100% on every policy, no orphans |
| Azure command validator | 721 pass / 0 fail |
| terraform validator (tflint) | 1,547 pass / 0 fail |
| Compliance mappings, bundle smoke scans | `ok` |
| `remediation-budget.json` | no entries left stale |
| `typos` | clean |

109 files changed, 394 insertions, 4,958 deletions.


tas50:
Provider follow-up filed: mondoohq/mql#9984 — remove `azure.subscription.mySql.server`, `azure.subscription.postgreSql.server` and `azure.subscription.monitor.logProfiles`, and decide whether `monitoringAgentAutoProvision` still means anything.

PR #3371, opened 2026-08-16
📝 content: rewrite OS policy check descriptions to state what is checked and why it matters

tas50:
## What this is

A full pass over every check description in the four operating system policies, asking two things of each: does the first sentence say what the check verifies, and would a reader learn the risk of ignoring it?

**298 descriptions reviewed, 244 rewritten.**

| Policy | Descs | Rewritten |
|---|---|---|
| Linux | 120 | 116 |
| Windows | 88 | 88 |
| macOS | 34 | 34 |
| FreeBSD | 56 | 8 |
| 5 smaller OS policies | 22 | 0 |

## The shape

Every description now follows one form:

> **This check verifies that Address Space Layout Randomization is fully enabled.**
>
> ASLR places a program's stack, heap, shared libraries, and executable at different memory addresses every time it runs, so an attacker cannot know in advance where anything is. The check reads the kernel parameter `kernel.randomize_va_space`, which must be `2`. A value of `1` randomizes everything except the `brk` heap, and `0` turns randomization off entirely.
>
> **Why this matters**
>
> Memory corruption exploits need addresses…

First sentence says what is verified, in capability terms. Second paragraph names the instrument and the value. Then the concrete risk. The setting is how the check measures the capability, never the subject of the description.

## What was wrong

**Linux and macOS** carried the same templated prose. Measured over the 120 Linux descriptions before this change: 72 ended in a paragraph that only restated the opener, 51 never named the value the check requires, 43 substituted "attack surface" or "principle of least functionality" for a concrete risk, 37 name-dropped compliance without naming a control, 14 hedged with "may contain vulnerabilities that could be exploited". Those counts are now 0, 0, 2, 0, 0. macOS was worse proportionally: 33 of 34 abstract, 20 hedged.

The six Linux kernel-module descriptions were byte-identical apart from the module name. The 19 "service is not enabled" descriptions shared one paragraph between them. Those now carry the risk specific to each protocol.

**Windows** had a different problem. 71 of 88 descriptions were verbatim CIS benchmark text with **no risk section at all**, typically ending "The recommended state for this setting is: `Disabled`." Accurate, and no help to someone deciding whether a finding matters. Each now explains the mechanism: why RC4 in Kerberos is what makes Kerberoasting work, why LLMNR yields credentials on an internal network within minutes, why WDigest turns a compromised server into plaintext passwords rather than hashes, and why `CrashOnAuditFail` is the one setting here where the secure-sounding option is the wrong one.

**FreeBSD** was already concise and capability-led and needed targeted work, not a rewrite.

**Left alone entirely:** `linux-workstation-security`, `linux-snmp-policy`, `linux-operational-policy`, `windows-workstation-security`, `windows-11-compatibility`. All 22 descriptions already meet the standard.

## Correctness fixes found while reading

- `permissions-on-etcsshsshd-config-are-configured` described only ownership while the query also forbids group and other read, write, and execute.
- `ssh-x11-forwarding-is-disabled` had the risk backwards. X11 forwarding exposes the *connecting client's* display, not the server.
- `ssh-protocol-is-set-to-2` discussed SSH1 as a live threat without noting the check is filtered to OpenSSH 6.0 through 7.5.
- **15 SSH audit sections told the reader to run `sshd -T`.** Those checks read `sshd.config.params`, which parses the config file and never shells out. For the seven directives whose OpenSSH default is already secure, `sshd -T` prints the passing value on a host where the directive is absent and the check fails, so anyone verifying a finding would have concluded the scanner was wrong. They now grep the config file and its `Include` directory and say explicitly that an absent directive fails.

## Validation

- `cnspec policy lint ./content` — valid
- `go test ./internal/bundle/...` — ok
- `typos` — clean on all four files
- All 298 descriptions confirmed to open with "This check" and carry a `**Why this matters**` section; no em dashes or double-hyphen asides
- Every `grep` command added to an audit section was executed against a fixture `sshd_config` with an indented directive, a commented directive, an `Include` directory, and a `Match` block
- No MQL, filters, tags, impact values, or remediation content changed. Prose only.

Versions: linux 2.8.1→2.8.2, macos 1.4.3→1.4.4, freebsd 1.2.2→1.2.3, windows 2.5.1→2.5.2.

## One thing for a maintainer, not fixed here

**`ssh-protocol-is-set-to-2` never evaluates on any supported distribution**, since its filter caps at OpenSSH 7.5, and where it does apply an absent `Protocol` directive fails a securely configured host. Filed as #3374, since fixing it changes behavior rather than prose.

An earlier revision of this description also claimed `network-access-named-pipes-that-can-be-accessed-anonymously-is-set-to-none` had no `mql:`. That was wrong. It is a variant parent, and both the `-dc` and `-ms` variants carry working queries; variant parents legitimately have no `mql` of their own. I checked every policy in `content/` and there are no queries that lack both `mql` and `variants`.


tas50:
Pushed `cdbedb54` reworking the kernel descriptions after review feedback.

The previous versions led with the sysctl parameter and treated the security feature as an aside, so a reader met `kernel.randomize_va_space` before they met ASLR. That is backwards: the parameter is how the check measures the capability, not what the check is about.

All 18 checks that assert on a kernel parameter now open with the capability and present the parameter as the instrument:

> Address Space Layout Randomization is the kernel feature that places a program's stack, heap, shared libraries, and executable at different memory addresses every time it runs. Nothing lands where it landed last time, so an attacker cannot know in advance where anything is.
>
> This check confirms that ASLR is running in its full mode by reading the kernel parameter `kernel.randomize_va_space`, which must be `2`. A value of `1` randomizes everything except the `brk` heap, and `0` turns randomization off entirely.

The subject of each description is now the named capability rather than the sysctl key: ASLR, kernel pointer hiding, low-address reservation, kernel ring buffer access, unprivileged eBPF, the Yama ptrace scope, core dump generation, IP forwarding, ICMP redirect generation and acceptance, source routing, martian packet logging, broadcast echo replies, bogus ICMP error handling, reverse path filtering, SYN cookies, and IPv6 router advertisements.

The six kernel module checks had the same problem, opening with `/etc/modprobe.d/` rather than with the protocol. They now name the protocol and what it is for first, then present the modprobe configuration as the means of confirming the module cannot load.

Two things surfaced while rewriting that are now stated where they apply:

- The redirect, source-route, forwarding, and router-advertisement checks accept an **unset** parameter as well as `0`. Users hitting these will want to know that.
- The reverse path filtering check requires **strict** mode. A host in loose mode, `rp_filter=2`, does not pass, which is worth saying out loud since loose mode is a legitimate configuration on multi-homed hosts.

Lint, `go test ./internal/bundle/...`, and `typos` all still clean.

PR #3389, opened 2026-08-16
🛑 reporter: remove the deprecated ScoreValue.score field

tas50:
Part of the v14 deprecation sweep. Marker: `// FIXME: DEPRECATED, remove in v13, replaced by riskScore below`.

## Rebased onto `main`

This PR previously targeted `pre` and was failing `generated-files`. The failure was not caused by the change — all three open `pre` PRs (#3385, #3386, #3389) failed identically.

`make prep/repos` clones mql's `main` branch **unpinned**, and mql main has moved to the unversioned module path `go.mondoo.com/mql` (v14.0.0-rc.1). `pre` still pins `go.mondoo.com/mql/v13`, so regeneration emitted `.pb.go` files importing `go.mondoo.com/mql/llx` while `go.mod` provided `go.mondoo.com/mql/v13/llx`:

```
policy/cnspec_policy.pb.go:13:2: no required module provides package go.mondoo.com/mql/llx
```

`pre` has since been bypassed: `main` is now `module go.mondoo.com/cnspec` on mql v14-rc, and still carries every deprecation marker this sweep targets. So the branch is rebased onto `main` and retargeted there, dropping the three commits already merged into `pre`.

## What

`ScoreValue.score` (field 1) is removed and its tag reserved:

```proto
message ScoreValue {
  // 1 was score, replaced by riskScore below
  reserved 1;
  reserved "score";

  string status = 2;
  uint32 riskScore = 3;
}
```

`gatherScoreValue` already dual-wrote both fields (`Score: score.Value` alongside `RiskScore: 100 - score.Value`), so nothing was ever derived from `score` alone. Dropping the assignment is a one-line change.

## ⚠️ Breaking: JSON reporter output

This is the one PR in the sweep with a user-visible output change. The `score` key disappears from `cnspec scan -o json` score values. Consumers read `riskScore` instead, which is the inverse:

```
riskScore == 100 - score
```

So a passing check that reported `"score": 100` now reports `"riskScore": 0`. This is exactly the kind of change a major exists for, but it is worth calling out in the release notes.

## Reporter testdata

`main` has `cnspec_report_compare.go` and its test, which `pre` did not. Both fixtures it loads used `"score"`, and `FromJSON` unmarshals through encoding/json, which drops unknown keys silently — a leftover `"score"` would have loaded as nothing and left every `RiskScore` at 0, leaving `TestReportComparisonWithExplicitMrn` to separate the two reports on `status` alone.

Rewriting the values as `riskScore` keeps the numeric dimension. Because `riskScore = 100 - score` and zero is `omitempty`, the pass entries drop the key entirely and the mixed fixture's failing entry gains `"riskScore": 100`.

Verified with two mutation controls:

- equalise both fixtures' `status` so **only** `riskScore` differs → the test still separates them, proving `riskScore` is read
- additionally strip `riskScore` so the fixtures are identical → the test fails, proving nothing else was propping it up

## Verification

- `make cnspec/generate` with protoc 33.6 (the version `.github/env` pins) — the reserved range and name land correctly in the raw descriptor (`J\x04\b\x01\x10\x02R\x05score`)
- `go build ./...` clean
- `go test ./cli/reporter/...` passes
- `go test ./cli/... ./policy/...` — one unrelated pre-existing failure, `TestNextAvailableFilename` in `policy/scan/pdque`, which collides on nanosecond-timestamp filenames on fast hardware and reproduces on pristine `main`

Note on regeneration: `protoc-gen-go` formats its output with the `go/format` of the toolchain that compiled the plugin. Regenerating locally under Go 1.27 reflows every doc-comment list in `policy/cnspec_policy.pb.go` (98 lines, comments only), which CI's pinned Go 1.26 does not produce. Confirmed by regenerating on a pristine `main` tree, where the identical churn appears with no changes applied, so that file is deliberately left untouched.

### DIFF
--- a/cli/reporter/cnspec_report.pb.go
+++ b/cli/reporter/cnspec_report.pb.go
@@ -138,11 +138,9 @@ func (x *ScoreValues) GetValues() map[string]*ScoreValue {
 }
 
 type ScoreValue struct {
-	state protoimpl.MessageState `protogen:"open.v1"`
-	// FIXME: DEPRECATED, remove in v13, replaced by riskScore below
-	Score         uint32 `protobuf:"varint,1,opt,name=score,proto3" json:"score,omitempty"` // ^
-	Status        string `protobuf:"bytes,2,opt,name=status,proto3" json:"status,omitempty"`
-	RiskScore     uint32 `protobuf:"varint,3,opt,name=riskScore,proto3" json:"riskScore,omitempty"`
+	state         protoimpl.MessageState `protogen:"open.v1"`
+	Status        string                 `protobuf:"bytes,2,opt,name=status,proto3" json:"status,omitempty"`
+	RiskScore     uint32                 `protobuf:"varint,3,opt,name=riskScore,proto3" json:"riskScore,omitempty"`
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache
 }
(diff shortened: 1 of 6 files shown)


arlimus:
Thank you @tas50

PR #3451, opened 2026-08-17
📝 azure: rewrite check descriptions as prose

tas50:
Rewrites all **319** check descriptions in `content/mondoo-azure-security.mql.yaml` to the description standard in `content/CLAUDE.md`. Prose only.

## What changed

| | Before | After |
|---|---|---|
| Opener | 318 of 319 began `This check ensures that <title restated>` | All 319 begin `This check` and name a **security capability**, not the resource |
| `**Why this matters**` | 319 bulleted lists | 319 prose paragraph sets, every bullet's substance kept |
| `**Risk mitigation**` | 319 occurrences | 0. Folded into the Why prose or the closing paragraph, never discarded |
| Admin surface | often unstated | Azure portal blade, `az` CLI command/flag, or ARM/Bicep property named |
| Bullet lines | 319 descriptions | 0 |
| Terminology | "Azure AD", "Azure Security Center" | Microsoft Entra ID, Microsoft Defender for Cloud, Azure AI services |
| Median length | 211 words | 271 words |

Structure conversion plus sharpening, not replacement. This policy's content was already good, so the goal was to keep every concrete fact while changing the shape.

## Example: specifics preserved

`mondoo-azure-security-compute-data-disks-encrypted`, closing paragraphs of the new text:

> With a Customer-Managed Key (CMK), those contents are unreadable to anyone who cannot reach your key vault, Microsoft personnel included [...] Rotating or revoking the key cuts off access to the encrypted data immediately, without deleting the disk or migrating the workload, and Key Vault access logs record every encryption and decryption operation tied to a specific identity [...]
>
> Changing the encryption type on an attached disk requires deallocating the virtual machine, so schedule a maintenance window.

and its opening names the exact administration surface, four ways:

> only a disk pointed at a disk encryption set carries an encryption type of `EncryptionAtRestWithCustomerKey` [...] In the portal the setting is the Encryption section of the disk [...] From the CLI the change is `az disk update --encryption-type EncryptionAtRestWithCustomerKey --disk-encryption-set` [...] In Terraform the argument is `disk_encryption_set_id` on `azurerm_managed_disk`; in Bicep it is `encryption.diskEncryptionSetId`.

Literals carried through across the policy include `allowBlobPublicAccess`, `supportsHttpsTrafficOnly`, `allowSharedKeyAccess`, `requireInfrastructureEncryption`, `enablePurgeProtection`, `enableRbacAuthorization`, `Key Vault Crypto Service Encryption User`, `Storage Blob Data Contributor`, `Microsoft.Sql/servers/databases/auditingSettings`, `require_secure_transport`, `AppGwSslPolicy20220101S`, `Microsoft_DefaultRuleSet`, `WAF_v2`, ports `22`/`3389`/`1433`/`5900-5903`, `0.0.0.0/0`, `DHGroup14`/`PFSGroup14`/`GCMAES256`, FIPS 140-2 Level 3, and the 7-to-90-day soft-delete window.

## Substance preservation

Automated diff of the set of backticked literals and the set of numeric thresholds between the description at `origin/main` and the new one, for all 319 checks.

**Total lost: 14**, all verified equivalents or a deliberate correction. Every one justified individually:

| Lost token | Checks | Justification |
|---|---|---|
| `` `ON` `` | 4 PostgreSQL parameter checks | New text uses `` `on` ``, matching the `az postgres flexible-server parameter set --value on` form. Case only. |
| `` `state: Enabled` `` | 2 SQL auditing checks | New text states the policy state must read `` `Enabled` `` in prose; the ARM key/value spelling is the only thing dropped. |
| `` `"1.2"` ``, `` `site_config.minimum_tls_version` `` | `function-app-min-tls-1-2` | New text has `` `minimum_tls_version` `` inside the `` `site_config` `` block plus `--min-tls-version 1.2`. Quoting and path-joining only. |
| `` `encryption.key_vault_key_identifier` `` | `appconfiguration-cmk-encryption` | New text has an `` `encryption` `` block carrying `` `key_vault_key_identifier` ``. Path-joining only. |
| `` `myacr.azurecr.io/app@sha256:abcdef...` `` | `container-app-image-digest-pinned` | New text has the same example without the trailing ellipsis. |
| `30` | `expiration-date-...-kv` | Illustrative "a key with 30 days remaining"; new text keeps the idea as "bounds that exploitation window to whatever remains on the clock". |
| `3`, `7` | `servicebus-namespace-public-network-access-disabled` | Detector artifact from "Layer-7" / "layer-3". The new text keeps the full defense-in-depth argument in prose. |
| `0` | `frontdoor-origin-https` | **Deliberate correction.** The old text claimed setting the origin HTTP port to `0` prevents fallback. The Front Door origin API constrains `httpPort` to 1-65535, so that is not possible; the new text says so and explains the control is asserted on the route's forwarding protocol instead. |

Twelve further specifics initially dropped were restored before commit rather than justified: `blobAuditingPolicy`, `Microsoft.Sql/servers/databases/auditingSettings`, `PFSGroup14`, `azurerm_virtual_network_gateway_connection`, `ike_encryption`, `ipsec_encryption`, `azurerm_log_analytics_linked_service`, `storage_account_name`, `azurerm_storage_account.customer_managed_key`, `credentialsType`, `...Protocols.Tls11`, and `"False"`.

## Validation

- `cnspec policy lint content/mondoo-azure-security.mql.yaml` → **valid policy bundle(s)**
- `typos content/mondoo-azure-security.mql.yaml` → **no output, exit 0**. No allowlist was touched.
- `go test ./internal/bundle/...` → **ok**
- Python gate over all 319 checks: **0** failing `.startswith('This check')`, **0** missing `**Why this matters**`, **0** em/en dashes or ` -- `, **0** `**Risk mitigation**`, **0** MQL accessor paths, **0** bullet lines, **0** byte-identical siblings, **0** stale "Azure AD" / "Azure Security Center" references.
- Structural diff: bundle parsed at `origin/main` and at this tip, every `docs.desc` and `policies[].version` replaced with a placeholder, trees compared → **identical**.
- Raw line diff confirms the only non-description line changed in the whole file is `version: 10.1.0` → `10.1.1`.

## No behavior change

**No `mql`, `filters`, `tags`, `impact`, `props`, `title`, `audit`, or `remediation` changed.** Only `docs.desc`, plus the one PATCH version bump. Nothing in this PR alters what any check evaluates or how any asset scores.

## Check defects found but NOT fixed

Reported for a follow-up; out of scope here. The greedy dotted-path trap was actively scanned for across the whole policy and **no instance was found** (three AKS checks reach sub-objects through the `aks` field alias, which avoids it).

- `ensure-that-notify-about-alerts-with-high-severity-is-on` uses `.all()` over security contacts, so a subscription with **zero** contacts passes. Its two siblings correctly use `.any()`.
- The 14 Defender-plan Terraform forms filter on "some pricing resource exists" then filter to the specific plan, so a config enabling only one plan passes every other plan's check vacuously.
- `datafactory-public-access-disabled` plan/state forms read `public_network_access_enabled`, which `azurerm_data_factory` does not have; the HCL form correctly reads `public_network_enabled`. Plan/state fail every asset.
- `ensure-that-ssl-enabled-latest-version-mysql` reads `sslEnforcement`, the retired Single Server property, against a Flexible Server.
- `sql-server-audit-on` and `sql-server-devops-auditing-enabled` Terraform forms are tautologies: the filter requires the resource type and the query only asserts it is present.
- `aks-rbac-enabled` and `aks-private-cluster-public-fqdn-disabled` HCL forms use `== true` / `== false` where the provider default already matches, so an omitted attribute reads `null` and false-fails.
- `log-analytics-cmk-for-query-bicep` reads `properties["forceCmkForQuery"]`; ARM puts it under `properties.features`, which its own remediation patches.
- `appgw-no-weak-ciphers`, `batch-encryption-configured`, `servicebus-namespace-cmk-encryption` and `synapse-cmk-encryption` have IaC forms that assert only block presence, not the property the title names.
- `redis-latest-version` Bicep form compares a string property to the integer `6`.
- `storage-min-tls-version`, `sql-server-min-tls-12` and `sql-managed-instance-min-tls-1-2` test equality with `1.2`, so a stricter floor is reported as failing.
- Several private-endpoint and diagnostic-setting IaC forms match any such resource anywhere in the file without tying it to the resource under evaluation.


tas50:
Check defects found while reading the Azure queries for this rewrite are now filed as **#3453**, six items verified against the query text and against `terraform providers schema -json` for azurerm 3.117.1 and 4.x. Two Terraform forms fail every correctly configured asset; one can never fail.

chris-rock (approved):

Issue #3450, opened 2026-08-17
mondoo-gcp-security: four GKE checks read through resource-name collisions, three pass vacuously and one fails on every cluster


tas50:
Surfaced during the description rewrite for this policy. Four GKE checks build their assertion on a dotted path that is itself a resource name, which is the failure mode `content/CLAUDE.md` documents under "A dotted path that is also a resource name is not a field read".

That document already names `gcp.project.gkeService.cluster.networkPolicy.*` as an example of the shape. These are live instances of it.

## The evidence

Every intermediate path in these four checks resolves as a resource, not as a field read:

```
gcp.project.gkeService.cluster.ipAllocationPolicy       IS A RESOURCE: True
gcp.project.gkeService.cluster.networkPolicy            IS A RESOURCE: True
gcp.project.gkeService.cluster.addonsConfig             IS A RESOURCE: True
gcp.project.gkeService.cluster.securityPostureConfig    IS A RESOURCE: True
```

Read from the installed provider schema. The cluster resource declares each as a field whose type is the same-named resource:

```
field ipAllocationPolicy      type='gcp.project.gkeService.cluster.ipAllocationPolicy'
field networkPolicy           type='gcp.project.gkeService.cluster.networkPolicy'
field addonsConfig            type='gcp.project.gkeService.cluster.addonsConfig'
field securityPostureConfig   type='gcp.project.gkeService.cluster.securityPostureConfig'
```

The compiler extends the resource path greedily, so the longest matching resource name wins. Each query builds a bare sub-resource with no id, the cluster's accessor never runs, and every field beneath it reads `null`.

## What each check then answers

The direction of the wrong answer depends on the comparison, and three of the four fail open.

**Fails on every cluster:**

```mql
# gke-ip-aliasing-enabled
gcp.project.gkeService.cluster.ipAllocationPolicy.useIpAliases == true
```
`null == true` is false. Every cluster fails regardless of configuration.

**Passes on every cluster:**

```mql
# gke-cluster-network-policy-enabled
gcp.project.gkeService.cluster.addonsConfig.networkPolicyConfig['disabled'] != true
```
`null != true` is true.

```mql
# gke-security-posture-enabled
gcp.project.gkeService.cluster.securityPostureConfig.mode != "DISABLED"
gcp.project.gkeService.cluster.securityPostureConfig.vulnerabilityMode != "VULNERABILITY_DISABLED"
```
Both `null != "…"` comparisons are true.

```mql
[...]

PR #3452, opened 2026-08-17
📝 gcp: rewrite check descriptions as prose

tas50:
Rewrites all 297 check descriptions in `content/mondoo-gcp-security.mql.yaml` to the
description standard in `content/CLAUDE.md`.

This is a **structural conversion plus sharpening, not a replacement**. The existing GCP
content was already good: it named real IAM permissions, real API fields, real service
defaults. All of that was preserved and folded into prose.

## What changed

Every one of the 297 descriptions previously carried the same generated skeleton: a
resource-first opener, a bulleted `**Why this matters**` list of benefit labels, and a
bulleted `**Risk mitigation**` section that mostly restated the remediation. All 297 had
both bulleted sections.

| | Before | After |
|---|---|---|
| Opener | Names the GCP resource, paraphrases the title | Begins `This check`, names the security **capability** and its required state |
| Why this matters | Bulleted `- **Label:** text` benefit list | Two to four prose paragraphs |
| Risk mitigation | Separate bulleted heading | Heading removed, substance folded into the prose and the closing paragraph |
| Where to administer it | Often absent | Named: console page, `gcloud` command, or API field |
| Closing | Restated the opener | Says when the control does not apply, or what to pair it with |
| Bulleted lines | 297 checks had them | 0 |

## Example: specifics preserved and sharpened

`compute-instances-block-project-wide-ssh-keys`

**Before** opened with the resource and listed benefit labels:

> This check ensures that Compute Engine VM instances are configured to block project-wide SSH keys... **Blast radius reduction:** ... **Principle of least privilege:** ... **Key revocation precision:** ...

**After** opens with the capability, and **adds** the metadata key, its required value, the
console location and the exact `gcloud` command, none of which the old text had:

> This check verifies that a VM instance accepts SSH keys only from its own instance metadata.
>
> Compute Engine trusts two sources of SSH public keys... Setting the `block-project-ssh-keys` metadata key to `true` removes the first source... Administrators set it from the `SSH Keys` section of the instance edit page in the Google Cloud console, or with `gcloud compute instances add-metadata INSTANCE_NAME --zone=ZONE --metadata block-project-ssh-keys=true`.

Every bullet's substance survives in the prose: blast radius, least privilege and revocation
precision all appear as explanation rather than as labels.

## Substance preservation

A mechanical gate diffs the set of backticked literals and numeric thresholds in the OLD text
at `origin/main` against the NEW text, per check.

**Total specifics lost: 4, across 2 checks of 297.** An initial pass lost 67; each was routed
back and restored. All 4 remaining losses are deliberate and fall into the two categories the
description standard forbids in prose:

| Check | Literal | Why dropped |
|---|---|---|
| `gke-network-policy-provider-specified` | `enabled == true` | Raw MQL fragment, not a user-facing field |
| `gke-network-policy-provider-specified` | `networkPolicy.enabled == true` | Raw MQL fragment |
| `gke-network-policy-provider-specified` | `networkPolicy.provider == "PROVIDER_UNSPECIFIED"` | Raw MQL fragment; the `PROVIDER_UNSPECIFIED` value itself is retained in the prose |
| `appengine-handler-secure-always` | `appengine-iap-or-vpc-only-ingress` | A check UID; house style forbids UIDs in prose, so it is now referred to as a companion check |

## No behavior changed

Only `docs.desc` and the policy `version` were touched. **No `mql`, `filters`, `tags`,
`impact`, `props`, `title`, `audit`, or `remediation` changed.** This is proven mechanically,
not asserted: the bundle is parsed at `origin/main` and at the tip, every `docs.desc` and
`policies[].version` is replaced with a placeholder, and the resulting trees are compared.

```
STRUCTURAL DIFF: trees identical apart from docs.desc and policy version
```

Policy version: `9.11.0` -> `9.11.1`.

## Validation

| Gate | Result |
|---|---|
| `cnspec policy lint content/mondoo-gcp-security.mql.yaml` | valid policy bundle(s) |
| `typos` | clean, no allowlist edited |
| `go test ./internal/bundle/...` | ok |
| Openers not starting `This check` | 0 of 297 |
| Missing `**Why this matters**` | 0 of 297 |
| `—`, `–`, ` -- ` | 0 |
| `**Risk mitigation**` remaining | 0 |
| MQL accessor paths in prose | 0 |
| Bulleted lines | 0 |
| Byte-identical sibling descriptions | 0 |
| Structural diff vs `origin/main` | trees identical apart from `docs.desc` and version |
| Specifics lost | 4 of 297 checks, all justified above |

## Check defects found but deliberately NOT fixed

These are pre-existing query bugs surfaced while reading the MQL. None are touched here,
since this PR changes prose only. Each is worth a follow-up.

- **`model-armor-blocking-mode`** asserts `enforcementType != "INSPECT_ONLY"`. Because
  `protoToDict` omits defaults, a template with no `templateMetadata` reads `null`, and
  `null != "INSPECT_ONLY"` is true, so it passes. The check proves "not explicitly
  inspect-only", never "blocking". The description was written to that honest statement.
- **`model-armor-rai-confidence`** has no null guard on the live query while all three
  Terraform forms guard with `== empty ||`. An absent key is `null`, and `null.all()` fails
  outright rather than passing vacuously, so a template with no RAI filters fails here. Parity
  defect against its own siblings.
- **`model-armor-sdp-enabled`** tests `sdpSettings != empty`, which proves the key is present,
  not that enforcement is `ENABLED`.
- **`vertexai-index-endpoint-not-public`** asserts `publicEndpointEnabled == false`, but its
  own audit text says a passing endpoint may report the field absent. `null == false` is
  false, so a genuinely private endpoint can fail. The Terraform forms use the null-safe
  `!= true`; the live query is the odd one out.
- **`vertexai-endpoint-private-service-connect`** accepts either a `network` value or a PSC
  config, so a VPC-peered endpoint with no PSC config satisfies a check titled "use Private
  Service Connect".
- **`vertex-ai-job-cmek-encryption`** covers metadata stores and Feature Store in its title and
  text, but the live query only evaluates tensorboards, custom jobs, indexes and index
  endpoints. Noted in the description's closing paragraph.
- **`vpc-sc-access-levels-defined`** passes when any one access policy defines a level, though
  the title reads as an organization-wide assertion.
- **`workflows-no-default-service-account`** depends on the provider resolving the default
  service account; if the field were ever absent, `null` satisfies both comparisons and the
  workflow passes.
- **Duplicate coverage:** `vertex-ai-job-cmek-encryption` and
  `vertexai-additional-resources-cmek-encryption` both assert CMEK on indexes, index endpoints
  and tensorboards on the same platform. `vertexai-custom-job-has-explicit-service-account` is
  a strict subset of `vertexai-custom-job-no-default-service-account`.
- **Terraform-only coverage:** `vertex-ai-workbench-no-public-ip` and
  `workstations-cluster-private-cluster` have no live GCP form, so no live asset is evaluated.


tas50:
Check defects found while reading the GCP queries for this rewrite are now filed as **#3454** (six items, including `compute-route-no-internet-gateway-to-sensitive-subnets` failing every project and two checks that can never fail). The four GKE resource-name collisions remain separate in **#3450**.

chris-rock (approved):

PR #3455, opened 2026-08-17
📝 content: rewrite AI, EDR and MCP check descriptions

tas50:
47 descriptions across three cross-platform agent policies, rewritten to the standard in `content/CLAUDE.md` (#3393).

| Policy | Checks |
|---|---|
| `mondoo-ai-security` | 38 |
| `mondoo-mcp-security` | 7 |
| `mondoo-edr-policy` | 2 |

## What changed

The opener now names the capability and its required state, the setting is named where the reader administers it rather than where the scanner reads it, and `**Why this matters**` is prose describing what an attacker does without the control. The `**Risk mitigation**` sections, which restated `docs.remediation`, are gone.

`mondoo-ai-security` had the heaviest duplication of the three: many descriptions differed only in the product name while describing the same generic risk.

## Verification

- `cnspec policy lint` on all three bundles — valid
- `typos` — clean
- `go test ./internal/bundle/...` — pass
- Structural diff against `origin/main` — trees identical apart from `docs.desc`

Descriptions only. No MQL, impact, or compliance tag is touched.

### DIFF
--- a/content/mondoo-ai-security.mql.yaml
+++ b/content/mondoo-ai-security.mql.yaml
@@ -4,7 +4,7 @@
 policies:
   - uid: mondoo-ai-security
     name: Mondoo AI Security
-    version: 2.1.0
+    version: 2.1.1
     license: BUSL-1.1
     tags:
       mondoo.com/category: security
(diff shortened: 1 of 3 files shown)


tas50:
Closing as a duplicate. This is the same content already merged in #3441; the branch was never pushed at the time and I re-pushed it against a stale local main. No change here that is not already on main.

PR #3459, opened 2026-08-17
📝 github: rewrite check descriptions as prose

tas50:
## What changed

Rewrites `docs.desc` for all 8 checks in `mondoo-github-best-practices` into the house prose format. Queries, filters, titles, impacts, audit sections, and remediation blocks are untouched, and the policy version is not bumped.

Each description now:

- opens with the literal `This check verifies that ...` and leads with the capability rather than the field being read;
- names the artifact the reader creates with its conventional filename and location, `CODE_OF_CONDUCT.md`, `SUPPORT.md`, `LICENSE`, `README.md`, `.gitignore`, and says where GitHub surfaces it, the "About" panel, the Community Standards page, the new issue page;
- replaces the bulleted bold labels with prose that says what actually goes wrong.

## Best-practices framing

This policy is `mondoo.com/category: best-practices`, so the descriptions do not invent an attacker. They name the concrete failure instead: someone scanning a few hundred repositories sees a column of names and gives up; an engineer whose dependency scanner reports the license as unknown picks a different library rather than asking legal; an on-call engineer reads unfamiliar code with no team alias to page; support questions land in the same queue as real defects, so genuine bugs wait behind them.

Two real security consequences do exist here, and both are stated plainly rather than reached for. A `.env` file or a downloaded service account key swept up by a wildcard add stays in the history after it is deleted and has to be rotated. And a reporter who finds neither a support contact nor a `SECURITY.md` is left choosing between saying nothing and opening a public issue that describes the flaw to everyone.

Median description length goes from **75 words to 302**.

## Gates

| Gate | Result |
| --- | --- |
| `cnspec policy lint ./content/mondoo-github-best-practices.mql.yaml` | `valid policy bundle(s)` |
| `typos content/mondoo-github-best-practices.mql.yaml` | silent |
| substance gate | **0** specifics lost across 0 of 8 checks |

The substance gate reports zero losses, so there are no waivers to justify: every backticked literal in the old text survives into the new text.

### DIFF
--- a/content/mondoo-github-best-practices.mql.yaml
+++ b/content/mondoo-github-best-practices.mql.yaml
@@ -59,13 +59,17 @@ queries:
       }
     docs:
       desc: |
-        Repositories should include a SUPPORT.md file to let people know how to get help with the project.
+        This check verifies that the repository publishes support resources.
+
+        Support resources go in a `SUPPORT.md` file in the repository root or in the `.github` directory, and an organization can publish one shared copy in its own `.github` repository that covers every repository it owns. The file says where to get help: links to the documentation, the issue tracker, any community forum or chat channel, and a contact address for the maintainers. GitHub reads it and shows a link to your support resources on the new issue page, so the person about to file sees it while they are still deciding what to do. A copy stored under `docs` is rendered by GitHub but is not inspected by this check, so keep the file in the root or in `.github`.
 
         **Why this matters**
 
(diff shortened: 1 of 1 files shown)


tas50:
Check defects found while reading these queries are now filed as **#3461**. The headline one: `code-of-conduct` and `support-resources` are inert on every user-owned repository because `github.organization` 404s and takes the whole query with it, so the `else` fallback never runs. Verified by live scan against a user repo that complies, a user repo that does not, and an org repo as control.

PR #3460, opened 2026-08-17
📝 mongodbatlas: rewrite check descriptions as prose

tas50:
## What changed

Rewrites `docs.desc` for all **40 checks** in `mondoo-mongodbatlas-security` in the house prose format. No other field is touched and the policy version is not bumped.

The old descriptions carried the full generated skeleton:

```
**Why this matters**

- **Credential theft is common:** passwords alone are exposed to phishing...
- **Broad blast radius:** organization members can reach production data...

**Risk mitigation**

- **Require MFA:** enable the organization setting so no member can sign in...
```

Bold labels tell a reader the *category* of a concern without ever saying what happens, and `**Risk mitigation**` restated `docs.remediation` one section above where the remediation already lives. Both are gone. Every label became the sentence it was standing in for.

## Shape of the new text

Each description opens by naming what is verified, says where the reader administers the setting in the Atlas console and what the Atlas Administration API calls it, gives two to four prose paragraphs on what an attacker concretely does without the control, and closes on when the check legitimately should not be applied or what to pair it with.

Atlas vocabulary is used throughout: Organization Settings, Access Manager, Database Deployments and Edit Configuration, Security and Database Access, Security and Network Access, Security and Advanced. Organization-level and project-level settings are named as such rather than flattened into "the organization", since the distinction is load-bearing for the federation, service account, and encryption checks.

The attacker paths are named as steps rather than as outcomes. For the MFA check that is: sign in at cloud.mongodb.com, open Network Access and add an address, open Database Access and create a user with broad roles, connect with an ordinary driver. For the broad-CIDR check it is: compromise any host inside the allowed range and meet only database authentication, with no need to be on the application host at all.

## Median word count

| | before | after |
|---|---|---|
| median `docs.desc` | 182.5 | 287 |

## Gates

| gate | result |
|---|---|
| `cnspec policy lint ./content/mondoo-mongodbatlas-security.mql.yaml` | `valid policy bundle(s)` |
| `typos content/mondoo-mongodbatlas-security.mql.yaml` | silent |
| substance gate | **0 specifics lost** across 0 of 40 checks |

Every backticked literal and numeric threshold in the old text survives, including `minimumEnabledTlsProtocol`, `encryptionAtRestProvider`/`NONE`, `backupEnabled`, `pitEnabled`, `terminationProtectionEnabled`, `redactClientLogData`, `$external`, `atlasAdmin`, `deleteAfterDate`, `DROP_DATABASE`/`DROP_COLLECTION`/`SHUTDOWN`, `readWriteAnyDatabase`/`dbAdminAnyDatabase`/`root`, `enabledForSearchNodes`, `auditAuthorizationSuccess`, and the thresholds 90 days, 168 hours, TLS 1.0 through 1.3, MongoDB 7.0, and the /8, /16, /24, /32 and 256-address CIDR values.

### DIFF
--- a/content/mondoo-mongodbatlas-security.mql.yaml
+++ b/content/mondoo-mongodbatlas-security.mql.yaml
@@ -121,18 +121,17 @@ queries:
       mongodbatlas.multiFactorAuthRequired == true
     docs:
       desc: |
-        Atlas can require every member of the organization to authenticate with a second factor. When multi-factor authentication is not required, an attacker who steals or guesses a single password can sign in and reach every project, cluster, and database in the organization.
+        This check verifies that every member of the Atlas organization must present a second authentication factor before they can sign in.
+
+        MongoDB Atlas exposes this as "Require Multi-Factor Authentication" in Organization Settings, reachable from the organization menu at cloud.mongodb.com. The Atlas Administration API reports the same control as `multiFactorAuthRequired` on the organization settings resource, and this check requires it to be enabled. The requirement is set once for the organization and applies to every member regardless of which projects they belong to.
 
         **Why this matters**
 
-        - **Credential theft is common:** passwords alone are exposed to phishing, reuse across sites, and breach dumps.
-        - **Broad blast radius:** organization members can reach production data and configuration across all projects.
-        - **Low-friction control:** MFA is a built-in setting that stops most automated account takeover attempts.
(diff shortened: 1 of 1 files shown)


tas50:
Check defects found while reading these queries are now filed as **#3463**. Headline: `project-cluster-supported-version` compares `mongoDBMajorVersion` as a string, so it inverts at MongoDB 10.0, and `project-cluster-encryption-at-rest` uses `!= "NONE"`, which passes when the field is absent.

PR #3462, opened 2026-08-17
📝 databricks: rewrite check descriptions as prose

tas50:
## What this changes

Rewrites `docs.desc` for all 25 checks in `content/mondoo-databricks-security.mql.yaml` into the house prose format. No query, filter, tag, impact, audit, or remediation text is touched, and the policy version stays at 1.1.0.

The old descriptions carried the full generated skeleton: a `**Why this matters**` section made of bulleted bold labels, and a `**Risk mitigation**` section that restated `docs.remediation`. A bold label tells the reader the *category* of a concern without ever saying what happens. Each one is now the sentence it was standing in for. This is a `security` policy, so every description names what an attacker concretely does rather than gesturing at "unauthorized access".

## Naming the right administrative surface

Databricks splits administration across three surfaces the old text tended to blur together, so each description now says which one the reader opens:

- **Account console** for private access settings, customer-managed keys, and service principals.
- **Workspace admin console** for enhanced security monitoring, the compliance security profile, automatic cluster update, IP access lists, legacy access, token settings, the init-script flags, and notebook result storage.
- **Cluster policies** for the per-cluster attributes, named alongside the individual attribute so it is clear that setting it on one cluster is not the control.

Where `docs.audit` or `docs.remediation` already carried the exact `databricks` CLI call or API field, the description uses that rather than an invented path.

## Numbers

Median word count: **161 before, 285 after**. Range 265 to 312; none identical.

## Gates

Run from the worktree unless noted:

| Gate | Result |
|---|---|
| `cnspec policy lint ./content/mondoo-databricks-security.mql.yaml` | `valid policy bundle(s)` |
| `typos content/mondoo-databricks-security.mql.yaml` | silent |
| `go test -count=1 ./internal/bundle/...` | pass |
| Structural diff vs `origin/main` | `trees identical apart from docs.desc and policy version` (version is in fact unchanged) |
| Substance gate | `TOTAL SPECIFICS LOST: 0 across 0 checks (of 25)` |

Every backticked literal and numeric threshold in the old text survives, including `RESTRICT_TOKENS_AND_JOB_RUN_AS` and `ALLOW_ALL`, the `LEGACY_*` and `DATA_SECURITY_MODE_*` access modes, the `{{secrets/scope/key}}` reference form, `0.0.0.0/0` and `::/0`, the `ENDPOINT` and `ACCOUNT` private access levels, the `users` secret-scope principal, and the 90-day and 730-day token lifetimes.

### DIFF
--- a/content/mondoo-databricks-security.mql.yaml
+++ b/content/mondoo-databricks-security.mql.yaml
@@ -95,18 +95,17 @@ queries:
       databricks.workspaceConf.enhancedSecurityMonitoringEnabled == true
     docs:
       desc: |
-        This check ensures that Enhanced Security Monitoring is turned on for the workspace so that compute nodes run a hardened operating system image and carry additional antivirus and behavior based monitoring agents. By default the setting is disabled, which leaves cluster hosts running a standard image with no host level threat detection.
+        This check verifies that the compute nodes behind a workspace boot a hardened operating system image and run antivirus and behavior based monitoring agents.
+
+        Enhanced security monitoring is a workspace admin setting, not an account console setting. An administrator finds it under Settings on the **Security and compliance** tab, and the Databricks CLI reads the same value through `databricks settings enhanced-security-monitoring get`, which reports `is_enabled`. This check requires that flag to be true. The setting is off by default, so a workspace that no one has hardened runs its clusters on a stock image with no host level detection at all.
 
         **Why this matters**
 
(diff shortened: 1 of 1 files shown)


tas50:
Check defects found while reading these queries are now filed as **#3464**. Headline: `delta-sharing-recipient-ip-restricted` is filtered to the account asset but its resource requires a workspace connection, so no asset can ever score it.

PR #3465, opened 2026-08-17
📝 dockerfile: rewrite check descriptions as prose

tas50:
Rewrites `docs.desc` for all 41 checks in the two Dockerfile bundles into the house prose format. Nothing else in either bundle changes.

## What changed

Every description now opens with `This check verifies that ...`, follows with a paragraph naming the Dockerfile instruction the reader edits and the form it has to take, and replaces the bolded bullet labels under `**Why this matters**` with prose that says what actually happens. The old bullets named a *category* of concern (`**Unauthorized access:**`, `**Image bloat:**`, `**Compliance risks:**`) without ever saying what goes wrong; each one is now the sentence it was standing in for.

The two bundles are deliberately written differently.

**`mondoo-dockerfile-security`** (`mondoo.com/category: security`) names the attacker step. Port 2375 is the unauthenticated Docker Remote API, so the description says the attacker creates a container that bind-mounts the host root and writes to `/etc/shadow`, which is root on the node rather than root in one container. `--network=host` reaches the link-local metadata endpoint at `169.254.169.254` and trades a build foothold for the runner's registry push credential. A missing `USER` means an attacker who lands through a deserialization bug holds the default capability set and rewrites the entrypoint to survive a restart. A repointed `latest` tag substitutes the base image with no diff and no review.

**`mondoo-dockerfile-best-practices`** (`mondoo.com/category: best-practices`) does not invent a threat, because most of these checks do not have one. `use-apt-get` is about `apt` having no stable scripted interface and printing a warning on every invocation that trains readers to ignore warnings, not about an attacker. The consequences named are engineering ones and are as concrete as the security ones: an install resolving against a package index cached in a layer built months ago, a `rm -rf` in a later instruction that records a deletion marker and reclaims no bytes because layers are append-only, an `apt-get install` without `-y` that holds a CI runner until the job timeout fires, a relative `WORKDIR` that silently relocates every path in the file when the base image tag changes, a multi-source `COPY` whose error message sends the reader looking at whether the directory exists rather than at the missing slash.

Instruction names stay in backticks throughout, because the thing being administered here is the Dockerfile instruction: `EXPOSE`, `RUN`, `USER`, `VOLUME`, `WORKDIR`, `COPY`, `ADD`, `ENV`, `HEALTHCHECK`, `FROM`, and the flags and paths that go with them.

## Size

| Bundle | Checks | Median words before | after |
|---|---|---|---|
| `mondoo-dockerfile-security` | 20 | 116 | 293 |
| `mondoo-dockerfile-best-practices` | 21 | 100 | 290 |

No description is shorter than 270 words or longer than 341.

## Not changed

No policy version bump, and no change to any query, `filters`, `impact`, `tags`, `docs.audit`, `docs.remediation`, or `refs` in either file. The structural diff below is the proof.

## Gates

Run against both files:

- `cnspec policy lint ./content/mondoo-dockerfile-security.mql.yaml` -> `valid policy bundle(s)`
- `cnspec policy lint ./content/mondoo-dockerfile-best-practices.mql.yaml` -> `valid policy bundle(s)`
- `typos` on both files -> silent
- `go test -count=1 ./internal/bundle/...` -> `ok`
- Structural diff against `origin/main`, run once per file -> `trees identical apart from docs.desc and policy version` for both
- Substance gate, `mondoo-dockerfile-best-practices` -> **0 specifics lost across 21 checks**
- Substance gate, `mondoo-dockerfile-security` -> **4 specifics lost across 1 check of 20**

All four losses are on `mondoo-docker-security-image-authors-label`, and all four are fragments of one compliance framework citation the old prose carried: `800`, `171`, `3.4`, and `1`, from a NIST 800-171 control reference. Framework citations are carried by the `compliance/*` tags rather than by description prose, and the fact the sentence was making, that an asset inventory has to record who owns a component, survives in words. Every backticked literal in both bundles survives, including the full sensitive-path lists, `--allow-insecure-repositories`, `-fsSLk`, `/var/run/docker.sock`, `--mount=type=cache`, `169.254.169.254`, and the port numbers 22, 2375, 8500, 6443, 1, and 65535.

### DIFF
--- a/content/mondoo-dockerfile-best-practices.mql.yaml
+++ b/content/mondoo-dockerfile-best-practices.mql.yaml
@@ -72,13 +72,19 @@ queries:
       docker.file.stages.all(run.all(commands.none(binary == "apt" && subcommand.in(["install", "update", "upgrade", "remove", "purge"]))))
     docs:
       desc: |
-        Dockerfile `RUN` instructions should invoke the `apt-get` CLI rather than `apt` for package management operations.
+        This check verifies that package operations in `RUN` instructions call `apt-get` rather than the bare `apt` command.
+
+        Debian ships two front ends over the same package system. `apt-get` is the scripting interface, with a command-line surface and exit behavior that Debian keeps stable across releases. `apt` is the interactive one, built for a person at a terminal, and its own maintainers state that it has no stable interface for scripts. The check requires that no `RUN` instruction call `apt` with install, update, upgrade, remove, or purge.
 
         **Why this matters**
 
-        - **Predictable behavior:** The `apt-get` CLI is designed for scripting and automation, offering consistent and reliable behavior in non-interactive Dockerfile contexts.
-        - **Stability:** `apt-get` maintains a stable command-line interface across releases, while `apt` is intended for interactive use and its output and behavior may change between versions.
-        - **Reduced risk of errors:** Using `apt-get` minimizes the chance of unexpected prompts or behavior changes interrupting the build process.
+        The instability is not theoretical. `apt` writes a progress bar and colored output that assume a terminal, so a build log captured to a file fills with control sequences instead of readable text. It prints a warning on every invocation saying it should not be scripted against, which trains everyone reading build output to ignore warnings there. Flags and defaults have changed between Debian and Ubuntu releases in ways `apt-get` did not, so a Dockerfile that works on one base image tag can stop working on the next.
+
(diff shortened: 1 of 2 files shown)


tas50:
Check defects found while reading these queries are now filed as **#3471**. Headline: the parser sets `binary` to the literal token, so `RUN /usr/bin/sudo apk add` parses as `binary: "/usr/bin/sudo"` and passes the impact-100 sudo check, while the wrapped `apk add` is invisible to the package-manager checks too. Verified by running each query against probe Dockerfiles.

PR #3466, opened 2026-08-17
📝 kubernetes: rewrite check descriptions as prose

tas50:
Rewrites `docs.desc` for all 49 checks in `mondoo-kubernetes-best-practices` into the house prose format, matching the sibling PRs in this campaign.

## What changed

Each description now opens by naming the capability being verified, follows with a paragraph saying what that capability does and which manifest field the reader edits to get it, and then gives a `**Why this matters**` section of prose. The bulleted bold labels the old text used are gone; each one has been turned into the sentence it was standing in for.

This is a `best-practices` policy, so rule 4 is applied in its engineering form. The descriptions say what concretely breaks, not what an attacker does:

- A Pod with no `resources.requests.memory` is packed onto a node whose memory is already promised elsewhere, is classified BestEffort, and is the first thing the kubelet evicts when that node runs short. If memory disappears faster than the kubelet reacts, the kernel out-of-memory killer takes whichever process it scores worst, which may belong to an unrelated workload on the same node.
- A Deployment with no `readinessProbe` has every new Pod counted as ready the instant it starts, so `maxUnavailable` and `maxSurge` are enforced against a signal that means nothing and a rollout can retire the entire previous generation while none of the replacements can serve.
- A DaemonSet that binds a `hostPort` claims that number in the port space of every node in the cluster, permanently and for every node added later, and any other workload that wants it becomes unschedulable with no event pointing at the cause.
- A `hostAliases` entry is written once at Pod creation and never refreshed, so when the target moves, cluster DNS follows and the pinned copy does not. If the address has since been reassigned, requests are delivered successfully to whatever now answers there.

Where a check does carry a real security consequence it is stated plainly rather than reached for: the control plane taint description says that etcd holds every Secret in the cluster and that a workload sharing its node is a great deal closer to it.

## Per-kind scoping

The policy scopes each group to one workload kind through `filters:` on `asset.platform`, so no description says "pods" when the check only looks at CronJobs. Each is written against the kind it actually covers:

- **CronJob** — schedules falling behind, `concurrencyPolicy`, `successfulJobsHistoryLimit` leftovers.
- **Job** — `parallelism` workers competing for one host port, `backoffLimit` burned on a placement problem rather than a code fault.
- **StatefulSet** — ordinals, `volumeClaimTemplates` that make a namespace move a data migration, ordinal-by-ordinal rollouts, the governing headless Service that already provides what `hostAliases` fakes.
- **DaemonSet** — one Pod per node with no second replica to fall back on, a rollout that walks the whole fleet at full speed without a readiness signal, requests multiplied by the node count.
- **Deployment** — `maxSurge` during a rollout, `HorizontalPodAutoscaler` producing Pods that cannot be scheduled.
- **ReplicaSet** — a template a parent Deployment overwrites on the next reconcile, and a controller with no rollout mechanism of its own.
- **Pod** — the bare-Pod cases, plus the note that Job-owned Pods are out of scope for the probe checks.

Median description length goes from **151 words** to **295**.

## Gates

Run from the worktree:

| Gate | Result |
|---|---|
| `cnspec policy lint ./content/mondoo-kubernetes-best-practices.mql.yaml` | `valid policy bundle(s)` |
| `typos content/mondoo-kubernetes-best-practices.mql.yaml` | silent |
| structural diff vs `origin/main` | `trees identical apart from docs.desc and policy version` |
| substance gate | `TOTAL SPECIFICS LOST: 0 across 0 checks (of 49)` |

Zero substance losses, so no waivers. Every backticked literal and numeric threshold in the old text survives into the new: `default`, `ownerReferences`, `NoSchedule`, `minReplicas`, `minReplicas: 1`, and the 15 day certificate margin.

The policy version is unchanged, as in the sibling PRs.

### DIFF
--- a/content/mondoo-kubernetes-best-practices.mql.yaml
+++ b/content/mondoo-kubernetes-best-practices.mql.yaml
@@ -118,18 +118,19 @@ queries:
       }
     docs:
       desc: |
-        This check verifies that Pods have the `ownerReferences` metadata field populated, which indicates they were created by a controller (such as a Deployment, ReplicaSet, DaemonSet, StatefulSet, or Job) rather than directly.
+        This check verifies that every Pod is managed by a controller rather than created on its own.
+
+        When a Deployment, ReplicaSet, StatefulSet, DaemonSet, or Job creates a Pod, the API server records the parent in the Pod's `ownerReferences` metadata. A Pod applied directly from a manifest has that list empty, and nothing in the cluster is responsible for it. This check requires `ownerReferences` to be present and non-empty.
 
         **Why this matters**
 
-        The `ownerReferences` metadata field establishes the relationship between a Pod and its managing controller. Pods without this metadata are typically "orphaned" or manually created, which creates operational risks:
+        A standalone Pod is the one workload Kubernetes does not put back. If its node fails, is drained for an upgrade, or is removed by a scale-down, the Pod is deleted and stays deleted; the self-healing behavior people expect from the cluster comes from the controller, not from the Pod. A service running as a bare Pod is therefore down until a person notices and reapplies the manifest, and node maintenance turns from a routine operation into an outage.
 
-          - **No automatic recovery**: Without a controller tracking the Pod, Kubernetes will not recreate it if it crashes or the node fails.
-          - **No garbage collection**: Orphaned Pods are not automatically cleaned up when their parent resources are deleted.
-          - **Manual lifecycle management**: Updates, scaling, and deletion must be handled manually for each Pod.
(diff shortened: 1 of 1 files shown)


tas50:
Check defects found while reading these queries are now filed as **#3468**. Headline: all seven `*-ports-hostport` checks walk only `containers` and miss a `hostPort` on an init container; `control-plane-taint` asserts only the `NoSchedule` effect, so an untainted control plane node scores compliant the moment it goes NotReady; and the two probe checks produce no verdict at all for a bare Pod because a list op on the absent `ownerReferences` key skips the whole check.

PR #3467, opened 2026-08-17
📝 unifi: rewrite check descriptions as prose

tas50:
## What this changes

Rewrites `docs.desc` for all **37 checks** in `content/mondoo-unifi-security.mql.yaml` into the house prose format used by the sibling PRs in this campaign.

The old descriptions all shared the same structural defect: a one-line opener followed by `**Why this matters**` and a list of bulleted bold labels. A label tells a reader the *category* of a concern without ever saying what happens. Each bullet is now the sentence it was standing in for.

Every description now:

- opens with `This check verifies that <capability> is <required state>`, leading with the capability rather than the field name;
- names the setting where the reader administers it, which for this bundle means the page in the UniFi Network application and the toggle on it, such as **Settings > Security > Gateway** and the **UPnP** setting, or **Settings > Networks > Global Switch Settings** and **DHCP Snooping**;
- says concretely what an attacker does without the control. This is a `security` category policy, so the failure story names the step: the rogue DHCP lease that installs the attacker as the default gateway and the DNS server, the forged deauthentication that triggers a handshake capture, the UPnP request that publishes a compromised workstation straight to the internet, the forwarded management port that a scanner finds within hours of it opening;
- respects the scope the query actually has. Several checks are scoped to one kind of UniFi object, and the text says so: the guest-mode check only looks at SSIDs whose name contains "guest", the guest network isolation check selects networks by their configured `purpose` rather than by name, and the ICMP check governs redirects the gateway *receives*, with sending being a separate setting on the same page.

Network gear gets its risk framing on its own terms. The device is the network, so a compromise means reading, mirroring, or dropping traffic beneath every control that operates on hosts; availability often outranks confidentiality, because a broadcast storm or a filled connection-tracking table takes out the management path an administrator would use to fix it; and weak defaults persist for years because this equipment is rarely reconfigured after installation.

No query, filter, tag, impact, title, or variant list was touched. **The policy version is not bumped**, matching the sibling PRs.

Median description length: **98 words before, 310 after**.

## Gates

| Gate | Result |
|---|---|
| `cnspec policy lint ./content/mondoo-unifi-security.mql.yaml` | `valid policy bundle(s)` |
| `typos content/mondoo-unifi-security.mql.yaml` | silent |
| substance gate | **1 specific lost across 1 check of 37** |

The single loss is the bare provider field name `requiredVersion` in `mondoo-unifi-security-devices-firmware-uptodate`. It falls under the raw-MQL-fragment exception, and the fact survives in words as "the minimum version the controller records as required for that device", together with the semantic-version ordering used for the comparison.

### DIFF
--- a/content/mondoo-unifi-security.mql.yaml
+++ b/content/mondoo-unifi-security.mql.yaml
@@ -122,15 +122,19 @@ queries:
           mondoo.com/filter-icon: terraform
     docs:
       desc: |
-        This check ensures that no enabled wireless networks on the UniFi controller use open (no encryption) authentication.
+        This check verifies that every enabled wireless network on the site requires encryption before a client can join.
+
+        In the UniFi Network application each SSID is configured under **Settings > WiFi**, and its **Security** setting decides whether the radio link is encrypted. Setting it to **Open** removes both the encryption and the credential, so any device in range associates and receives an address. The check reads every enabled wireless network on the site and requires **Security** to be anything other than **Open**, which in practice means WPA Personal with a passphrase or WPA Enterprise with 802.1X.
 
         **Why this matters**
 
-        Open wireless networks transmit all traffic in plaintext, allowing anyone within radio range to intercept sensitive data:
-          - Credentials, session tokens, and personal data can be captured with basic tools.
-          - Attackers can perform man-in-the-middle attacks without any authentication barrier.
-          - Open networks provide no protection against rogue clients joining the network.
-          - Even guest networks should use WPA with a shared key or captive portal authentication.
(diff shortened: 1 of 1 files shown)


tas50:
Check defects found while reading these queries are now filed as **#3472**. Headline: `wlans-use-wpa2-or-wpa3` asserts on `security`, which names the authentication family rather than the WPA generation, so a WPA1/TKIP network passes an impact-90 check; the provider exposes `wpaMode`, `wpa3Support`, and `wpaEnc` and none is referenced in the bundle.

PR #3470, opened 2026-08-17
📝 virt: rewrite check descriptions as prose

tas50:
Rewrites `docs.desc` for every check in the two virtualization security
policies: 46 checks in `mondoo-proxmox-security` and 28 in
`mondoo-nutanix-security`. One PR, one commit, both bundles.

## What changed

Both bundles carried the full generated skeleton: a short opener, a
`**Why this matters**` section made of bulleted bold labels, and a
`**Risk mitigation**` section that restated `docs.remediation`. The
labels named the *category* of a concern without ever saying what
happens, and the duplicated remediation section is being removed
everywhere in this campaign.

Every description is now prose:

- The first sentence says what is verified.
- The capability leads; the setting is the instrument, not the subject.
- The setting is named where the reader administers it. For Proxmox
  that is the web interface path, the `pve*`/`qm`/`pct` command, or the
  file under `/etc/pve` an admin edits. For Nutanix it is the specific
  console, Prism Element or Prism Central, and the `ncli` or `acli`
  verb run on a Controller VM.
- Both bundles are `mondoo.com/category: security`, so each says what
  an attacker concretely does rather than that access "may" be gained.

The risk framing is specific to virtualization rather than generic
"attack surface" language: the hypervisor is a trust boundary between
guests that believe they are isolated, so a management-plane compromise
is a compromise of every guest at once, and a guest has no way to
observe a compromise that happened beneath it.

The two bundles are written as two different products. Proxmox
descriptions talk about the cluster filesystem at `/etc/pve`, the web
interface on port 8006, the LXC feature flags, and the sysctl and
`login.defs` baseline that Proxmox inherits from Debian. Nutanix
descriptions talk about Controller VMs, the distinction between Prism
Element and Prism Central, storage containers and volume groups, and
the v4 REST APIs.

Median description length:

| bundle | checks | median before | median after |
|---|---|---|---|
| `mondoo-proxmox-security` | 46 | 130 | 291 |
| `mondoo-nutanix-security` | 28 | 125 | 299 |

## Gates

Run against **both** files; the structural and substance gates were run
once per file.

| gate | proxmox | nutanix |
|---|---|---|
| `cnspec policy lint` | `valid policy bundle(s)` | `valid policy bundle(s)` |
| `typos` | silent | silent |
| `go test -count=1 ./internal/bundle/...` | pass | pass |
| structural diff vs `origin/main` | trees identical apart from docs.desc and policy version | trees identical apart from docs.desc and policy version |
| substance gate | **0 losses across 46 checks** | **0 losses across 28 checks** |

The substance gate compares the set of backticked literals and numeric
thresholds in the old `docs.desc` against the new text. Every one
survives in both bundles, including the values that are easy to drop in
a rewrite: `PermitRootLogin prohibit-password`, `0640`, `policy_in:
DROP`, `unprivileged: 1`, `+spec-ctrl` and `+ssbd`, `pcie_acs_override`,
`2592000`, CVE-2021-4204 and CVE-2022-23222, `PASS_MAX_DAYS` with its
1-to-365 range and the 99999 Debian default, `YESCRYPT`, KMIP port
`5696`, `0.0.0.0/0.0.0.0`, and the 90-day stale-account window.

Nothing outside `docs.desc` is touched: no MQL, no filters, no titles,
no impacts, no tags, no remediation.

### DIFF
--- a/content/mondoo-nutanix-security.mql.yaml
+++ b/content/mondoo-nutanix-security.mql.yaml
@@ -119,18 +119,19 @@ queries:
       nutanix.clusters.where(clusterType == "AOS").all(encryptionScopes.length > 0)
     docs:
       desc: |
-        This check ensures that data-at-rest encryption is enabled on each AOS storage cluster. Without encryption, the data written to physical drives is recoverable by anyone who gains physical access to a disk that is decommissioned, stolen, or returned under warranty.
+        This check verifies that each AOS storage cluster has data-at-rest encryption turned on with a configured scope.
+
+        Encryption is enabled under Settings, then Data-at-Rest Encryption, where you choose the key manager and then the scope, either the whole cluster or individual storage containers. The cluster can use software encryption, self-encrypting drives, or hardware encryption depending on what the hardware supports. There is no command that enables it; `ncli data-at-rest-encryption get-status` on a Controller VM only reports the result.
 
         **Why this matters**
 
-        - **Physical theft:** Encrypted drives are unreadable when removed from the cluster, defeating drive theft and improper disposal.
-        - **Regulatory data protection:** Many frameworks require sensitive data to be encrypted while stored.
-        - **Defense in depth:** Encryption protects data even when other controls, such as access control on the array, are bypassed.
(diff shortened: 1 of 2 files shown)


tas50:
Check defects found while reading these queries are now filed as **#3474**. Headline: `corosync-is-encrypted` accepts `transport: knet`, which has been the PVE 6 default and is not encryption, so essentially every modern cluster passes regardless of its `crypto_cipher` setting; and `vm-spectre-mitigations` passes vacuously on a VM with no `cpu:` line, which is the default and is unmitigated.

tas50:
Correction to my previous comment: the issue number is **#3475**, not #3474. (#3474 is an unrelated PR.)

PR #3474, opened 2026-08-17
📝 mariadb+mysql: rewrite check descriptions as prose

tas50:
Rewrites `docs.desc` for all 72 checks across the two SQL database bundles: 37 in `mondoo-mariadb-security` and 35 in `mondoo-mysql-security`. Both carried the full generated skeleton, bulleted bold labels under **Why this matters** plus a **Risk mitigation** section that restated `docs.remediation`. Both are gone.

## What changed

Each description now opens with what is verified, names the system variable the way an administrator writes it and the option file group it belongs in, and says concretely what an attacker reads or changes without the control. Both policies read configuration from disk rather than opening a SQL connection, so the text says where a value belongs in the option file and notes where a runtime change would not survive a restart. The local socket versus network distinction is stated wherever it decides the answer, and checks that concern an account's rights describe the privilege the account holds rather than a table read.

## The two bundles are not the same product

MariaDB and MySQL overlap in subject but have diverged, so the descriptions were written separately against each bundle's own MQL and remediation rather than adapted from one another:

- MariaDB's `simple_password_check` and `password_reuse_check` plugins against MySQL's `validate_password` component, `default_password_lifetime`, `password_history` / `password_reuse_interval`, and `password_require_current`
- MariaDB's `server_audit` plugin, with its separate load and logging switches and its exclusion list, against MySQL's audit log plugin and `audit_log_policy`
- MariaDB's numeric `log_warnings` verbosity against MySQL's `log_error_verbosity`
- MariaDB's `innodb_encrypt_tables` with its three values including `FORCE`, `encrypt_binlog`, `encrypt_tmp_disk_tables`, and the `file_key_management` key file, against MySQL's `default_table_encryption`, `binlog_encryption`, and redo and undo log encryption backed by a keyring component
- MariaDB's Galera replication and state snapshot transfer checks, which have no MySQL equivalent
- MariaDB's client option groups `[client]`, `[client-server]`, `[client-mariadb]`, `[mariadb-client]` against MySQL's `[client]`, `[client-server]`, `[mysql]` and its `mysql_config_editor` login paths
- MySQL only: `caching_sha2_password` versus `mysql_native_password`, the self-signed certificate MySQL generates on first start when none is configured, the comma-separated `bind_address` list from 8.0.13, and the fact that `skip_grant_tables` also enables `skip_networking` on MySQL

Median description length: 167 words before and 270 after for MariaDB, 160 before and 278 after for MySQL. No policy version bump, matching the sibling PRs in this campaign.

## Gates

All run against both files.

| Gate | Result |
|---|---|
| `cnspec policy lint ./content/mondoo-mariadb-security.mql.yaml` | valid policy bundle(s) |
| `cnspec policy lint ./content/mondoo-mysql-security.mql.yaml` | valid policy bundle(s) |
| `typos` on both files | silent |
| structural diff vs `origin/main`, MariaDB | trees identical apart from docs.desc and policy version |
| structural diff vs `origin/main`, MySQL | trees identical apart from docs.desc and policy version |
| substance gate, MySQL | 0 specifics lost across 35 checks |
| substance gate, MariaDB | 1 specific lost across 37 checks |

The single substance loss is the literal `` `[mysql]` `` in `mondoo-mariadb-security-client-options-no-plaintext-password`. It is dropped deliberately, because it is wrong: `[mysql]` is a MySQL client option group. The MariaDB resource reads `[client]`, `[client-server]`, `[client-mariadb]`, and `[mariadb-client]`, and the new text names those instead.

### DIFF
--- a/content/mondoo-mariadb-security.mql.yaml
+++ b/content/mondoo-mariadb-security.mql.yaml
@@ -139,20 +139,19 @@ queries:
       mariadb.conf.bindAddress.none(_ == "*" || _ == "0.0.0.0" || _ == "::")
     docs:
       desc: |
-        This check ensures that `bind_address` names specific addresses instead of accepting connections on every interface. MariaDB binds to every interface when the option is unset, so an installation that never sets it is reachable from every network the host is attached to.
+        This check verifies that the server answers TCP connections only on addresses an administrator has named, rather than on every interface the host happens to have.
 
-        Where the database only serves applications on the same host, `skip_networking` removes the TCP listener entirely and is stronger still. Where remote access is required, naming the addresses keeps the listener off networks the database was never meant to serve.
+        `bind_address` is written as `bind-address` under the `[mysqld]` group of the server option file, typically `/etc/mysql/mariadb.conf.d/50-server.cnf` or a fragment under `/etc/my.cnf.d`. It takes a comma-separated list of addresses. The check fails the wildcard forms `*`, `0.0.0.0`, and `::`, and it also fails an installation that never sets the option, because an unset value means the same thing. The option cannot be changed at runtime, so it has to be written to the option file and the server restarted.
 
         **Why this matters**
 
-        - **Smaller network surface:** Only the addresses you name can accept a connection attempt.
-        - **Defense in depth behind the firewall:** Binding narrowly means a firewall change does not immediately expose the database.
-        - **Explicit intent:** A named address documents which network the server is expected to serve.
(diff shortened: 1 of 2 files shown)


tas50:
Check defects found while reading these queries are now filed as **#3477**. Headline: both Galera checks short-circuit on `wsrepOn == false`, and `wsrepOn` is read only from the `[galera]` option group, which `ServerGroups` deliberately excludes. A real cluster with its wsrep options under `[mysqld]` reads `false` and passes both checks at impact 85 and 75 with unencrypted replication and rsync SST.

PR #3476, opened 2026-08-17
📝 content: rewrite vLLM and Grafana check descriptions as prose

tas50:
Rewrites `docs.desc` for all 25 checks in two bundles: `mondoo-vllm-security` (14 checks) and `mondoo-grafana-security` (11 checks).

## What changed

Both bundles shipped the generated skeleton: a `**Why this matters**` section made of bulleted bold labels, followed by a `**Risk mitigation**` section that restated `docs.remediation`. A label tells the reader the *category* of a concern without ever saying what happens. Every label is now the sentence it was standing in for, and the duplicated section is gone.

Each description leads with the capability rather than the field, names the setting where the reader actually administers it, and says what an attacker concretely does without the control. Both bundles are `mondoo.com/category: security`, so rule 4 is the attacker form throughout.

**vLLM** is launched from a command line or container entrypoint, so the descriptions name the launch flag or environment variable: `--api-key` and `VLLM_API_KEY`, `--ssl-keyfile` / `--ssl-certfile`, `--allowed-origins`, `--disable-fastapi-docs`, `--enable-tokenizer-info-endpoint`, `--enable-server-load-tracking`, `--profiler-config` and `VLLM_TORCH_PROFILER_DIR`, and `VLLM_SERVER_DEV_MODE`. The framing is that an unauthenticated inference endpoint is compute someone else can spend, and a read path into the system prompt, the retrieval context, and whatever tools the model has been wired into.

**Grafana** is administered from the console, the HTTP API, or `grafana.ini`, so the descriptions name the console path (**Administration > Users and access > Service accounts**, **Connections > Data sources**, **Alerting > Notification policies**), the API field (`basicAuth`, `access: proxy`, `disableResolveMessage`, `url`), or the `[security]` keys `admin_user` and `admin_password` with their `GF_SECURITY_ADMIN_*` equivalents. The framing is that Grafana holds data source credentials and a map of the estate: an attacker with a Grafana session reads the dashboards that describe what exists and where, and can often reach the data sources themselves through the proxy.

Median description length: vLLM 139 → 279 words, Grafana 130 → 295 words. No policy version bump, matching the sibling PRs in this campaign.

## Gates

| Gate | vLLM | Grafana |
|---|---|---|
| `cnspec policy lint` | valid policy bundle(s) | valid policy bundle(s) |
| `typos` | silent | silent |
| `go test -count=1 ./internal/bundle/...` | ok | ok |
| structural diff vs `origin/main` | trees identical apart from docs.desc and policy version | trees identical apart from docs.desc and policy version |
| substance gate | 0 specifics lost across 0 of 14 checks | 0 specifics lost across 0 of 11 checks |

Every backticked literal and numeric threshold from the old text survives in the new text, so there are no waived losses to explain.

### DIFF
--- a/content/mondoo-grafana-security.mql.yaml
+++ b/content/mondoo-grafana-security.mql.yaml
@@ -75,18 +75,17 @@ queries:
     mql: grafana.serviceAccounts.none(role == "Admin")
     docs:
       desc: |
-        This check ensures that service accounts are configured to use a least-privilege role rather than the Admin role. Admin service accounts can modify organization settings, manage users, and install plugins, creating a broad attack surface if a token is compromised.
+        This check verifies that no Grafana service account holds the Admin role.
+
+        Service accounts are the machine identities that dashboard-as-code pipelines, provisioning jobs, and alerting integrations authenticate as, and each one carries an organization role. That role is set under **Administration > Users and access > Service accounts** in the Grafana UI, or through the service account API, and it should be Viewer for read-only automation and Editor for anything that writes dashboards. This check requires that none of them is Admin.
 
         **Why this matters**
 
-        - **Least privilege:** Restricting service accounts to Viewer or Editor roles aligns automation credentials with the minimum access they need to function.
-        - **Reduced blast radius:** A compromised non-Admin token cannot manage users, alter datasource credentials, or install plugins.
-        - **Auditability:** Limiting Admin assignments to human-reviewed accounts keeps privileged access decisions deliberate and traceable.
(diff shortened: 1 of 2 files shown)


tas50:
Check defects found while reading these queries are now filed as **#3478**. Headline: `version-not-exposed` nulls the field on any probe error and asserts `== empty`, so a host that times out scores compliant, while the sibling `*Exposed` checks assert `== false` and correctly fail the same unknown. And `use-tls` resolves to `strings.HasPrefix(c.baseURL, "https://")`, so it tests the URL the scanner was given rather than anything about the server.

tas50:
Correction: the issue number is **#3483**, not #3478.

PR #3478, opened 2026-08-17
📝 content: rewrite search and cloud database check descriptions as prose

tas50:
Rewrites `docs.desc` for all 14 checks across five bundles: `mondoo-elasticsearch-security`, `mondoo-opensearch-security`, `mondoo-neon-security`, `mondoo-clickhousedb-security`, and `mondoo-clickhousecloud-security`.

Every description carried the full generated skeleton: a bulleted `**Why this matters**` list of bold category labels followed by a `**Risk mitigation**` section that restated `docs.remediation`. Both are gone. Each description now opens by naming what is verified, says where the reader administers the setting and what the console, CLI, API, or configuration file calls it, then explains in prose what an attacker concretely does without the control and what to pair the check with.

Each bundle is written from its own product's configuration surface rather than from a shared template. Elasticsearch names the X-Pack settings in `elasticsearch.yml`; OpenSearch names the Security plugin's `config.yml` and `audit.yml` and the `securityadmin.sh` push that makes a dynamic edit take effect. Self-hosted ClickHouse names `config.xml`, `users.xml`, and the `system` tables; ClickHouse Cloud names the console IP access list and the service API, which have no such files. Neon is described as the managed control plane it is, including the branch history a compromised member account reaches and the creation-time-only nature of project password storage.

Median description length per bundle, before and after:

| Bundle | Checks | Before | After |
|---|---|---|---|
| elasticsearch | 4 | 154.0 | 314.0 |
| opensearch | 3 | 145.0 | 323.0 |
| neon | 3 | 156.0 | 329.0 |
| clickhousedb | 2 | 113.5 | 341.0 |
| clickhousecloud | 2 | 115.0 | 335.5 |

Gates run:

- `cnspec policy lint`: valid policy bundle(s) for all five files
- `typos`: silent on all five files
- `go test -count=1 ./internal/bundle/...`: ok
- structural diff against `origin/main`: trees identical apart from docs.desc and policy version, for all five files
- substance gate: 0 specifics lost, 0 of 14 checks affected

### DIFF
--- a/content/mondoo-clickhousecloud-security.mql.yaml
+++ b/content/mondoo-clickhousecloud-security.mql.yaml
@@ -66,19 +66,20 @@ queries:
           mondoo.com/filter-title: Terraform State
           mondoo.com/filter-icon: terraform
     docs:
-      desc: |-
-        This check ensures that no ClickHouse Cloud service is reachable from any IP address. A service is open to all IPs when its IP access list contains a wildcard entry such as `0.0.0.0/0` or `::/0`, which exposes the database endpoint to the entire internet.
+      desc: |
+        This check verifies that no ClickHouse Cloud service accepts connections from every address on the internet.
+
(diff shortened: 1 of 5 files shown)


tas50:
Check defects found while reading these queries are filed as **#3489**. Headline: `elasticsearch-anonymous-access-disabled` reads a field the provider decodes from a non-pointer struct, and Elasticsearch omits `security.anonymous` entirely when security is disabled, so a cluster running `xpack.security.enabled: false` scores pass at impact 80.

PR #3479, opened 2026-08-17
📝 content: rewrite database check descriptions as prose

tas50:
Rewrites `docs.desc` for all 44 checks across the five self-managed database policies.

| Bundle | Checks |
|---|---|
| `mondoo-postgresql-security` | 29 |
| `mondoo-mssql-security` | 4 |
| `mondoo-mongodb-security` | 4 |
| `mondoo-cassandra-security` | 4 |
| `mondoo-redis-security` | 3 |

## What changed

All five bundles carried the generated skeleton: a `**Why this matters**` section made of bulleted bold labels, followed by a `**Risk mitigation**` section that restated `docs.remediation`. The labels named the *category* of a concern without ever saying what happens, and the duplicated section added nothing the remediation did not already carry. Both are gone.

Each description now opens with what is verified, names the setting **where the reader administers it**, and says what an attacker concretely does without the control rather than gesturing at "unauthorized access".

Because these are five different engines with five different configuration mechanisms, the descriptions deliberately do not read as one template:

- **PostgreSQL** names the directive and the file it lives in, says whether a reload or a restart is required, and notes where `ALTER SYSTEM` writes `postgresql.auto.conf` and overrides the value the check reads. `pg_hba.conf` checks describe the rule columns; `pg_ident.conf` checks describe the mapping.
- **SQL Server** names the `sp_configure` option, the `show advanced options` prerequisite and the `RECONFIGURE` that makes a value running, or the Configuration Manager and Management Studio path where the setting is a server property.
- **MongoDB** names the `mongod.conf` key under `security:` or `net:` and the `mongod` restart.
- **Cassandra** treats the `authenticator` and the `authorizer` in `cassandra.yaml` as the separate classes they are, and notes the rolling restart.
- **Redis** names the `redis.conf` directive and the `CONFIG REWRITE` a runtime change needs to persist, and the `protected-mode` and `requirepass` descriptions say plainly why those settings exist.

Descriptions that turn on the difference between a local socket connection and a network one, or between an authenticated and an anonymous default, say so explicitly rather than treating it as incidental.

## Median word count

| Bundle | Before | After |
|---|---|---|
| PostgreSQL | 172 | 265 |
| SQL Server | 128 | 286 |
| MongoDB | 129 | 274 |
| Cassandra | 126 | 281 |
| Redis | 142 | 286 |
| **All 44** | **161** | **269** |

No policy version was bumped.

## Gates

Run per file, all five:

- `cnspec policy lint ./content/<policy>.mql.yaml` → `valid policy bundle(s)` for all five.
- `typos content/<policy>.mql.yaml` → silent for all five.
- `go test -count=1 ./internal/bundle/...` → `ok`.
- Structural diff against `origin/main` → `trees identical apart from docs.desc and policy version` for all five.
- Substance gate → **0 specifics lost** on every bundle: 0/29 PostgreSQL, 0/4 SQL Server, 0/4 MongoDB, 0/4 Cassandra, 0/3 Redis. Every backticked literal and numeric threshold in the old text survives into the new, so nothing was waived.

### DIFF
--- a/content/mondoo-cassandra-security.mql.yaml
+++ b/content/mondoo-cassandra-security.mql.yaml
@@ -66,19 +66,20 @@ queries:
     mql: |
       cassandra.cluster.security.authenticationEnabled == true
     docs:
-      desc: |-
-        This check ensures that the configured authenticator is not `AllowAllAuthenticator`, so clients must present valid credentials before they can connect. With `AllowAllAuthenticator`, any client that reaches the cluster connects with no credentials at all.
+      desc: |
+        This check verifies that a client must present credentials before the cluster will open a session.
+
+        The `authenticator` setting in `cassandra.yaml` names the class that validates connections. The check fails on `AllowAllAuthenticator`, which is the shipped default and accepts anyone, and passes on a real implementation such as `PasswordAuthenticator`. Changing it needs a rolling restart so every node picks it up, and roles with passwords have to exist before authorized clients can reconnect.
 
         **Why this matters**
 
-        - **No anonymous access:** Requiring authentication ties every connection to a known role rather than allowing open access to anyone on the network.
-        - **Credential enforcement:** `PasswordAuthenticator` validates a username and password before granting a session.
-        - **Foundation for authorization:** Permission enforcement is meaningless without an authenticated identity to enforce it against.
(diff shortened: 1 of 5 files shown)


tas50:
Check defects found while reading these queries are filed as **#3484**.

PR #3480, opened 2026-08-17
📝 content: rewrite Netlify and HCP check descriptions as prose

tas50:
Rewrites `docs.desc` for all four checks across `mondoo-netlify-security` and `mondoo-hcp-security`.

All four carried the full generated skeleton: a bulleted `**Why this matters**` list of bold category labels followed by a `**Risk mitigation**` section that restated `docs.remediation`. Both are gone. Each description now opens by naming what is verified, says where the reader administers the setting in the product's own UI or portal and what the API or Terraform resource calls it, then explains in prose what an attacker concretely does without the control and what to pair the check with.

The two products share nothing, so the descriptions do not either. The Netlify pair is written around what control of a team account or an unencrypted first request does to site visitors, and it keeps the note that MFA enforcement is a paid-team capability. The HCP pair is written around a Vault cluster as the holder of everything else's credentials, naming the `hcp_vault_cluster` arguments and the audit destinations the `audit_log_config` block accepts.

Median description length per bundle, before and after:

| Bundle | Checks | Before | After |
|---|---|---|---|
| netlify | 2 | 154.0 | 341.0 |
| hcp | 2 | 153.0 | 362.5 |

Gates run:

- `cnspec policy lint`: valid policy bundle(s) for both files
- `typos`: silent on both files
- `go test -count=1 ./internal/bundle/...`: ok
- structural diff against `origin/main`: trees identical apart from docs.desc and policy version, for both files
- substance gate: 0 specifics lost, 0 of 4 checks affected

### DIFF
--- a/content/mondoo-hcp-security.mql.yaml
+++ b/content/mondoo-hcp-security.mql.yaml
@@ -66,19 +66,20 @@ queries:
           mondoo.com/filter-title: Terraform State
           mondoo.com/filter-icon: terraform
     docs:
-      desc: |-
-        This check ensures that no HCP Vault cluster exposes an unrestricted public endpoint. A cluster may keep its public endpoint enabled only when an IP allowlist restricts which source addresses can reach it; a public endpoint with no allowlist puts the secrets engine on the open internet.
+      desc: |
+        This check verifies that an HCP Vault cluster is not reachable from the whole internet.
+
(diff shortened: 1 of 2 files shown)


tas50:
Check defects found while reading these queries are filed as **#3489**. For this PR specifically: `netlify-account-enforce-mfa` asserts `enforceMfa != "not_enforced"`, and `EnforceMfa` is a plain `string` in the provider, so an account record that omits `enforce_mfa` reads `""` and passes at impact 90. The check's docs identify that as the free-plan case.

PR #3481, opened 2026-08-17
📝 content: rewrite protocol check descriptions as prose

tas50:
## What changed

`docs.desc` rewritten into the house prose format for the three network-protocol policies. Each description now leads with the capability being verified, names the concrete artifact the reader edits — a cipher list directive, an `MX` record set, a response header and its directives — and replaces bulleted bold labels with prose that says what an attacker on the network path, or one who controls a name, concretely does. The `**Risk mitigation**` section, which duplicated `docs.remediation`, is removed throughout.

| Bundle | Rewritten | Left alone |
|---|---|---|
| `mondoo-tls-security` | 23 of 23 | — |
| `mondoo-dns-security` | 8 of 8 | — |
| `mondoo-http-security` | 4 of 13 | 9 already converted, byte-identical to `origin/main` |

All three are `mondoo.com/category: security`, so every description states the attack rather than an engineering failure mode: a downgraded negotiation that lets a recorded session be decrypted later, a first cleartext request that an attacker on the path answers instead of the server, a stale `MX` record that delivers password resets to a host somebody else now controls, a wildcard that makes `login-secure.example.com` resolve and get a certificate issued.

## Specifics preserved

The substance gate reports zero losses across all three bundles, with no waivers. That includes every protocol version and cipher suite name the old text carried: the draft ChaCha20-Poly1305 identifiers `0xCC13`, `0xCC14`, and `0xCC15` and their RFC 7905 replacements, the 40-bit and 56-bit export key sizes, the 398-day certificate lifetime cap, the 30/21/14/7 day expiry scoring bands, the four retired Microsoft mail endpoints, and the five Google Workspace `MX` hosts with their preferences.

## Median word count

| Bundle | Before | After |
|---|---|---|
| tls, 23 checks | 158 | 290 |
| dns, 8 checks | 107.5 | 295.5 |
| http, 4 rewritten | 153.5 | 293 |
| http, whole bundle of 13 | 280 | 290 |

## Gates

Run against all three files:

- `cnspec policy lint ./content/<policy>.mql.yaml` — `valid policy bundle(s)` for each.
- `typos content/<policy>.mql.yaml` — silent.
- `go test -count=1 ./internal/bundle/...` — ok.
- Structural diff against `origin/main` — `trees identical apart from docs.desc and policy version` for each of the three.
- Substance gate — 0 lost specifics in tls, 0 in dns, 0 in http.

### DIFF
--- a/content/mondoo-dns-security.mql.yaml
+++ b/content/mondoo-dns-security.mql.yaml
@@ -72,13 +72,17 @@ queries:
       dns.records.where(type == "CNAME") == empty
     docs:
       desc: |
-        This check verifies that the root domain has no CNAME record in its DNS settings. The DNS standard prohibits a CNAME at the apex domain because a CNAME cannot coexist with the SOA and NS records that every zone apex must carry.
+        This check verifies that the zone apex, the bare domain with no label in front of it, resolves through an address record and carries no CNAME.
+
+        A CNAME declares that this name is an alias for another name and that the other name's records are authoritative for everything. That is incompatible with the apex, which must carry its own SOA and NS records for the zone to exist at all. The fix in your DNS provider's control panel is to delete the apex CNAME and publish an `A` or `AAAA` record, or, where the target is a hostname whose address moves, the provider's alias record type: `ALIAS`, `ANAME`, or CNAME flattening, each of which resolves the target and answers with an address.
 
         **Why this matters**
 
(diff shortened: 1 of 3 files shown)


tas50:
Check defects found while reading these queries are filed as **#3485**.

PR #3482, opened 2026-08-17
📝 saas: rewrite check descriptions as prose

tas50:
Rewrites `docs.desc` for all 50 checks across the four collaboration SaaS bundles:

| Bundle | Checks | Median words before | after |
|---|---:|---:|---:|
| `mondoo-m365-security` | 23 | 217 | 307 |
| `mondoo-google-workspace-security` | 12 | 144 | 304 |
| `mondoo-slack-security` | 8 | 178 | 309 |
| `mondoo-atlassian-security` | 7 | 194 | 303 |

## What changed

All four bundles carried the full generated skeleton: bulleted bold labels under `**Why this matters**` plus a `**Risk mitigation**` section that restated `docs.remediation`. Both are gone.

Each description now leads with the capability being verified rather than the field name, names the console and the path where the reader actually administers the setting, and says in prose what an attacker concretely reads and how they got there. The bold labels are turned into the sentences they were standing in for.

Because these are four different vendors, the admin surfaces are named per product rather than generically:

- **Microsoft 365** points at the right console for each control: Microsoft Entra admin center for Conditional Access, roles, and consent; Microsoft 365 admin center for password expiry and Global Administrator assignment; Microsoft Defender portal for Safe Links, Safe Attachments, anti-phishing, DKIM, and outbound forwarding; Exchange admin center for transport rules and remote domains; Intune admin center for device profiles; SharePoint admin center for external sharing; Microsoft Purview for the unified audit log. Graph PowerShell is named where it is the practical path.
- **Google Workspace** points at the Admin console at admin.google.com, and calls out organizational unit scoping where it is load-bearing, which it is for 2-step verification enforcement and for the security-key requirement on super admins.
- **Slack** distinguishes workspace settings from Enterprise Grid org settings, and notes where a control or a console column is absent on some plans.
- **Atlassian** separates organization settings at admin.atlassian.com from per-product settings inside Jira and Confluence, including the Jira Service Management knowledge base link that drives unlicensed Confluence access.

No policy version bumps. No field other than `docs.desc` is touched.

## Gates

Run against all four files:

| Gate | Result |
|---|---|
| `cnspec policy lint` | `valid policy bundle(s)`, all four |
| `typos` | silent |
| structural diff | `trees identical apart from docs.desc and policy version`, all four |
| substance gate | 4 losses across 50 checks |

The 4 substance losses are all on the deliberate exception list, and no backticked literal was lost in any bundle:

- `800` twice in `mondoo-m365-security`, from the `NIST SP 800-63B` citation in the password-expiry and mobile-password-length checks. Framework and guidance citations do not belong in prose; the fact survives as "current federal password guidance".
- `27001` once each in `mondoo-google-workspace-security` and `mondoo-slack-security`, from an `ISO 27001` citation in a compliance sentence. The `compliance/*` tags already carry that mapping.

### DIFF
--- a/content/mondoo-atlassian-security.mql.yaml
+++ b/content/mondoo-atlassian-security.mql.yaml
@@ -72,19 +72,19 @@ queries:
       atlassian.admin.organization.policies.any(policyType == "ip-allowlist" && status == "enabled")
     docs:
       desc: |
-        This check ensures that the Atlassian organization has at least one IP allowlist policy enabled. Without an active IP allowlist, every Atlassian Cloud product is reachable from any address on the public internet, so a stolen credential or API token can be used from anywhere an attacker happens to be.
+        This check verifies that the Atlassian organization restricts product access to a list of approved source addresses.
+
+        An IP allowlist names the network ranges from which Atlassian Cloud products accept traffic, and it is evaluated before any credential is examined. You create one at admin.atlassian.com under **Security > IP allowlists**, name it, add the ranges of corporate offices, VPN exit nodes, and approved cloud egress points, then apply it to the products you want to protect. This check requires at least one allowlist on the organization whose status is enabled. Jira, Jira Service Management, Confluence, and Compass need a Premium plan for the feature, and Atlassian Analytics needs Enterprise.
 
         **Why this matters**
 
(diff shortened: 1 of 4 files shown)


tas50:
Check defects found while reading these queries are filed as **#3486**.

PR #3487, opened 2026-08-17
📝 content: rewrite check descriptions as prose

tas50:
## What changed

The operating-system policies had most of their check descriptions converted to the house prose format in an earlier pass. This converts the **41 stragglers** that still carried the old bulleted skeleton. Every already-converted description in these files is left byte-identical, so the diff is confined to the checks that needed work.

| Bundle | Rewritten / total | Category |
|---|---|---|
| `mondoo-linux-security` | 17 / 120 | security |
| `mondoo-windows-11-compatibility` | 8 / 8 | best-practices |
| `mondoo-linux-workstation-security` | 5 / 5 | security |
| `mondoo-windows-workstation-security` | 4 / 4 | security |
| `mondoo-linux-snmp-policy` | 3 / 3 | best-practices |
| `mondoo-linux-operational-policy` | 2 / 2 | best-practices |
| `mondoo-freebsd-security` | 2 / 56 | security |

Each rewritten description leads with the capability rather than the setting, names the setting where the reader administers it, and says concretely what goes wrong without the control. Two banned headings are removed along the way: the `**Risk mitigation**` section that five `mondoo-linux-security` descriptions still carried, which duplicated `docs.remediation`, and the `**Scope**` heading in the FreeBSD pkg signature check, which is folded into prose.

`mondoo-windows-11-compatibility` and `mondoo-linux-operational-policy` are best-practices policies, so those descriptions state the engineering failure the check prevents rather than inventing an attacker.

Median word count across the 41 rewritten descriptions: **122 before, 303 after**. No policy versions were bumped.

## Gates

All run against all seven modified files.

- `cnspec policy lint` — `valid policy bundle(s)` for each of the seven files
- `typos` — silent
- `go test -count=1 ./internal/bundle/...` — ok
- structural diff against `origin/main` — `trees identical apart from docs.desc and policy version` for each of the seven files
- substance gate — **6 specifics lost, all in `mondoo-linux-security`, all on the deliberate exception list. Zero losses in the other six bundles.**

The six:

- `mondoo-linux-security-aide-is-installed`, four numbers: `11.5`, `800`, `53`, `7`. These are the PCI-DSS and NIST citation numbers from the old text. Compliance framework citations do not belong in prose; the `compliance/*` tags carry the mapping.
- `mondoo-linux-security-file-deletion-events-by-users-are-collected`, two literals. Both are artifacts of a fenced shell block in the old description: the `awk` one-liner that read `UID_MIN` out of `/etc/login.defs`, and the sentence fragment the backtick regex split off from around it. The fact survives in words, which name `UID_MIN`, `/etc/login.defs`, and the `auid>=1000` substitution.

### DIFF
--- a/content/mondoo-freebsd-security.mql.yaml
+++ b/content/mondoo-freebsd-security.mql.yaml
@@ -6823,13 +6823,17 @@ queries:
         - url: https://man.freebsd.org/cgi/man.cgi?query=ntpd&sektion=8
           title: FreeBSD ntpd(8)
       desc: |
-        This check verifies that a time synchronization daemon is enabled and running. FreeBSD ships `ntpd` in the base system, but it is not enabled by default; OpenNTPD and chrony are available as alternatives from packages.
+        This check verifies that a time synchronization daemon is enabled and running. FreeBSD ships `ntpd` in the base system but leaves it disabled, and OpenNTPD and chrony are available from packages as alternatives. Any of the three satisfies the check, and each is turned on with its own `_enable` variable in `/etc/rc.conf`.
 
         **Why this matters**
 
-        - **Log and forensic correlation:** Accurate clocks let events from multiple hosts be lined up on a single timeline. Skewed clocks make it impossible to reconstruct the true order of events during an incident investigation.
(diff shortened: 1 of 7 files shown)


tas50:
Check defects found while reading these queries are now filed as **#3488**. Headline: `linux-snmp-v3-user-file-protected` is admitted by a group filter that matches any Linux host with `snmpd` or `net-snmp`, but hardcodes `/var/lib/snmp/snmpd.conf` owned by `Debian-snmp`, so every RHEL and SUSE host fails it regardless of configuration.

PR #3523, opened 2026-08-18
✨ redis: cover network exposure, the default user, persistence, and client certs

chris-rock:
Stacked on #3522 (the live validation harness). Review that one first.

Related: **mondoohq/mql#10066** — a provider defect this PR works around by removing a check.

## Why

The policy assessed three things: TLS, ACL passwords, protected mode. That was read as a provider ceiling — "Redis at 4 resources, any depth increase needs provider work first". It is not. `redisdb.instance` already exposes `bind`, `bindsAllInterfaces`, `tlsAuthClients`, `aclFile`, `mode`, and `users[].isDefault`; `redisdb.instance.config` exposes persistence, `maxmemory`, and the eviction policy. The fields were sitting unused.

One trap worth knowing for anyone extending this: the bare `redisdb.config` resource returns nothing. It has to be reached as `redisdb.instance.config`.

## The six checks

Each is from [Redis's recommended security practices](https://redis.io/docs/latest/operate/rs/security/recommended-security-practices/):

| Check | Impact | |
|---|---|---|
| `default-user-disabled` | 70 | The account nobody owns. Carries `~* &* +@all` on a fresh server, and its name is known in advance on every Redis in the world. |
| `not-bound-to-all-interfaces` | 80 | A wildcard bind follows the host's addresses, not the operator's intent — a later public IP publishes the database with no config change. |
| `tls-client-certificate-auth` | 60 | Without it a leaked password is a live session. With it the connection is refused at the handshake. |
| `persistence-enabled` | 40 | Losing rate-limit counters to a restart removes the brake on credential stuffing at the moment the service returns. |
| `memory-limit-set` | 30 | An unbounded in-memory database is a DoS primitive for anyone who can write to it; the OOM killer takes the process with no final snapshot. |

## One check removed

`mondoo-redis-security-plaintext-port-disabled` read `redisdb.instance.port == 0`. The provider populates `port` from `INFO`'s `tcp_port`, which reports **whichever listener is active** — so on a server with `port 0` and a TLS listener bound, the configuration the check's own remediation produces, the provider returns 6379 and the check fails.

```console
$ redis-cli --tls ... CONFIG GET port
port
0
$ # provider reports: port: 6379
```

The nearest workaround, `tlsEnabled && port == tlsPort`, is correct only *because* of the bug and inverts once it is fixed, so it is not worth adopting. Filed as mondoohq/mql#10066; the check returns with the provider fix, and the `redis-tls` fixture already reproduces it.

**Net: 3 → 8 checks.**

## Fixtures

`content/validation/live/redisdb.py` — three configurations, every check seen from both sides:

- `defaults` — the image as it ships
- `hardened` — ACL users, one bound address, memory capped, persistence on
- `tls-only` — a TLS listener with client certificates off, persistence off

The hardened fixture sits on its own Docker network with a fixed address. A container on the default bridge gets its address at start, so a config file baked beforehand cannot name one, and `bind` would have to stay a wildcard — leaving the bound-address check with no pass fixture at all.

## Verification

```
make test/content/live/redisdb   →  40 passed, 0 failed
```

`cnspec policy lint` clean; compliance mapping tests pass; `remediation/code/bash.py` exits 0 over the new snippets.

### DIFF
--- a/content/mondoo-redis-security.mql.yaml
+++ b/content/mondoo-redis-security.mql.yaml
@@ -3,7 +3,7 @@
 policies:
   - uid: mondoo-redis-security
     name: Mondoo Redis Security
-    version: 1.0.0
+    version: 1.1.0
     license: BUSL-1.1
     tags:
       mondoo.com/category: security
@@ -13,15 +13,16 @@ policies:
     authors:
       - name: Mondoo, Inc.
         email: hello@mondoo.com
-    summary: Harden Redis and Valkey authentication, network exposure, and transport encryption
+    summary: Harden Redis and Valkey authentication, network exposure, transport encryption, and durability
     docs:
       desc: |
         The Mondoo Redis Security policy connects to a running Redis or Valkey server and assesses its effective, runtime state. It reads the server's live `INFO` output, its `CONFIG GET` parameters, and the access-control users from `ACL LIST`, so the checks reflect the configuration the server is actually enforcing rather than a file on disk.
 
-        This policy provides security checks across three areas:
-        - Authentication strength, enforced through access-control (ACL) passwords
-        - Network exposure, governed by protected mode
-        - Transport encryption for client connections via TLS
+        This policy provides security checks across four areas:
+        - Authentication strength, enforced through access-control (ACL) passwords and the state of the built-in `default` user
+        - Network exposure, governed by protected mode and by the addresses the server binds to
+        - Transport encryption for client connections via TLS, including the plaintext listener and client-certificate policy
+        - Durability and resource limits, covering persistence to disk and the ceiling on memory the server will consume
 
         Left with a passwordless default user, protected mode off, and no TLS, a Redis or Valkey server on a reachable network hands full read and write access to anyone who can open a socket to its port, with every command and credential crossing the wire in cleartext.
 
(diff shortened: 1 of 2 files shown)


chris-rock:
### Update: the provider fixes landed, so the removed check is back

The mql-side fixes are implemented and verified against real servers in a worktree (`chris-rock/db-provider-fixes`, not yet pushed):

| | |
|---|---|
| mondoohq/mql#10066 | `redisdb.instance.port` now reads `CONFIG GET port` instead of `INFO`'s `tcp_port` |
| mondoohq/mql#10067 | `clickhousedb` decodes a scalar `auth_type`, so the 24.8 LTS line assesses at all |
| mondoohq/mql#10068 | `clickhousedb.instance` gains `tcpPort`, `tcpPortSecure`, `httpPort`, `httpsPort`, `tlsEnabled` via `getServerPort()` |

Both checks this stack removed are restored, **unchanged in intent** — the question each asked was right, the field underneath answered a different one:

- `mondoo-redis-security-plaintext-port-disabled` — body is still `redisdb.instance.port == 0`
- `mondoo-clickhousedb-security-secure-tcp-port-enabled` — body is now `clickhousedb.instance.tcpPortSecure != 0`, and its audit steps say where the value actually lives, since `system.server_settings` will never have it

Verified with locally-built providers:

```
make test/content/live/redisdb        →  45 passed, 0 failed
```

One thing worth calling out, because it is the harness earning its keep twice in one session: the 24.8 fixture recorded four `error` expectations for the `auth_type` defect. When the provider fix landed, **those four expectations failed** — the fixture noticed the bug was gone and sent us back to update it. They now expect the same verdicts as the 25.8 fixture.

⚠️ These PRs now need a provider newer than the fixes. They should not merge before the mql side does.

chris-rock:
mql PRs are open: mondoohq/mql#10077 (redisdb port) and mondoohq/mql#10078 (clickhousedb LTS + listeners). This PR needs a provider built from those before its live suite passes.

PR #3525, opened 2026-08-18
✨ clickhousedb: cover auth strength, host restrictions, privilege scope, quotas

chris-rock:
Stacked on #3522 (the live validation harness). Review that one first.

Related: **mondoohq/mql#10067** (checks error on the 24.8 LTS line) and **mondoohq/mql#10068** (no network or TLS posture is queryable).

## Why

The policy asked two questions. **One could never be answered yes, and the other passed on a DeepSeek-shaped deployment.**

`users.all(hasPassword == true)` returns pass against the official image with `CLICKHOUSE_PASSWORD` set — an account reachable from `::/0` whose credential is stored in plaintext in `users.d`:

```console
$ clickhouse-client --query "SELECT name, auth_type, host_ip FROM system.users"
default    ['plaintext_password']    ['::/0']
```

`hasPassword` is true. That is the shape of every publicly reported ClickHouse exposure, and the policy scored it clean.

Effective coverage before this PR: **0 working checks on 24.8, 1 on 25.8.**

## The four checks

| Check | Impact | |
|---|---|---|
| `users-strong-authentication` | 90 | No `no_password` or `plaintext_password`. The entrypoint writes `CLICKHOUSE_PASSWORD` into `users.d` in the clear and **rewrites it on every start** — verified by deleting the file and restarting. |
| `users-host-restricted` | 80 | An account created without a `HOST` clause is stored as `::/0`. This is the last control between a reachable port and a leaked credential. |
| `users-least-privilege` | 70 | `ON *.*` covers databases nobody has created yet. On the default account it includes the table engines that let a query read from `file()`, `url()`, and remote hosts, plus `displaySecretsInShowAndSelect`. |
| `quota-applies-to-all-users` | 40 | Limits attached to a named list decay the first time an account is added by another team. A bulk read against a quota fails partway and leaves a record — a signal that arrives *during* an exfiltration. |

## One check removed

`mondoo-clickhousedb-security-secure-tcp-port-enabled` filtered `serverSettings` for `tcp_port_secure`. That setting is not in `system.server_settings` on any version:

```console
$ clickhouse-client -q "select count() from system.server_settings"   # 143 on 24.8, 228 on 25.8
$ clickhouse-client -q "select name from system.server_settings where name ilike '%port%'"
(no rows)
```

So the query was `[].any()` — **false on every ClickHouse server, forever**. A check that cannot pass is an unfixable HIGH on every asset, which is worse than no check. ClickHouse exposes ports through `getServerPort()` rather than a table; filed as mondoohq/mql#10068 with a suggested resource surface, and the check returns with it.

**Net: 2 → 5 checks**, and the 5 actually run.

## Fixtures

`content/validation/live/clickhousedb.py` — three containers, one of which exists to **hold a provider defect still**.

`clickhouse-lts` runs 24.8, where four of five checks return no verdict at all:

```
sql: Scan error on column index 1, name "auth_type": unsupported Scan,
storing driver.Value type string into type *[]string
```

`system.users.auth_type` is a scalar `Enum8` before 25.x and an `Array(Enum8)` after; the provider decodes only the array. Those four `error` expectations are the regression marker — **the fixture fails when the provider is fixed**, and someone comes back to update it. Filed as mondoohq/mql#10067.

Two fixture details worth flagging:

- The users file is named `zz-cnspec-live.xml`. Files in `users.d` merge in filename order and the image writes its own `default-user.xml`; a name sorting before it lets that fragment re-add a `default` account with no authentication method, and **the server refuses to start**.
- The hardened fixture writes a bootstrap account in, uses it to create the quota, and writes it back out before scanning. An account with access management is the only way to create a quota, and it would then be the one thing failing the least-privilege check.

## Verification

```
make test/content/live/clickhousedb   →  25 passed, 0 failed
```

`cnspec policy lint` clean; compliance mapping tests pass; `remediation/code/bash.py` exits 0 over the new snippets.

### DIFF
--- a/content/mondoo-clickhousedb-security.mql.yaml
+++ b/content/mondoo-clickhousedb-security.mql.yaml
@@ -3,7 +3,7 @@
 policies:
   - uid: mondoo-clickhousedb-security
     name: Mondoo ClickHouse Security
-    version: 1.0.0
+    version: 1.1.0
     license: BUSL-1.1
     tags:
       mondoo.com/category: security
@@ -13,12 +13,17 @@ policies:
     authors:
       - name: Mondoo, Inc.
         email: hello@mondoo.com
-    summary: Harden self-hosted ClickHouse account authentication and transport encryption
+    summary: Harden self-hosted ClickHouse account authentication, network reach, privilege scope, and query quotas
     docs:
       desc: |
-        The Mondoo ClickHouse Security policy connects to a self-hosted ClickHouse server and assesses its authentication and transport posture: whether every account requires a password and whether the secure native TCP port is enabled.
+        The Mondoo ClickHouse Security policy connects to a self-hosted ClickHouse server and assesses the posture of its accounts: how they authenticate, where they may connect from, what they are allowed to reach, and what a single account is allowed to consume.
 
-        Settings are read from the server's `system` tables, so the checks reflect the server's effective configuration rather than a file on disk.
+        This policy provides security checks across three areas:
+        - Authentication, covering both whether an account has a credential and how that credential is stored
+        - Access control, covering the addresses an account may connect from and the scope of the privileges it holds
+        - Resource limits, covering the quotas that bound what any one account can read
+
+        Accounts, grants, and quotas are read from the server's `system` tables, so the checks reflect the server's effective configuration rather than a file on disk. Exposed ClickHouse servers tend to share one shape: a default account reachable from any address, a credential stored in the clear or absent entirely, and server-wide privileges attached to it. These checks are aimed at that shape.
 
         Learn more about scanning databases in the [cnspec documentation](https://mondoo.com/docs/cnspec).
     groups:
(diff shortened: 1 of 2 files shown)


chris-rock:
### Update: the provider fixes landed, so the removed check is back

The mql-side fixes are implemented and verified against real servers in a worktree (`chris-rock/db-provider-fixes`, not yet pushed):

| | |
|---|---|
| mondoohq/mql#10066 | `redisdb.instance.port` now reads `CONFIG GET port` instead of `INFO`'s `tcp_port` |
| mondoohq/mql#10067 | `clickhousedb` decodes a scalar `auth_type`, so the 24.8 LTS line assesses at all |
| mondoohq/mql#10068 | `clickhousedb.instance` gains `tcpPort`, `tcpPortSecure`, `httpPort`, `httpsPort`, `tlsEnabled` via `getServerPort()` |

Both checks this stack removed are restored, **unchanged in intent** — the question each asked was right, the field underneath answered a different one:

- `mondoo-redis-security-plaintext-port-disabled` — body is still `redisdb.instance.port == 0`
- `mondoo-clickhousedb-security-secure-tcp-port-enabled` — body is now `clickhousedb.instance.tcpPortSecure != 0`, and its audit steps say where the value actually lives, since `system.server_settings` will never have it

Verified with locally-built providers:

```
make test/content/live/redisdb        →  45 passed, 0 failed
```

One thing worth calling out, because it is the harness earning its keep twice in one session: the 24.8 fixture recorded four `error` expectations for the `auth_type` defect. When the provider fix landed, **those four expectations failed** — the fixture noticed the bug was gone and sent us back to update it. They now expect the same verdicts as the 25.8 fixture.

⚠️ These PRs now need a provider newer than the fixes. They should not merge before the mql side does.

chris-rock:
mql PRs are open: mondoohq/mql#10077 (redisdb port) and mondoohq/mql#10078 (clickhousedb LTS + listeners). This PR needs a provider built from those before its live suite passes.

PR #3526, opened 2026-08-18
✨ clickhousedb: restore the TLS listener check on the fixed provider

chris-rock:
Stacked on #3525. **Blocked on a provider release carrying mondoohq/mql#10078** — do not merge before then.

## Why this is its own PR

`cnspec policy lint` compiles every check against the *installed* provider schema. `clickhousedb.instance.tcpPortSecure` does not exist in the released clickhousedb 13.0.1, so a bundle referencing it fails lint outright:

```
failed to compile: cannot find field 'tcpPortSecure' in clickhousedb.instance
```

That is one red check taking the whole bundle with it. Splitting it out keeps #3525 mergeable today and isolates the part that genuinely has to wait.

## What it restores

`mondoo-clickhousedb-security-secure-tcp-port-enabled`, removed in #3525 because `system.server_settings` contains no ports on any ClickHouse version, so its query was `[].any()` — false on every server, forever.

The question it asked was always the right one. mondoohq/mql#10078 adds `tcpPort`, `tcpPortSecure`, `httpPort`, `httpsPort`, and `tlsEnabled` on `clickhousedb.instance` via `getServerPort()`, and the check now reads `tcpPortSecure` directly. Its audit steps name `getServerPort()` and say explicitly that the settings table will never carry this, so the next reader does not repeat the search.

The hardened fixture gains a certificate and a secure listener, so the check has a real pass side rather than an exemption.

## Verified

Against a locally-built provider from mql#10078:

```
```

with the check seen failing on both plaintext fixtures and passing on the TLS one.

## Merge order

1. mondoohq/mql#10078
2. a provider release carrying it
3. #3522 → #3525
4. this

### DIFF
--- a/content/mondoo-clickhousedb-security.mql.yaml
+++ b/content/mondoo-clickhousedb-security.mql.yaml
@@ -32,6 +32,10 @@ policies:
         checks:
           - uid: mondoo-clickhousedb-security-users-have-passwords
           - uid: mondoo-clickhousedb-security-users-strong-authentication
+      - title: Transport Encryption
+        filters: asset.platform == "clickhousedb"
+        checks:
+          - uid: mondoo-clickhousedb-security-secure-tcp-port-enabled
       - title: Access Control
         filters: asset.platform == "clickhousedb"
         checks:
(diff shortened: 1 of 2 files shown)


chris-rock:
### The red `lint` check is expected

```
failed to compile: cannot find field 'tcpPortSecure' in clickhousedb.instance
```

CI installs the released `clickhousedb` provider (13.0.1), which predates mondoohq/mql#10078. That is exactly why this check lives in its own PR rather than in #3525 — `cnspec policy lint` compiles every check against the installed schema, so one unreleased field turns the whole bundle red.

It goes green once a provider release carries mql#10078. Nothing to fix here until then; #3525 below it is fully green and mergeable.

PR #3552, opened 2026-08-22
📃 ADR-5: folder-local exceptions

arlimus:
Currently in Draft to get feedback. Please post any ideas to this PR!

### DIFF
--- a/docs/adr/0005-folder-local-exceptions.md
+++ b/docs/adr/0005-folder-local-exceptions.md
@@ -189,12 +189,45 @@ The resolved policy is **not** re-checksummed after this mutation. Changing impa
 not change reporting-job UUIDs, so result storage keys stay valid, but
 `GraphExecutionChecksum` must keep matching what was cached.
 
-**With an upstream**, cnspec sends the exceptions for the asset *before* the asset's
-policy is resolved. The upstream decides what to do with them — apply them, hold them for
-review, reject them — and then returns the resolved policy as it always does, now
-accounting for whatever it accepted. cnspec does not mutate an upstream-resolved policy.
-Both the local report and the upstream record derive from the same resolution, so they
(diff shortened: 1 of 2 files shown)


docs/adr/0005-folder-local-exceptions.md
```diff
@@ -0,0 +1,536 @@
...
+| `title` | no | Free text, carried through to upstream when set. Not an identifier. |
+| `checks` | yes | Check UID, or MRN when precision is wanted. Several per entry. **No globs** in the first pass — you name what you mean. |
+| `action` | yes | See below. |
+| `justification` | **yes** | An exception without a stated reason cannot be defended when it comes up for renewal. |
+| `valid_until` | no | RFC3339. Only meaningful for the scoring actions; on `disable` it has no effect, and lint warns. |
+
+Deliberately absent:
+
+- **No `paths`.** Tools whose ignore file is global need a path field to scope entries. A
+  context config is already scoped to its folder.
```
kaiehrhardt: The “paths” feature would meet the requirement that different solutions/exeptions could be implemented within the same folder. This is essential for mono repos or repos with multiple modules

kaiehrhardt (changes requested):
lgtm, except for the "paths" feature

docs/adr/0005-folder-local-exceptions.md
arlimus: Good point! I'll update it so that we have it covered even in the initial version.

kaiehrhardt (approved):
lgtm

arlimus:
Thank you @kaiehrhardt !! I'll have the implementation up with the v14 release and mark it experimental, i.e.  we can move things around for this release if we find something doesn't work well in practice

PR #3570, opened 2026-08-25
🐛 copy pre-release artifacts to releases.mondoo.com

philipbalinov:
## Summary

Symmetric to [mondoohq/mql#<num>](https://github.com/mondoohq/mql/pulls/philipbalinov): adds a `copy-prerelease-to-releases-mondoo-com` job that fires on pre-release tags (`-alpha`/`-beta`/`-pre`/`-rc`) and copies the GH release artifacts to `releases.mondoo.com/cnspec/<version>/` — without repointing apt/yum "latest".

## Why

The `trigger-installer-release` job in this file is gated on `needs.goreleaser.outputs.skip-publish != 'true'`, so for a pre-release tag it never fires. That means installer's `pkg_published_trigger_build_release.yaml` doesn't run either, and no `releasr release cp` happens. Pre-release artifacts (like v14.0.0-pre.1 today) never make it to `releases.mondoo.com`.

This PR closes that gap for pre-releases without touching the stable-release path.

## Change

New `copy-prerelease-to-releases-mondoo-com` job at the bottom of the file: `if: skip-publish == 'true'`, `needs: goreleaser`, calls `mondoohq/releasr/.github/workflows/github_copy.yaml@main` via `uses:` with `include-prereleases: false` and `update-parent: true`. `secrets: inherit` propagates the GCP + mergebot secrets releasr's workflow needs.

The existing `trigger-installer-release` and `mondoo-operator-cnspec` jobs are unchanged; they still handle the stable path.

## Why `uses:` and not `repository_dispatch:`

Explicit call chain — the copy job shows up as a downstream node under the goreleaser run in the Actions view, rather than a fire-and-forget dispatch landing on releasr. Synchronous, easier to debug, one unified log tree.

## Dependency

Needs [mondoohq/releasr#<num>](https://github.com/mondoohq/releasr/pulls/philipbalinov) merged first — that PR adds the `workflow_call:` trigger to `github_copy.yaml`.

## Test plan

- [ ] Land releasr PR first.
- [ ] Tag a `v14.0.0-pre.2` on cnspec main → verify the new job runs and succeeds → verify `curl -sI https://releases.mondoo.com/cnspec/14.0.0-pre.2/cnspec_14.0.0-pre.2_linux_amd64.tar.gz` returns 200.
- [ ] Verify apt `stable` component for cnspec still returns v13.35.2 (no latest demotion).
- [ ] Tag a stable `v13.35.3` (hypothetical) on the v13 branch → verify the new job **does NOT run**, and the existing installer chain still handles the stable copy.

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -280,3 +280,23 @@ jobs:
           event-type: update
           client-payload: >-
             {"version": ${{ toJSON(github.ref_name) }}}
+
+  # For pre-release tags (-alpha, -beta, -pre, -rc) the trigger-installer-release
+  # job above is skipped (gated on skip-publish != 'true'), so the installer's
+  # pkg_published_trigger_build_release chain never runs and pre-release artifacts
+  # don't end up on releases.mondoo.com. Call releasr's github_copy workflow
+  # explicitly so the deb/rpm/tarball assets from the GH pre-release show up at
+  # releases.mondoo.com/cnspec/<version>/ -- without repointing apt/yum "latest"
+  # to the pre-release (include-prereleases: false in releasr's `idx` step).
+  copy-prerelease-to-releases-mondoo-com:
+    name: Copy pre-release to releases.mondoo.com
+    if: ${{ needs.goreleaser.outputs.skip-publish == 'true' }}
+    needs: goreleaser
+    uses: mondoohq/releasr/.github/workflows/github_copy.yaml@main
+    with:
+      repository: cnspec
+      version: ${{ github.ref_name }}
+      bucket: releases-us.mondoo.io
+      include-prereleases: false
+      update-parent: true
+    secrets: inherit
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -292,6 +292,11 @@ jobs:
     name: Copy pre-release to releases.mondoo.com
     if: ${{ needs.goreleaser.outputs.skip-publish == 'true' }}
     needs: goreleaser
+    # id-token: write lets the called workflow auth to GCP via WIF.
+    # contents: read is required to download the releasr binary.
+    permissions:
+      contents: read
+      id-token: write
     uses: mondoohq/releasr/.github/workflows/github_copy.yaml@main
     with:
       repository: cnspec


chris-rock:
Closing this — the problem it solves has since been fixed upstream, in the place the comment at the top of the job points at.

Thanks for writing it; the diagnosis was right, and the same reasoning is what the installer-side fix was built on.

### It is superseded

The premise here is that `trigger-installer-release` is skipped for a pre-release, so the installer chain never runs and nothing lands on `releases.mondoo.com`. That was true, and is what mondoohq/installer#768 and #771 fixed:

- **#768** split the packaging gate from the bucket publication, so `reindex` runs for a pre-release while apt/yum/chocolatey/homebrew/AUR stay stable-only.
- **#771** fixed the trigger. `prereleased` fires when a pre-release is *created*, not when it is published — `v14.0.0-rc.3`'s release object was created at 18:17:02 and published at 22:07:10, so that activity had long passed by publish time. It listens to `published` now.

Verified end to end on `v14.0.0-rc.3`:

```
cnspec/preview.json   14.0.0-rc.3   channel=preview
cnspec/latest.json    13.38.1       channel=stable     <- unmoved
mql/preview.json      14.0.0-rc.3
install.mondoo.com/package/cnspec/latest.json?channel=preview -> 14.0.0-rc.3
```

Merging this now would give us two publishers writing the same artifacts to the same prefix.

### Two things that would need fixing first anyway

**It calls a private repository's workflow from a public one.** The `uses:` here points at a repo a public reader cannot open. That constraint is why the installer repo sits in the middle as the seam at all — cnspec and mql dispatch to installer, and installer drives the publishing tools.

**The gate no longer means what it meant.** `skip-publish` used to mean both "manual dry run" and "this is a pre-release", and conflating those is what kept pre-releases out of the bucket. #768 separated them, so on `main` today:

```yaml
- name: Skip publish (manual dry run)
  id: skip-publish
  if: inputs.skip-publish == true
```

`if: needs.goreleaser.outputs.skip-publish == 'true'` would now only be true for an explicit dry run, so the job would never fire on a pre-release tag.

Minor: `include-prereleases: false` is accepted but ignored now — channel membership is derived from the version's semver pre-release segment rather than chosen per invocation.

### Still live from this line of work

mondoohq/mql#10619 and #3603 are both wanted and still needed — they close the mql → cnspec bump, which is the last manual step in the chain. #3603 needs a rebase.

PR #3577, opened 2026-08-26
🐛 mark --asset-name as a requested name so discovery keeps it

tas50:
Companion to mondoohq/mql#9880. **Draft until that merges and the mql dependency is bumped** — `Asset.NameOverride` does not exist in the pinned version, so this does not compile on its own.

## Why

`--asset-name` has been a no-op since 13.x. `cnspec` sets the name on the CLI asset, then mql connects; connecting replaces the root asset with the provider's connection asset, and a provider names that asset in its `detect` step after whatever it connected to. Several providers do that unconditionally, so the flag's name is gone before anything reports it.

mql#9880 restores a requested name after connecting. What it cannot do is work out *which* names were requested. A name on the inventory asset is not evidence anyone asked for it — seventeen providers set `Asset.Name` in their own `ParseCLI` before any user input is involved, and improve it in `detect` once connected:

```
mql discover docker 04be48ed3018
inferring the request:  "04be48ed3018"    ParseCLI's raw container id
after this pair:        "vj-name-test"    what detect resolved
```

So the caller marks it. `Asset.NameOverride` says "a person asked for this name," and only that outranks what `detect` computed.

## The change

One line plus a comment, in `getCobraScanConfig`. `scan`, `sbom`, `aibom`, and `vuln` all route through it, so all four are covered.

```go
if assetName != "" && cliRes.Asset != nil {
    cliRes.Asset.Name = assetName
    cliRes.Asset.NameOverride = true
}
```

Assets named in an inventory file are marked on the mql side, in the inventory loader, so those keep winning too.

## Degradation

No marker, no restore. An older `cnspec` against a newer mql gets today's behavior — `--asset-name` still ignored — rather than mis-named assets. A newer `cnspec` against an older mql sets a field nothing reads.

## Verified

Built against mql#9880, using podman for the container:

| command | before | after |
|---|---|---|
| `scan docker 04be48ed3018` | `vj-name-test` | `vj-name-test` |
| `scan docker 04be48ed3018 --asset-name my-chosen-name` | `vj-name-test` | `my-chosen-name` |
| `scan terraform plan … --asset-name tf-plan-ford-sandbox` | `Terraform HCL plan_gcp_simple` | `tf-plan-ford-sandbox` |


vjeffrey:
whats the expected case when we do this on a multi asset scan, e.g. `go run apps/cnspec/cnspec.go scan aws --asset-name testing3 --parallelism 5` nothing gets the name in that case. just wondering if thats expected

PR #3603, opened 2026-08-31
🐛 cnquery-update: handle pre-release tags and unversioned mql

philipbalinov:
## Summary

Teaches `cnquery-update.yml` to handle mql **pre-release tags** (`v14.0.0-rc.1`, `v14.0.0-pre.2`, etc.) and the **unversioned mql module path** on cnspec `main`. Together with [mondoohq/mql#`<num>`](https://github.com/mondoohq/mql/pull/`<num>`) (which unblocks the mql-side dispatch on pre-releases), the mql → cnspec bump becomes fully automatic.

### 1. Version regex accepts pre-release suffixes

Before: `^v?[0-9]+\.[0-9]+\.[0-9]+$` — rejects `v14.0.0-rc.1`. Confirmed empirically today: `gh workflow run cnquery-update.yml -f version=v14.0.0-rc.1` failed at the `Determine Version and Validate` step with "Error: Version is not a valid semantic version."

After: `^v?[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?$` — matches standard semver pre-release identifiers.

### 2. Bump mql handles unversioned mql module paths

Before: `go get "${MQL_MODULE}@${MQL_VERSION}"` — fails when `MQL_MODULE` is unversioned (`go.mondoo.com/mql`, i.e., cnspec/main today) and `MQL_VERSION` is v2+:

```
go: go.mondoo.com/mql@v14.0.0-rc.1: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v14
```

After: check if the module path has a `/vN` suffix. If yes, use the semver ref directly. If no, resolve the tag to a commit SHA via `gh api` and pass that to `go get` — Go accepts SHAs and produces a pseudo-version, matching cnspec/main's current pin shape (`go.mondoo.com/mql v0.0.0-<ts>-<sha>`).

## Why this is safe for stable v13 too

- Cnspec's `v13` support branch imports `go.mondoo.com/mql/v13` (versioned). The `if [[ … /v[0-9]+$ ]]` check keeps the current semver-ref path for that case — no behavior change.
- Cnspec's `main` imports unversioned mql. New SHA path applies. Same as what a human does today with `go get go.mondoo.com/mql@<sha>` when opening a manual bump PR.

## Test plan

- [ ] Merge this + the companion mql PR.
- [ ] Cut a mql pre-release (`v14.0.0-rc.N`) — verify cnspec bump PR opens automatically against cnspec `main`, and its go.mod update produces a pseudo-version.
- [ ] Cut a mql stable v13 patch (`v13.36.N`) on the mql v13 branch — verify the bump PR opens against cnspec `v13` with a proper semver import.
- [ ] `workflow_dispatch cnquery-update.yml -f version=v14.0.0-rc.1 -f dry-run=true` — verify it accepts the pre-release semver.

### DIFF
--- a/.github/workflows/cnquery-update.yml
+++ b/.github/workflows/cnquery-update.yml
@@ -139,7 +139,13 @@ jobs:
           if [[ "${MQL_MODULE}" =~ /v[0-9]+$ ]]; then
             REF="${MQL_VERSION}"
           else
-            REF=$(gh api "/repos/mondoohq/mql/git/refs/tags/${MQL_VERSION}" --jq '.object.sha')
+            # /commits/<ref>, not /git/refs/tags/<tag>: for an ANNOTATED tag
+            # the latter's .object.sha is the tag object, not the commit, and
+            # `go get` on it fails. Every mql tag is lightweight today, but
+            # nothing enforces that -- a tag created through the releases API,
+            # or with `git tag -m`, is annotated. This endpoint resolves either
+            # kind to the commit.
+            REF=$(gh api "/repos/mondoohq/mql/commits/${MQL_VERSION}" --jq '.sha')
             if [ -z "${REF}" ]; then
               echo "Could not resolve ${MQL_VERSION} to a commit SHA on mondoohq/mql"
               exit 1
--- a/.github/workflows/cnquery-update.yml
+++ b/.github/workflows/cnquery-update.yml
@@ -32,14 +32,21 @@ jobs:
           # Determine which version should be released based on event type
           VERSION: ${{ (github.event_name == 'workflow_dispatch' && inputs.version) || github.event.client_payload.version }}
         run: |
-          # Semantic Versioning regex (with optional 'v' prefix). The capture
-          # group matters: what lands in $GITHUB_OUTPUT below is re-emitted
-          # from BASH_REMATCH, not from $VERSION, so only the matched semver
-          # can ever reach a workflow environment file. Validating and then
-          # echoing the raw value would leave the output one regex bug away
+          # Semantic Versioning regex (with optional 'v' prefix and optional
+          # pre-release segment, so 14.0.0-rc.4 is accepted as well as 13.38.1).
+          #
+          # The capture group matters: what lands in $GITHUB_OUTPUT below is
+          # re-emitted from BASH_REMATCH, not from $VERSION, so only the matched
+          # semver can ever reach a workflow environment file. Validating and
+          # then echoing the raw value would leave the output one regex bug away
           # from an attacker-controlled newline injecting extra step outputs.
-          if [[ ! "$VERSION" =~ ^(v?[0-9]+\.[0-9]+\.[0-9]+)$ ]]; then
-            echo "Error: Version is not a valid semantic version."
+          #
+          # The pre-release group therefore has to sit INSIDE the outer one.
+          # Written as two sibling groups it still validates, and BASH_REMATCH[1]
+          # silently loses the suffix -- v14.0.0-rc.4 would bump cnspec to
+          # v14.0.0, a tag that does not exist, with no error anywhere.
+          if [[ ! "$VERSION" =~ ^(v?[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?)$ ]]; then
+            echo "Error: Version is not a valid semantic version: $VERSION"
             exit 1
           fi
           MQL_VERSION="${BASH_REMATCH[1]}"
@@ -110,6 +117,7 @@ jobs:
       - name: Bump mql
         env:
           MQL_VERSION: ${{ steps.version.outputs.version }}
+          GH_TOKEN: ${{ github.token }}
         run: |
           # Take the module path from the branch we are bumping: support
           # branches import go.mondoo.com/mql/v{major}, while main tracks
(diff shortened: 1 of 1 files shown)


chris-rock:
Reviewed against current `main`. The substance is right and still needed — but the rebase has a trap in it that fails silently, so flagging before anyone applies it.

## 1. The rebase must keep the pre-release inside the outer capture group

`main` has changed under hunk 1. It now captures and re-emits from `BASH_REMATCH`, deliberately:

```bash
# ... what lands in $GITHUB_OUTPUT below is re-emitted from BASH_REMATCH, not
# from $VERSION, so only the matched semver can ever reach a workflow
# environment file.
if [[ ! "$VERSION" =~ ^(v?[0-9]+\.[0-9]+\.[0-9]+)$ ]]; then
  ...
fi
MQL_VERSION="${BASH_REMATCH[1]}"
```

The obvious rebase — append this PR's optional group as a **second** group — compiles, passes the validation, and drops the pre-release:

```
                          naive rebase        outer group
v14.0.0-rc.4     ->       v14.0.0             v14.0.0-rc.4
v13.38.1         ->       v13.38.1            v13.38.1
```

`v14.0.0` is a tag that does not exist. `go get` fails, or worse it resolves something unintended, and `echo "${MQL_VERSION}" > VERSION` writes the wrong version. No error anywhere.

So the optional group has to live **inside** the existing one:

```bash
if [[ ! "$VERSION" =~ ^(v?[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?)$ ]]; then
```

Verified both forms in bash; the table above is its output.

Hunk 2 (`Bump mql`) still matches `main` exactly and applies clean.

## 2. `.object.sha` is the tag object for an annotated tag, not the commit

```bash
REF=$(gh api "/repos/mondoohq/mql/git/refs/tags/${MQL_VERSION}" --jq '.object.sha')
```

For an annotated tag (`git tag -a`, and tags GitHub creates through the releases API) `.object.type` is `tag` and `.object.sha` is the tag object — `go get` on it fails.

It works today because every mql tag is lightweight; I checked `v13.38.1`, `v14.0.0-rc.3` and `v14.0.0-rc.4` and all three report `object.type=commit`. But nothing enforces that, and the failure would look like an unrelated `go get` error.

`/repos/mondoohq/mql/commits/<tag>` resolves either kind to the commit:

```bash
REF=$(gh api "/repos/mondoohq/mql/commits/${MQL_VERSION}" --jq '.sha')
```

Same result for `v14.0.0-rc.4` (`b3f038614bd3`), and it can't regress.

## 3. The pseudo-version logic is correct, and already load-bearing

The unversioned-module-path case is real — Go rejects a v2+ semver reference against a path with no major suffix. `cnspec/main` currently pins:

```
go.mondoo.com/mql v0.0.0-20260912163759-b3f038614bd3
```

`b3f038614bd3` is `v14.0.0-rc.4`'s commit, so what this PR automates is exactly what has been done by hand for every rc so far.

## 4. Two sequencing notes

- The regex here doesn't accept build metadata (`+…`). Reasonable for a tag, just noting it's a deliberate narrowing rather than an oversight.
- #3647 renames this file to `mql-update.yml` and is held as a draft behind this PR. I'll rebase mine once this lands — nothing for you to do.

chris-rock:
Rebased onto `main` and pushed. Two commits now — yours, rebased, plus one fix of mine kept separate so it's easy to drop.

## The conflict, and how it's resolved

Hunk 1 conflicted exactly as flagged. `main` now captures and re-emits from `BASH_REMATCH` so that only matched text can reach `$GITHUB_OUTPUT`. Your optional group is now **inside** that capture group rather than beside it:

```bash
if [[ ! "$VERSION" =~ ^(v?[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?)$ ]]; then
```

Written as two sibling groups it still validates and `BASH_REMATCH[1]` silently drops the suffix — `v14.0.0-rc.4` would have bumped cnspec to `v14.0.0`, a tag that does not exist, with no error anywhere. There's a comment in the file now saying so, since the next person to touch this regex will hit the same trap.

Your better error message (`: $VERSION`) is kept. Hunk 2 applied clean.

## The second commit

`.object.sha` on a ref is the **tag object** for an annotated tag, not the commit, and `go get` on a tag-object SHA fails. It works today only because every mql tag happens to be lightweight (checked `v13.38.1`, `v14.0.0-rc.3`, `v14.0.0-rc.4` — all `object.type=commit`), and nothing enforces that. Notably #3604 creates cnspec's tags with `git tag -m`, which *is* annotated, so the shape is already in this pipeline.

```diff
-REF=$(gh api "/repos/mondoohq/mql/git/refs/tags/${MQL_VERSION}" --jq '.object.sha')
+REF=$(gh api "/repos/mondoohq/mql/commits/${MQL_VERSION}" --jq '.sha')
```

Same answer for `v14.0.0-rc.4`, and it can't regress. Separate commit — drop it if you'd rather handle it elsewhere.

## Verified

**Validation and capture**

| input | | captured |
|---|---|---|
| `v14.0.0-rc.4` | accept | `v14.0.0-rc.4` |
| `v13.38.1` | accept | `v13.38.1` |
| `14.0.0-rc.4` | accept | `14.0.0-rc.4` |
| `v13.38.1-pre.2` | accept | `v13.38.1-pre.2` |
| `v14.0.0-rc.4` + newline + `injected=1` | reject | — |
| `v14.0.0-rc.4+9` / `1.0` / `bad` | reject | — |

The newline case is the property main's capture-group comment exists to protect; it still holds.

**The bump reproduces what's already on main.** `/repos/mondoohq/mql/commits/v14.0.0-rc.4` → `b3f038614bd32568d837b178ebcb5d9434c637d2`, and `cnspec/main`'s `go.mod` currently pins `go.mondoo.com/mql v0.0.0-20260912163759-b3f038614bd3`. So what this automates is byte-for-byte the pin that has been produced by hand for every rc.

**Base-branch routing** — `determine-mql-bump-base.sh` already handles pre-releases (its own docstring names `14.0.0-rc.1`, and the bats suite has three pre-release cases). Ran it directly against the four combinations that matter:

| main's go.mod | version | base |
|---|---|---|
| unversioned | `v14.0.0-rc.4` | `main` |
| unversioned | `v13.38.2-rc.1` | `v13` |
| `/v13` suffixed | `v13.38.2-rc.1` | `main` |
| `/v13` suffixed | `v14.0.0-rc.4` | `v14` |

Row 1 is the live case.

**Rest of the file** — no other assumption of a stable version. `version/mql_update_v14.0.0-rc.4` is a valid ref name, and #3604's `${HEAD_REF#version/mql_update_}` still recovers the version from it.

## Sequencing

This goes **first**. Until it lands, `cnquery-update.yml` rejects any `-rc` version, so mql#10619 — which turns on the dispatch for pre-release tags — would send every mql pre-release into a hard validation failure. #10619 has its own review note about the gate it changes.

#3647 renames this file to `mql-update.yml` and is a draft held behind this PR; I'll rebase it once this merges.

PR #3628, opened 2026-09-07
✨ oci: add Terraform variants to four checks the provider does cover

tas50:
Four checks in `mondoo-oci-security` carried a `# No Terraform variants:` comment claiming the field was "not represented by a verified Terraform resource argument". Three of those comments recorded that nobody had looked, not that nothing was there. Resolved against the `oracle/oci` provider schema, all four fields exist:

| Check | Terraform surface |
|---|---|
| `adb-standby-acl-not-open` | `oci_database_autonomous_database.standby_whitelisted_ips` |
| `genai-endpoint-content-moderation` | `oci_generative_ai_endpoint` → `content_moderation_config` → `is_enabled` |
| `genai-endpoint-private-endpoint` | `oci_generative_ai_endpoint.generative_ai_private_endpoint_id` |
| `datascience-notebook-private-endpoint` | `oci_datascience_notebook_session` → `private_endpoint_id` |

Each gains the conventional `-oci` / `-terraform-hcl` / `-terraform-plan` / `-terraform-state` quad plus a `terraform` remediation snippet. That takes the policy from 72 of 99 checks with Terraform coverage to 76.

The pass/fail fixtures for all four have been in the tree since #3324, each marked `KNOWN_BUG: missing-check` — *"is not present in this bundle yet; it is added by a separate pull request."* This is that pull request; the ten markers are deleted here, since a stale marker fails the build.

### Two variants that needed more than a field read

**The standby ACL variant does not read `standby_whitelisted_ips` alone.** `are_primary_whitelisted_ips_used` decides which list the standby actually enforces: when it is set the standby reuses `whitelisted_ips`, otherwise it enforces `standby_whitelisted_ips`. Reading only the standby list would pass a database that inherits an open primary allowlist, so each branch is checked against the list that governs it. New fixtures `fail/inherits-open-primary` and `pass/inherits-restricted-primary` cover the branch the obvious form would have missed.

**The notebook session variant searches both spellings of the configuration block.** The provider accepts `notebook_session_config_details` and `notebook_session_configuration_details` with identical fields and neither deprecated, so matching only one would pass a session that uses the other. New fixture `pass/private-alias-block`.

### The two that stay declined

`genai-endpoint-prompt-injection-detection` and `genai-endpoint-pii-detection` keep their `# No Terraform variants:` comments, now stated in terms of the schema rather than as an absence of verification: `oci_generative_ai_endpoint` exposes `content_moderation_config` and nothing else, and the provider carries no guardrail resource of any kind.

### Verification

- `cnspec policy lint` — valid. The two `query-deprecated-symbol` warnings are pre-existing and untouched.
- 13 HCL fixture scenarios pass across the four checks.
- **Plan and state variants asserted by hand in both directions.** `TestTerraformVariantCoverage` only gates `-terraform-hcl`, so the plan and state halves ship ungated. Both were checked against documents generated by `terraform plan` + `terraform show -json` against a compliant and a violating config — 16 assertions, all correct.
- `TestTerraformVariantCoverage`: `mondoo-oci-security -terraform-hcl: 76/76 covered (100.0%)`.
- Policy version 4.6.1 → 4.7.0; MINOR, since this is new variants.

Note that the remediation snippets are validated here against the pinned `~> 8.0` constraint; #3623 bumps that to `~> 9.0`, and all four fields are present in both.


tas50:
### CI failures addressed

Both red shards were the same check, and neither was from this PR:

- `terraform (shard 4/12)` — `mondoo-aws-security-no-static-credentials-in-providers-terraform-hcl/pass/no-creds`
- `remediation (shard 4/6)` — `TestRemediationSatisfiesCheck/mondoo-aws-security-no-static-credentials-in-providers-terraform-hcl`

This branch does not touch `content/mondoo-aws-security.mql.yaml` (`git diff origin/main...HEAD -- content/mondoo-aws-security.mql.yaml` is empty), and the check fails standalone against the fixture as committed on `main`:

```
$ mql run terraform hcl . -c '<the check, verbatim from main>'
1 error occurred:
	* dict value does not support field `find`
```

**Root cause.** The block form scores each operand of a `||` as its own assessment:

```
terraform.providers.where( nameLabel == "aws" ) {
  arguments["access_key"] == null || arguments["access_key"].find(...).all("AKIAIOSFODNN7EXAMPLE")
}
```

That one line produced two results — the disjunction, correctly `true`, and the right-hand operand evaluated alone, which errors because an absent argument is a dict-typed null and `find` is a string method. The erroring half is what scored, so **every** provider without a hard-coded key failed, which is why the check's own `pass` fixture and its own remediation snippet both went red. `all()` scores the disjunction as one value, which is what the guard was written to rely on.

**On keeping the regexes.** I did not replace the regex with a plain comparison against the example keys. `access_key = var.aws_access_key` reaches MQL as the literal string `"var.aws_access_key"` — the practice the AWS provider docs recommend — and a bare equality test fails it. The middle operand passes anything containing no key-shaped literal, so interpolated, profile-based and `assume_role` providers pass on their own merits. Added `pass/interpolated-vars`, since nothing covered the shape a naive fix breaks.

Verified with `mql run terraform hcl` against all four fixtures and all three HCL blocks of the check's own Terraform remediation:

| fixture | verdict |
|---|---|
| `pass/no-creds` | passes |
| `pass/interpolated-vars` (new) | passes |
| `fail/static-access-key` | fails |
| `fail/static-secret-key` | fails |

The `-terraform-plan` sibling uses bare comparisons with no disjunction and is unaffected.

It sits in its own commit (cc07555) with an `aws:` scope — happy to split it into its own PR if you would rather this one stayed purely OCI, though this branch stays red until it lands somewhere.

PR #3632, opened 2026-09-08
🐛 content: report unreachable endpoints instead of scoring them clean

chris-rock:
## The bug

A group filter that performs network I/O decides which checks apply by reaching out over the network. When that call fails, the expression **errors** rather than returning empty — and a filter that errors is not a filter that is false. The resolver drops the checks behind it from the resolved policy, so the failure leaves no trace: no checks, no errors, and a passing score.

Verified against a host with nothing listening on the scanned port:

```
cnspec scan host dnssec-failed.org -f content/mondoo-tls-security.mql.yaml
  → 0 checks, LOW (0)

cnspec scan host dnssec-failed.org -f content/mondoo-http-security.mql.yaml
  → 0 checks, LOW (0)
```

A TLS policy answering "clean" for a host that speaks no TLS is the worst available answer, and in the report it is indistinguishable from a host that passed every check. Control: `ftp.gnu.org`, which does listen on 443, gets all 22 TLS checks — the mechanism is the failed connection, not the host being uninteresting.

## The fix

One check per policy, in a group whose filter is `asset.platform == 'host'` and therefore cannot fail, asserting the endpoint answers at all:

| policy | new check | version |
|---|---|---|
| `mondoo-tls-security` | Ensure the host negotiates a TLS connection | 1.8.0 → 1.9.0 |
| `mondoo-http-security` | Ensure the host answers an HTTP request | 1.3.2 → 1.4.0 |

The detail groups keep their connection filters on purpose, so a host that legitimately serves no TLS gets a single row rather than a wall of failures for ciphers and certificates that were never negotiated.

## `mondoo-dns-security`

The same shape reached the four zone-wide checks added in #3579, filtered on `dns.fqdn == dns.zone.fqdn`. `dns.zone` walks the delegation over the network, and dnsshake reports a query failure as an error rather than a null zone — deliberately, so a resolver outage is not read as *this name has no zone*. A timed-out lookup therefore removes those four silently.

The filter now tests the registrable domain first:

```
dns.fqdn == domainName.effectiveTLDPlusOne || dns.fqdn == dns.zone.fqdn
```

For a name whose apex is its registrable domain — every name this policy is normally pointed at — the first arm is true and `||` short-circuits before `dns.zone` is touched. This follows the reasoning already recorded on the HSTS group filter in `mondoo-http-security`, which was joined with `&&` for the same short-circuit reason. A delegated apex such as `cs.cmu.edu` still needs the lookup and stays exposed, so this **shrinks the window rather than closing it**; 1.6.0 → 1.6.1, since nothing changes when lookups succeed.

## Verification

Scan counts before (`main`) and after, per host:

| host | DNS | TLS | HTTP |
|---|---|---|---|
| `mondoo.com` | 8 → 8 | 22 → 23 | 13 → 14 |
| `www.mondoo.com` | 4 → 4 | 22 → 23 | 13 → 14 |
| `cs.cmu.edu` | 8 → 8 | 22 → 23 | 13 → 14 |
| `dnssec-failed.org` | 8 → 8 | **0 → 1** | **0 → 1** |

DNS is unchanged everywhere, which is the intent — the reorder only matters when a lookup fails. The `+1` on reachable hosts is the new check passing. `#3579`'s own results still hold: `www.mondoo.com` scores no zone-wide checks, `cs.cmu.edu` is audited as the zone it is.

Green locally: `cnspec policy lint ./content`, `remediation/code/bash.py` (3600 PASS, 0 FAIL), `go test ./content/validation/compliance/...`, `typos`.

## What this does not fix

The partial case. If a host answers but a detail group's own network call fails transiently, that group still disappears silently and the new check passes — observed once on a healthy host, where one run returned the three HSTS checks and the next did not.

Closing that means teaching the resolver to tell a filter that **errored** from one that was **false**, rather than dropping both. That is an engine change rather than a content change, and it is not attempted here; each policy carries a comment recording the limit at the filter it affects.

### DIFF
--- a/content/mondoo-dns-security.mql.yaml
+++ b/content/mondoo-dns-security.mql.yaml
@@ -3,7 +3,7 @@
 policies:
   - uid: mondoo-dns-security
     name: Mondoo DNS Security
-    version: 1.6.0
+    version: 1.6.1
     license: BUSL-1.1
     tags:
       mondoo.com/category: security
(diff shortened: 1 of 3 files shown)


chris-rock:
merging as those errors are unrelated to this change

PR #3647, opened 2026-09-12
🧹 ci: rename cnquery-update.yml to mql-update.yml

chris-rock:
> [!IMPORTANT]
> **Draft — merge after #3603.** That PR edits `cnquery-update.yml`; a rename landing first turns it into a rename/edit conflict. Once #3603 is in, I'll rebase this and mark it ready.

## Why

The dependency has been `go.mondoo.com/mql` for a while. The filename was the last place in `.github/workflows/` still calling it cnquery — the workflow's own `name:` already reads **Bump mql**.

## Why this is safe to rename

Nothing resolves this workflow by filename:

| Entry point | Keyed on | Affected? |
|---|---|---|
| `repository_dispatch: types: [update-mql]` | event type — mql's `goreleaser.yml` sends `event-type: update-mql` | no |
| `workflow_dispatch` | picked from the Actions UI list | no |

A code search across `mondoohq/cnspec`, `mondoohq/mql` and `mondoohq/installer` finds exactly one other mention, a comment in `validation-dependency-updates.yaml`, updated here.

## Deliberately unchanged

The `version/mql_update_` branch prefix. #3604 matches on it to decide whether a merged PR should be auto-tagged; renaming the prefix in the same change would mean coordinating two PRs for no gain.

## Note

Renaming a workflow file starts a fresh run history for it in the Actions UI — the old runs stay under the old name rather than moving. That's cosmetic, but worth expecting.


chris-rock:
Rebased onto `main` now that #3603 has merged, and marked ready.

Git tracked it as a pure rename with zero content change, so #3603's work carries through untouched — the renamed file is byte-identical to what just landed as `cnquery-update.yml` (`diff` confirms), including the capture-group fix and the `/commits/` tag resolution.

One reference appeared while this sat as a draft: #3646 added a comment in `gh-release.yaml` naming `cnquery-update.yml` as the thing that writes `VERSION`. Updated in the same commit, so the rename is complete again:

```
$ grep -rn "cnquery-update" .
  (no references left anywhere)
```

Unchanged from the original description: nothing resolves this workflow by filename — the entry points are `repository_dispatch: types: [update-mql]` (mql's `goreleaser.yml` sends `event-type: update-mql`) and `workflow_dispatch` from the Actions UI. The `version/mql_update_` branch prefix is deliberately left alone, since #3604 matches on it.

chris-rock:
Rebased onto `main` after #3649. It had gone into conflict — this PR edited a comment in `gh-release.yaml`, and #3649 deleted that file. Resolved by taking the deletion; the edit was moot once the file was gone.

#3604 also landed in between and introduced two fresh references to the old name, one of them in a comment I wrote:

```
auto-tag-after-mql-bump.yml:3:  # Fires when a "🧹 Bump mql to <version>" PR from cnquery-update.yml
auto-tag-after-mql-bump.yml:41: # `v?`, matching cnquery-update.yml, which is what builds this branch
```

Both updated, so the rename is complete again:

```
$ grep -rn "cnquery-update" .
  (no references left anywhere)
```

Net diff is now the rename plus two comment fixes. Unchanged from the original rationale: nothing resolves this workflow by filename — `repository_dispatch: types: [update-mql]` and `workflow_dispatch` are the entry points — and the `version/mql_update_` branch prefix is deliberately untouched, since `auto-tag-after-mql-bump.yml` matches on it.

PR #3879, opened 2026-09-13
📝 openstack: name the swift client for swift container ACL changes

### LINKED ISSUE #3858: mondoo-openstack-security: Swift container checks: desc names nonexistent openstack container set ACL flags
The `desc` of two Swift checks tells readers to change container ACLs with an `openstack container set` flag that doesn't exist.

### The query

Neither check's query is affected. The problem is only in the prose:

- `mondoo-openstack-security-container-not-public`: "The ACL is set with `openstack container set --read-acl` or the `container_read` argument..."
- `mondoo-openstack-security-container-no-wildcard-write`: "The list is set with `openstack container set --write-acl` or the `container_write` argument..."

### Why it fails

OpenStackClient's `container set` only sets `X-Container-Meta-*` properties. A reader who copies the named flag gets an argument error. The remediation methods correctly use `swift post -r` and `swift post -w` from python-swiftclient.

### Evidence

With python-openstackclient 10.3.0:

```
$ openstack container set --help
usage: openstack container set [-h] --property <key=value> <container>
```

With python-swiftclient 4.11.0:

```
$ swift post --help
Usage: swift post [--read-acl <acl>] [--write-acl <acl>] ...
  -r, --read-acl <acl>  Read ACL for containers. ...
  -w, --write-acl <acl> Write ACL for containers. ...
```

### Suggested fix

In both descriptions, name `swift post --read-acl` or `swift post --write-acl` from python-swiftclient instead of `openstack container set`.

Found during the remediation deep dive in #3857.

tas50:
## Summary

Both Swift container ACL checks told readers, in their `desc`, to change the ACL with an `openstack container set` flag that does not exist. This corrects the prose to name `swift post --read-acl` / `swift post --write-acl` from python-swiftclient, and tightens the two audits so the commands and the fields they produce match what the clients actually print.

No query changes. No policy version change.

## Why

`openstack container set` takes only `--property`, and those properties are sent as `X-Container-Meta-*` headers, which cannot carry an ACL.

Verified locally with python-openstackclient 10.3.0:

```
$ openstack container set --help
usage: openstack container set [-h] --property <key=value> <container>

  --property <key=value>
                        Set a property on this container (repeat option to set
                        multiple properties)
```

The implementation confirms the header prefix, in `openstackclient/api/object_store_v1.py`:

```python
headers = self._set_properties(properties, 'X-Container-Meta-%s')
```

Container ACLs are the `X-Container-Read` and `X-Container-Write` headers. python-swiftclient 4.11.0 is what sets them:

```
$ swift post --help
  -r, --read-acl <acl>  Read ACL for containers. ...
  -w, --write-acl <acl> Write ACL for containers. ...
```

Both remediations already used `swift post -r` / `swift post -w` correctly, so only the `desc` sent readers to a command that errors out.

While confirming this, two audit details were also worth correcting:

- `openstack container show` does read the ACLs back. `container_show` copies `x-container-read` into a `read_acl` field and `x-container-write` into `write_acl`, and omits each field when the header is absent. The old audit for the read check said "the container read ACL" without naming the field, so a reader could not tell an absent ACL from a missing field.
- `swift stat` labels the same two values `Read ACL` and `Write ACL`, per `swiftclient/command_helpers.py`.

The remediation command validator never caught this, because it extracts fenced `bash` blocks only. The bad flags were in prose, not in a code block.

## Changes

`mondoo-openstack-security-container-not-public`:

- `desc`: names `swift post --read-acl` and the `X-Container-Read` header, and states that `openstack container set` writes only `X-Container-Meta-*` properties.
- `audit`: the placeholder is now `<container>`, matching the command's own metavar, and the passing and failing results are stated in terms of the `read_acl` field, including the absent-field case. Keeps `swift stat` as the python-swiftclient equivalent.
- `remediation` (cli): says why `openstack container set` cannot do it, rather than just asserting it.

`mondoo-openstack-security-container-no-wildcard-write`:

- `desc`: names `swift post --write-acl` and the `X-Container-Write` header, same correction.
- `audit`: adds `openstack container show <container>` as the OpenStackClient route to the same header, and covers the no-write-ACL-at-all passing case.
- `remediation` (cli): same clarification as above.

## Validation

```
$ python3 content/validation/remediation/commands/validate.py openstack
...
[PASS] mondoo-openstack-security-container-not-public
       openstack container show <container>
[PASS] mondoo-openstack-security-container-no-wildcard-write
       openstack container show <container>
```

134 PASS, 0 FAIL, with python-openstackclient 10.3.0 plus the five plugins the validator declares in `OPENSTACK_PLUGINS` (python-octaviaclient, python-magnumclient, python-troveclient, python-barbicanclient). No plugin was missing.

Also run:

- `cnspec policy lint content/mondoo-openstack-security.mql.yaml` → `valid policy bundle(s)`
- YAML parse of the bundle
- `typos content/mondoo-openstack-security.mql.yaml` → clean

No fixtures were touched: the queries are unchanged, so no IaC variant suite applies.

Fixes #3858


tas50:
Filed #3893 for the validator gap that let this through: `extract_bash_blocks` only picks up fenced ```bash blocks, so the 830 inline command mentions in `desc:` and `audit:` prose across `content/*.mql.yaml` are never checked. That is why `validate.py openstack` reported 134 PASS / 0 FAIL both before and after this fix.

PR #3898, opened 2026-09-13
🐛 okta: read password recovery methods from policy rules on identity engine

### LINKED ISSUE #3855: mondoo-okta-security: password recovery checks: Identity Engine sets recovery in rules, not policy settings
### The query

```mql
okta.policies.password.all( settings['recovery']['factors']['okta_email']['status'] == "ACTIVE" )
okta.policies.password.all( settings['recovery']['factors']['okta_sms']['status'] == "INACTIVE" )
okta.policies.password.all( settings['recovery']['factors']['recovery_question']['status'] == "INACTIVE" )
```

These are the `-api` variants of `password-settings-password-email-recovery`, `-password-sms-recovery`, and `-password-question-recovery`.

### Why it fails

On Okta Identity Engine, recovery methods are chosen in each password policy's rules (`actions.selfServicePasswordReset`), under **Users can initiate recovery with** and **Additional verification is**, not in the policy's `settings.recovery`. If Identity Engine does not keep `settings.recovery.factors` in step with the rules, these checks report on a setting the admin cannot change from the console, and the remediation cannot close them.

This is a question, not a confirmed defect.

### Evidence

- Okta SDK model `PasswordPolicyRecoverySettings`: "With Identity Engine, you can specify recovery factors inside the password policy rule instead of in the policy settings object. Recovery factors for the rule are defined inside the `selfServicePasswordReset` action."
- Okta help "Configure the password authenticator" (Identity Engine): recovery is configured with **Add Rule**, not in the password settings.

Not confirmed: what `settings.recovery.factors` returns on an Identity Engine tenant before and after changing a rule's recovery methods.

### Suggested fix

Against an Identity Engine tenant, change a password policy rule's recovery methods and compare `GET /api/v1/policies?type=PASSWORD` with `GET /api/v1/policies/{id}/rules` before and after. If the policy object does not change, the Identity Engine case needs to read the rules, which likely needs a typed field in the okta provider.

Found during the remediation deep dive in #3850.

tas50:
## Summary

The three password recovery checks (`-api` variants of `password-settings-password-email-recovery`, `-password-sms-recovery`, and `-password-question-recovery`) read `settings.recovery.factors` on the password policy. On Okta Identity Engine the recovery methods live in the policy's **rules** instead. Each variant now reads the rules when they carry Identity Engine recovery requirements, and keeps reading the settings object otherwise.

## Why

Okta's own API spec says so. From the `PasswordPolicyRecoverySettings` schema in the Okta management OpenAPI document, vendored at `okta-sdk-golang/v6@v6.1.7/okta/model_password_policy_recovery_settings.go:33`:

> **Note:** With Identity Engine, you can specify recovery factors inside the password policy rule instead of in the policy settings object. Recovery factors for the rule are defined inside the `selfServicePasswordReset` action.

The rule shape is in the same spec (`okta/api/openapi.yaml:78673-78735`):

- `PasswordPolicyRuleActions.selfServicePasswordReset` is a `SelfServicePasswordResetAction` with `access` and `requirement`.
- `SsprRequirement` is marked `<x-lifecycle class="oie">` and carries `accessControl` (`AUTH_POLICY` or `LEGACY`), `primary`, and `stepUp`.
- `SsprPrimaryRequirement.methods` is an enum of `push`, `sms`, `voice`, `email`, `otp`.
- `SsprStepUpRequirement.methods` is an enum whose only member is `security_question`, paired with `required`.

So on an Identity Engine org that configures recovery in its rules, the old queries score `settings.recovery.factors`, which is not the object the Admin Console edits and not the object the remediation steps move. The console remediation on all three checks already tells Identity Engine admins to change the rule, so the check could not be closed by following its own instructions.

No provider addition was needed. `okta.policy.rules` and `okta.policyRule.actions` already exist in the installed okta provider (`providers/okta/resources/okta.lr:1147` and `:1217`), and `okta.policies.signOn` already reads `_.rules` at lines 852 and 853 of this policy.

The rule-driven assertion deliberately skips a rule whose `requirement.accessControl` is `AUTH_POLICY`. Such a rule defers its authenticator requirements to the Okta account management policy and carries no `primary` or `stepUp` of its own, so failing it would be a false positive.

## Changes

For each of the three checks:

- `mql` is now two statements. The first scores `settings.recovery.factors` but only over policies whose rules carry no `selfServicePasswordReset.requirement`, which is every Classic Engine policy. The second scores the rules of every policy, over the rules that allow self-service password reset and list their own methods:
  - email: `requirement.primary.methods` must contain `email`
  - sms: `requirement.primary.methods` must not contain `sms`
  - question: `requirement.stepUp.methods` must not contain `security_question`
- `desc` now says where each engine keeps its recovery methods and what the check asserts in each case.
- `audit` gains Identity Engine console steps, the `GET /api/v1/policies/{id}/rules` call, and a passing-response description for both shapes.

No `version:` line changed. The Terraform variants and the remediation blocks are untouched.

## Validation

- `cnspec policy lint content/mondoo-okta-security.mql.yaml` → `valid policy bundle(s)`. The lint compiles the MQL: renaming `rules` to `rulez` in a scratch copy produced `cannot find field or resource 'rulez' in block for type 'okta.policy'`, so `rules` and the nested `actions` binder both resolve.
- No Okta org is available, so the queries were exercised against literal API-shaped data with `cnquery run` and `parse.json` over five policy fixtures: `classic-good`, `classic-bad`, `oie-good` (rules offer `email`, `push`; stale settings say email `INACTIVE`, sms `ACTIVE`, question `ACTIVE`), `oie-bad` (rules offer `sms` and require `security_question`; stale settings look clean), and `oie-authpolicy` (`accessControl: AUTH_POLICY`).

  | fixture | settings statement | rules statement (email / sms / question) |
  |---|---|---|
  | classic-good | in partition, passes | vacuous pass / pass / pass |
  | classic-bad | in partition, fails | vacuous pass / pass / pass |
  | oie-good | excluded from partition | ok / ok / ok |
  | oie-bad | excluded from partition | failed / failed / failed |
  | oie-authpolicy | excluded from partition | pass / pass / pass |

  On `origin/main`'s query, `oie-good` fails all three checks and `oie-bad` passes all three, which is the defect in both directions.
- `python3 content/validation/remediation/commands/validate.py okta` → `32 passed, 0 failed`, including the two new `GET /api/v1/policies/$POLICY_ID/rules` audit calls.
- `typos content/mondoo-okta-security.mql.yaml` → clean. YAML parses.

Fixes #3855


tas50:
Filed #3904 for a gap this change does not close: a password policy whose every self-service reset rule defers to the Okta account management policy with `requirement.accessControl: AUTH_POLICY` carries no methods of its own, so it is excluded from both halves of the query and passes all three checks silently. Closing that needs correlation with the account management policy rather than a provider addition.

PR #3919, opened 2026-09-13
🧹 repo: rename CLAUDE.md to AGENTS.md, and record the ADR convention

chris-rock:
Implements [ADR 0002](https://github.com/mondoohq/.github/blob/main/docs/adr/0002-agent-instructions-in-agents-md.md), accepted in mondoohq/.github#59.

## The rename

`AGENTS.md` is the [cross-tool convention](https://agents.md/) for repository agent instructions. The content here — where things live, how to build and test, which idioms are deliberate — is not tool-specific, and a tool-specific filename leads to a second copy for the second tool.

Three files move, and `CLAUDE.md` becomes a **relative symlink** beside each so nothing reading the old name breaks:

```
AGENTS.md              file        CLAUDE.md          -> AGENTS.md
content/AGENTS.md      file        content/CLAUDE.md  -> AGENTS.md
policy/AGENTS.md       file        policy/CLAUDE.md   -> AGENTS.md
```

Nested files are the sanctioned monorepo pattern, not a local invention: the spec says agents read the nearest file and the closest takes precedence, which already matches how this repo is organised.

Every cross-reference is updated — including the GitHub URLs in `skills/mql/SKILL.md`, the drift-check message in `validation-upstream-drift.yaml`, and the assertion in `reports/ocsf/deps_test.go`.

`agents/AGENTS.md` is unrelated and untouched: a generated skills manifest written by `scripts/generate-agents`.

## The ADR convention

`AGENTS.md` gains a section recording how ADRs are written here, which was documented nowhere: `docs/adr/`, one numbered file per decision, following the [org template](https://github.com/mondoohq/.github/blob/main/docs/adr/adr-template.md).

The part worth writing down is that **Security and Performance sections are required** and easy to drop by accident — `"None — this decision does not change the threat model"` plus the reason is a complete answer; an absent section is not.

Also noted: the log is append-only, and `docs/adr/0001` through `0006` here predate the org template and use an older `# ADR-NNNN:` heading. They are left alone rather than retrofitted.

## Known cost

Git for Windows does not create symlinks by default. Without `core.symlinks=true` and Developer Mode, `CLAUDE.md` checks out as a text file containing the string `AGENTS.md`, so a tool reading it gets one useless line. Nothing errors — which is why it is recorded in the ADR's Negative section rather than left to be discovered.


chris-rock:
Unrelated error, I am merging it as a fix for lint is already in work

PR #3691, opened 2026-09-13
🐛 gcp: fix remediation steps that would not close their checks

tas50:
## Summary

Audited every remediation method (console, cli, terraform, and REST examples) on the 297 checks with a `remediation:` block in `mondoo-gcp-security`, reading each check's query and its variants' queries first and verifying commands against gcloud `--help` (565.0.0 and the current 584.0.0), bq, the Terraform provider schema, the mql GCP provider source, and Google Cloud docs. 180 edits on 141 checks fix commands that gcloud rejects as written (missing required flags, wrong enum case, mutually exclusive flags, invalid REST bodies, removed images), steps that leave the flagged resource in place or skip a required service agent key grant, disruptive steps with no warning (DNSSEC DS records, node recreation, `functions deploy` without `--source`), and placeholders that were never explained. Three GCP runtime queries are also fixed because they could not pass for the documented fix (OS Login, block project-wide SSH keys) or passed an insecure configuration (default service account with full API scope).

## Changes

Each bullet says what was wrong, then what changed and why that makes the instructions correct or the check pass. Two groups, **Fix would not close the check** and **Values, placeholders, and why not explained**, are in the first comment on this PR because the full list does not fit in the body. The evidence behind every edit (the `--help` output, doc quote, or provider source it was checked against) is in the evidence comments.

- **Markdown formatting:** long remediation paragraphs that mixed a fix, a procedure, and a warning are now numbered steps, bullet lists of placeholders, flags, and service agents, and warning paragraphs that open with a bold sentence. Wording and meaning are unchanged.

### Queries that could not pass for the documented fix, or passed an insecure configuration

<details><summary>3 edits on 3 checks</summary>

- **Ensure instances do not use default service account with full API access** (`mondoo-gcp-security-instances-not-configured-with-default-service-account-full-access-cloud-api`, mql): The GCP variant compares the whole scopes list to a single string, which is true only when cloud-platform is the sole scope, so a default service account holding cloud-platform plus any other scope passes even though it has full API access. Testing list membership with contains fails every default-account instance whose scopes include cloud-platform, matching the plan and state variants that already use scopes.contains, and keeps passing instances whose remediated scopes omit it.
[...]

Issue #3923, opened 2026-09-13
mondoo-gcp-security: cloud-dns-rsasha1-zsk-not-used-gcp reads dnsKeys, which no released gcp provider has, so Lint Policies fails on main


tas50:
The `Lint Policies` workflow has been failing on `main` since 13:29 UTC today. Two queries in `content/mondoo-gcp-security.mql.yaml` read gcp provider fields that exist only in a locally built provider, not in the released one the lint action downloads. One of them is `mondoo-gcp-security-cloud-dns-rsasha1-zsk-not-used-gcp`.

### The error

From the `lint` job on `main` (run 34770318366):

```
Could not compile policy bundle: failed to validate query
'//local.cnspec.io/run/local-execution/queries/mondoo-gcp-security-cloud-dns-rsasha1-zsk-not-used-gcp':
failed to compile query '...
gcp.dns.managedzone.dnsKeys.where(type == 'zoneSigning').all(algorithm != 'rsasha1' && algorithm != 'rsasha1-nsec3-sha1')
': cannot find field 'dnsKeys' in gcp.project.dnsService.managedzone
```

The query came in with #3905.

### Why it fails

`dnsKeys` is not in any released gcp provider. Checked against the exact schema the lint action resolves:

| Provider build | `gcp.project.dnsService.managedzone.dnsKeys` |
|---|---|
| released `gcp` 13.37.1 (`https://releases.mondoo.com/providers/gcp/13.37.1/schema.json`, the current `stable` per `latest.json`) | absent |
| locally built `gcp` 13.37.0 from `mondoohq/mql` `main` | present |

Reproduced locally by staging the released provider ahead of the installed one:

```bash
curl -sL -o gcp.tar.xz https://releases.mondoo.com/providers/gcp/13.37.1/gcp_13.37.1_darwin_arm64.tar.xz
mkdir -p staging/gcp && tar -xJf gcp.tar.xz -C staging/gcp
PROVIDERS_PATH="$PWD/staging:$HOME/.config/mondoo/providers" \
  cnspec policy lint content/mondoo-gcp-security.mql.yaml
```

That reports exactly the CI error and nothing else once the companion GKE defect is fixed.

### Options

1. Rewrite the query against fields the released provider already has. `gcp.project.dnsService.managedzone.dnssecConfig` is a dict and is present in 13.37.1, so a form built on it compiles today, at the cost of auditing key generation defaults rather than the live signing keys, which is the very thing #3790 set out to fix.
[...]

PR #3925, opened 2026-09-13
🐛 gcp: pass Cloud SQL MySQL instances that report no IP addresses

### LINKED ISSUE #3780: mondoo-gcp-security: cloud-sql-mysql-instances-not-publicly-exposed: fails instances with no IP addresses that the PostgreSQL and SQL Server checks pass
### The query

```mql
# mondoo-gcp-security-cloud-sql-mysql-instances-not-publicly-exposed-gcp
gcp.sql.instance.ipAddresses != empty &&
gcp.sql.instance.ipAddresses.all(type != "PRIMARY")
```

The PostgreSQL and SQL Server versions of the same control use only the second half:

```mql
# mondoo-gcp-security-cloud-sql-postgres-instances-not-publicly-exposed-gcp
# mondoo-gcp-security-cloud-sql-sql-server-instance-not-publicly-exposed-gcp
gcp.sql.instance.ipAddresses.all(type != "PRIMARY")
```

### Why it fails

An instance with no IP addresses at all has no public address, so it meets the control. The MySQL check fails it because of the extra `ipAddresses != empty` guard, while the PostgreSQL and SQL Server checks pass the same configuration. The three engines give different verdicts for the same network setup.

### Evidence

- Query text of the three `-gcp` variants on `main`, quoted above.

Not confirmed: which real configurations return an empty `ipAddresses` list. A Private Service Connect only instance with public IP and private services access both off is the likely case, but no live instance was scanned.

### Suggested fix

Drop the `ipAddresses != empty` line from the MySQL variant so it matches its siblings, or, if an empty list is meant to fail, add the same guard to the PostgreSQL and SQL Server variants and document why.

Found during the remediation deep dive in #3691.

tas50:
## Summary

The GCP runtime variant of `mondoo-gcp-security-cloud-sql-mysql-instances-not-publicly-exposed` failed any Cloud SQL for MySQL instance that reports no IP addresses at all, while the PostgreSQL and SQL Server variants of the same control passed it. This drops the extra guard so all three engines agree, and updates the check's description and audit step, which documented the old behavior.

## Why

The three `-gcp` variants on `main`:

```mql
# mondoo-gcp-security-cloud-sql-mysql-instances-not-publicly-exposed-gcp
gcp.sql.instance.ipAddresses != empty &&
gcp.sql.instance.ipAddresses.all(type != "PRIMARY")

# mondoo-gcp-security-cloud-sql-postgres-instances-not-publicly-exposed-gcp
# mondoo-gcp-security-cloud-sql-sql-server-instance-not-publicly-exposed-gcp
gcp.sql.instance.ipAddresses.all(type != "PRIMARY")
```

`PRIMARY` is the only public address type. The Cloud SQL Admin API documents `ipAddresses` as "The assigned IP addresses for the instance", and the `SqlIpAddressType` enum as:

| Type | Meaning |
|---|---|
| `PRIMARY` | "IP address the customer is supposed to connect to. Usually this is the load balancer's IP address" |
| `OUTGOING` | source address a replica uses to reach an external primary |
| `PRIVATE` | "Private IP used when using private IPs and network peering" |
| `MIGRATED_1ST_GEN` | V1 address of a migrated instance |

So an instance with no `ipAddresses` entries has no `PRIMARY` entry and therefore no public IP. It satisfies the control, and the MySQL guard failed it anyway.

That empty shape is not hypothetical. It is what a Private Service Connect instance looks like. Google's own PSC setup guide creates one with `--no-assign-ip --enable-private-service-connect`, equivalently `"ipv4Enabled": false` over REST, and the sample `gcloud sql instances describe` output for such an instance shows no `ipAddresses` field. The address lives under `settings.ipConfiguration.pscConfig.pscAutoConnections.ipAddress` instead. PSC is the most private connectivity option Cloud SQL offers, and the check was failing it on MySQL while passing it on PostgreSQL and SQL Server.

The guard was added in #2535 with no recorded rationale beyond "ipAddresses != empty guard", and the description sentence asserting the behavior was written later, in the prose pass, to match the query rather than the other way round.

## Changes

`content/mondoo-gcp-security.mql.yaml`, inside `mondoo-gcp-security-cloud-sql-mysql-instances-not-publicly-exposed` only:

- `...-gcp`: dropped the `gcp.sql.instance.ipAddresses != empty` line, leaving the same single assertion the PostgreSQL and SQL Server variants use.
- `docs.desc`: the closing paragraph said the check "fails an instance that reports no addresses at all rather than passing it silently". Replaced with what the query now asserts, naming the PSC case.
- `docs.audit` step 3: it told the auditor to verify a `type: PRIVATE` entry exists, which the query has never required. Rewritten around the absence of `type: PRIMARY`, and it now says a PSC instance shows no `ipAddresses` entries at all.

No `version:` change. No Terraform variant touched: they assert `ipv4_enabled == false` and were never affected.

## Validation

Verdicts on literal `ipAddresses` data via `mql run -c` with `parse.json`, old query and new query against the three shapes:

| `ipAddresses` | old query | new query |
|---|---|---|
| `[]` (PSC-only instance) | `[failed] false && _` | `[ok] value: true` |
| `[{"type":"PRIVATE"}]` | `[ok] true` | `[ok] value: true` |
| `[{"type":"PRIMARY"},{"type":"PRIVATE"}]` | `[failed] true && false` | `[failed]` |

The only verdict that moves is the empty one. A public address still fails.

- `cnspec policy lint content/mondoo-gcp-security.mql.yaml` reports `valid policy bundle(s)`. The 19 `query-deprecated-symbol` warnings are all pre-existing and on unrelated checks.
- `python3 content/validation/remediation/commands/validate.py gcp` passes, exit 0, no FAIL lines.
- `typos content/mondoo-gcp-security.mql.yaml` clean; YAML parses.
- No IaC fixtures affected: the changed variant is the GCP runtime one, and the Terraform variants are untouched. No `KNOWN_BUG.md` marker references this check.

## Overlapping PRs

#3691, the GCP remediation deep dive, is open against this same file and is where this issue was found. Several sibling PRs are also editing `content/mondoo-gcp-security.mql.yaml` right now; this change is confined to the three lines above.

Fixes #3780


tas50:
CI note: the `lint` check is red, but not because of this change.

`Lint Policies` fails with `bundle-compile-error` on two queries this PR does not touch:

```
cannot find field 'dnsKeys' in gcp.project.dnsService.managedzone
cannot find field 'networkPolicyAddon' in gcp.project.gkeService.cluster.addonsConfig
```

Both fields exist in mql `main` but not in the released gcp provider that the lint job's `mondoo/cnspec:13` container resolves against. The same two errors fail `Lint Policies` on `main` in runs that predate this merge: 34770841362 (`572a4561d`, 17:11Z), 34770944372 (`df797bcc0`, 17:13Z), 34771332782 (`f3ec15fe0`, 17:21Z).

Tracked in #3923 (`dnsKeys`) and #3936 (`networkPolicyAddon`). Both have to clear before the job goes green.

Everything else passed: 52 checks green, and `Mondoo Code Review` approved.

PR #3926, opened 2026-09-13
🐛 digitalocean: fail world-open port ranges that contain SSH, RDP, and database ports

### LINKED ISSUE #3847: mondoo-digitalocean-security: SSH, RDP, and database-port firewall checks: world-open port ranges pass
### The query

`mondoo-digitalocean-security-firewall-no-unrestricted-ssh-digitalocean`:

```coffee
digitalocean.firewall.ingressRules.where(openToInternet).none(protocol == "tcp" && ports == "22")
```

The RDP check uses `ports == "3389"`, and the database-port check uses `ports.in(["3306", "5432", "27017", "6379", "9092", "9200"])`. The HCL, plan, and state variants compare `port_range` the same way, for example `arguments["port_range"] == "22"` and `/^(3306|5432|27017|6379|9092|9200)$/`.

### Why it fails

A Cloud Firewall rule's `ports` can be a range, and every variant matches only a rule for exactly that port. A rule that admits SSH from `0.0.0.0/0` through a range passes all three checks:

| Rule | SSH check | Actually exposes 22 |
|---|---|---|
| `protocol:tcp,ports:22,address:0.0.0.0/0` | fails | yes |
| `protocol:tcp,ports:20-30,address:0.0.0.0/0` | **passes** | yes |
| `protocol:tcp,ports:1-1024,address:0.0.0.0/0` | **passes** | yes |

The full `0`, `all`, and `1-65535` ranges are caught by `firewall-no-all-ports-open`, but any narrower range that contains the port is missed.

### Evidence

- The OpenAPI `firewall_rule` model: ports are "a string containing a single port, a range (e.g. "8000-9000"), or "0" when all ports are open for a protocol."
- The `doctl compute firewall` help: "The `ports` attribute may contain a single port, a range of ports (e.g. `8000-9000`), or `all`".
- Verified by reading the queries. Not run against a live firewall.

### Suggested fix

Match ranges that contain the port. MQL has no string-to-int conversion in these contexts, so either add a provider helper on `digitalocean.firewall.ingressRule`, for example a `portRange` start and end or `coversPort(22)`, or split `ports` on `-` and compare the bounds where the language allows it. Apply the same logic to the Terraform variants' `port_range`.

Found during the remediation deep dive in #3839.

tas50:
## Summary

The SSH, RDP and database-port Cloud Firewall checks compared a rule's port field for equality against the exact port, so a world-open rule that reached the port through a range passed. All twelve queries, runtime plus HCL, plan and state, now match any port string that covers the port.

## Why

`digitalocean.firewall.ingressRule.ports` is a plain string. The provider schema documents the three forms it takes:

```
// Affected port or range
//
// A single port ("3306"), a range ("8000-9000"), or "0"/"all" for
// every port.
ports string
```
`providers/digitalocean/resources/digitalocean.lr:370-374` in mondoohq/mql. `doctl compute firewall` and the OpenAPI `firewall_rule` model describe the same three forms, and the Terraform `digitalocean_firewall` `port_range` attribute is the same string.

The old queries matched only the exact-port form, so a rule admitting SSH from `0.0.0.0/0` through `20-30` or `1-1024` passed. Proven by running the new fail fixtures against the queries on `origin/main`: all eighteen of them passed a check they should fail.

```
--- FAIL: .../firewall-no-unrestricted-ssh-terraform-hcl/fail/ssh-range-ipv4
--- FAIL: .../firewall-no-unrestricted-ssh-terraform-hcl/fail/ssh-range-ends-on-port
--- FAIL: .../firewall-no-unrestricted-ssh-terraform-hcl/fail/ssh-range-single-port
--- FAIL: .../firewall-no-unrestricted-ssh-terraform-hcl/fail/ssh-range-straddles-port
--- FAIL: .../firewall-no-unrestricted-ssh-terraform-hcl/fail/ssh-range-ipv6
--- FAIL: .../firewall-no-unrestricted-ssh-terraform-hcl/fail/ssh-all-ports
(same six shapes for rdp and db-ports)
```

MQL has no string-to-int conversion, so the containment test is encoded as an anchored regex with two alternations, a lower bound at or below the port and an upper bound at or above it. For SSH:

```
/^(22|all|0|(\d|1\d|2[0-2])-(2[2-9]|[3-9]\d|\d{3,5}))$/
```

The database check carries one such pair per port. The upper bounds accept up to five digits, which over-matches above 65535; DigitalOcean rejects those ports on the way in.

The source side needed no change. `openToInternet` compares each source address to `0.0.0.0/0` and `::/0` exactly, and `.contains("0.0.0.0/0")` on a Terraform `source_addresses` list is element equality, not a substring test, so `10.0.0.0/8` is not a false positive:

```
$ mql run -c 'parse.json("t.json").params["a"].contains("0.0.0.0/0")'   # a = ["10.0.0.0/8"]
[failed] true && false
$ mql run -c 'parse.json("t.json").params["b"].contains("0.0.0.0/0")'   # b = ["0.0.0.0/0"]
[ok] true
```

## Changes

- `mondoo-digitalocean-security-firewall-no-unrestricted-ssh`, `-rdp` and `-db-ports`: the runtime, Terraform HCL, plan and state queries now match an exact port, a range containing the port, and `0` or `all`.
- Parent `docs.desc`, `docs.audit` and both remediation descriptions on the three checks now say the check fails a covering range and an all-ports rule, not only an exact port rule. No remediation command changed.
- Eighteen new fail fixtures and three new pass fixtures on the HCL variants, covering an exact port, a range ending on the port, a single-port range, a straddling range, a wide range, `all`, and ranges that sit just below and just above the port.

## Validation

- Each bound of every regex verified exhaustively over 0-65535 in Python: for each of the eight ports the lower half matches exactly `p <= port` and the upper half exactly `p >= port`, zero mismatches. Single-port strings match only the target ports and `0`.
- Boundaries re-verified through the MQL engine itself with `mql run -c '"<s>" == /<re>/'`. For SSH: `22`, `22-22`, `1-22`, `21-23`, `20-30`, `1-65535`, `all`, `0` match; `21`, `23`, `3-9`, `220-230`, `2200` do not. RDP and the six database ports checked the same way, including `9092-9200` matching and `9093-9199` not.
- `go test -tags iac_variants ./content/validation/scans -run 'TestTerraformVariants/mondoo-digitalocean-security/mondoo-digitalocean-security-firewall-no-unrestricted-{ssh,rdp,db-ports}-terraform-hcl/'` and the `-no-all-ports-open-terraform-hcl` suite: all pass.
- Mutation check: with the fixtures in place and the policy file reverted to `origin/main`, all eighteen new fail fixtures fail the suite, which is the bug.
- Plan and state variants are not fixture-gated, so they were run by hand against scratch `plan.json` and `state.json` files. A `20-30` rule from `0.0.0.0/0` fails the SSH plan query and a `5000-6000` rule from `::/0` fails the database state query, while ranges that sit between the ports pass.
- `cnspec policy lint content/mondoo-digitalocean-security.mql.yaml`: valid. The one warning is the pre-existing `digitalocean.spacesKey.grants` deprecation, untouched here.
- `make test/content/iac/coverage`: `mondoo-digitalocean-security -terraform-hcl: 18/18 covered (100.0%)`.
- `TestRemediationSatisfiesCheck` for the three HCL variants: pass.
- `python3 content/validation/remediation/commands/validate.py digitalocean`: 146 passed, 0 failed.
- `typos` on the policy file: clean.
- The runtime `digitalocean-firewall` queries could not be scanned against a live account. They were compile-checked by lint and their regexes exercised directly in the MQL engine.

The policy `version` is unchanged.

Fixes #3847


tas50:
The `lint` check failure is pre-existing on `main` and unrelated to this change. It is a bundle compile error in `mondoo-gcp-security`:

- `mondoo-gcp-security-cloud-dns-rsasha1-zsk-not-used-gcp`: `cannot find field 'dnsKeys' in gcp.project.dnsService.managedzone`
- `mondoo-gcp-security-gke-cluster-network-policy-enabled-gcp`: `cannot find field 'networkPolicyAddon' in gcp.project.gkeService.cluster.addonsConfig`

The same job fails on `main` at `1ebfd600`, the commit this branch is based on, and on the four `main` commits after it. This PR touches only `content/mondoo-digitalocean-security.mql.yaml` and DigitalOcean fixtures; `cnspec policy lint content/mondoo-digitalocean-security.mql.yaml` reports a valid bundle.

PR #3953, opened 2026-09-13
📝 gcp: correct the detached pub/sub subscription docs

### LINKED ISSUE #3946: mondoo-gcp-security: pubsub-subscription-not-detached: desc blames a seek for detachment
### The text

`content/mondoo-gcp-security.mql.yaml`, `mondoo-gcp-security-pubsub-subscription-not-detached`, in `docs.desc`:

> Investigate any detached subscription to determine whether it resulted from an accidental `seek` or a deliberate detachment, then either restore delivery or delete the subscription if it is genuinely no longer needed.

### Why it is wrong

A `seek` cannot detach a subscription, so it is not one of the two causes a reader should be weighing. The two operations are unrelated:

- `projects.subscriptions.seek` in the Pub/Sub v1 discovery document: "Seeks an existing subscription to a point in time or to a given snapshot [...] you can set the acknowledgment state of messages in an existing subscription to the state captured by a snapshot." It moves the acknowledgment cursor and leaves the topic attachment alone.
- `projects.subscriptions.detach` is the only method that produces the state this check fails on: "Detaches a subscription from this topic. All messages retained in the subscription are dropped."

The sentence sends an operator investigating a real finding to look at snapshot and replay history, which will never explain the detachment.

### Suggested fix

Point the investigation at the operation that actually causes it. The detach call is recorded in Cloud Audit Logs, which the next sentence of the same paragraph already recommends alerting on, so the two can be merged:

> Investigate any detached subscription to find the `Detach` call in Cloud Audit Logs and establish whether it was deliberate, then either recreate the subscription or leave it removed if it is genuinely no longer needed.

Not touched in #3945, which only retires the check's Terraform variants.

tas50:
## Summary

Docs-only fix to `mondoo-gcp-security-pubsub-subscription-not-detached`. The `desc` told operators to weigh an accidental `seek` against a deliberate detachment, and described the post-detach state backwards. No query, filter, tag, or version changed.

## Why

**A `seek` cannot detach a subscription.** The two operations are unrelated, per the Pub/Sub v1 discovery document:

- `pubsub.projects.subscriptions.seek`: "Seeks an existing subscription to a point in time or to a given snapshot [...] you can set the acknowledgment state of messages in an existing subscription to the state captured by a snapshot." It moves the acknowledgment cursor and leaves the topic attachment alone.
- `pubsub.projects.subscriptions.detach`: "Detaches a subscription from this topic. All messages retained in the subscription are dropped. Subsequent `Pull` and `StreamingPull` requests will return FAILED_PRECONDITION. If the subscription is a push subscription, pushes to the endpoint will stop."

Detach is the only method that produces the state this check fails on, so an operator sent to look at snapshot and replay history will never find the cause.

**Two adjacent sentences in the same paragraph block were wrong for the same reason**, and fixing only the `seek` sentence would have left the reader with a contradictory picture:

- "An application reading from a detached subscription keeps processing its existing backlog normally, then falls silent once the backlog drains." Pull and streaming pull return `FAILED_PRECONDITION` immediately. There is no backlog left to drain, and a pull subscriber gets a hard error rather than silence. Silence is the push case.
- "Meanwhile the subscription continues to hold storage for whatever backlog it still has." [Detach subscriptions](https://cloud.google.com/pubsub/docs/detach-subscriptions): "After the Pub/Sub service detaches the subscription from the topic, the Pub/Sub service deletes any messages that it retains for the subscription. You can't retrieve these messages from the subscription or reattach the subscription to a topic. To free up your Google Cloud project quota, delete the subscription." The messages are gone; what remains is the quota the empty subscription still consumes.
- Both remediation blocks carried the same mistake, warning the operator to confirm consumers had drained the backlog before deleting. For a detached subscription that backlog no longer exists, so the warning asks for a step that cannot be taken and implies the deletion is riskier than it is.

The `desc`'s existing statement that a detached subscription cannot be reattached is confirmed by the same doc sentence, and its own wording of "restore delivery" contradicted it. That is now "recreate the subscription".

## Changes

`content/mondoo-gcp-security.mql.yaml`, inside `mondoo-gcp-security-pubsub-subscription-not-detached` only:

- `docs.desc`: the investigation sentence now points at the detach call in Cloud Audit Logs instead of an accidental `seek`, and merges with the alerting sentence that already recommended those logs.
- `docs.desc`: the consumer paragraph now describes the `FAILED_PRECONDITION` on pull and the silent stop on push.
- `docs.desc`: the storage paragraph now says the retained messages are deleted and the empty subscription keeps consuming project quota.
- `docs.desc`: the recovery paragraph now explains that the replacement starts with no backlog, replacing the drain-the-backlog-first advice.
- `remediation.console` and `remediation.cli`: same correction to the backlog warning.

#3945 retired this check's Terraform variants; the `desc` never mentioned them, so nothing there needed updating.

## Validation

- `python3 -c "import yaml; yaml.safe_load(...)"` parses the bundle, 1323 queries.
- `PROVIDERS_PATH=<released gcp schema> cnspec policy lint content/mondoo-gcp-security.mql.yaml` reports exactly two errors, `cannot find field 'dnsKeys'` and `cannot find field 'networkPolicyAddon'`. Both reproduce identically on an unmodified `origin/main` copy of the file and are the inherited breakage #3948 is fixing. Nothing new from this branch, which is docs-only.
- `python3 content/validation/remediation/commands/validate.py gcp` exits 0.
- `typos content/mondoo-gcp-security.mql.yaml` is clean.
- API claims checked against the live Pub/Sub v1 discovery document at `https://www.googleapis.com/discovery/v1/apis/pubsub/v1/rest` and the current [Detach subscriptions](https://cloud.google.com/pubsub/docs/detach-subscriptions) page, both fetched 2026-09-13.

Fixes #3946


tas50:
CI is complete: 60 of 64 checks green, including every content suite (terraform, cloudformation, bicep, kubernetes, remediation, coverage, validate-cli, typos).

The four red checks are `lint`, `go-test`, `go-mod`, and the `Test Results` aggregator that reports `go-test`. All four fail identically on `bff4ea909`, the `main` commit this branch is based on, and they are the only non-success checks there too:

- `lint` is the inherited `dnsKeys` / `networkPolicyAddon` compile breakage that #3948 fixes. Neither field is in a check this PR touches.
- `go-mod` is `go mod tidy` drift on `main`: the tidy adds `go.mondoo.com/cnspec/v13 v13.38.1` and `go.mondoo.com/mql/v13 v13.38.1` and drops two indirect entries.

This PR changes only prose inside one check's `docs` and `remediation`, so none of them are attributable to it.

PR #3955, opened 2026-09-13
🐛 gcp: fail the GKE security posture check on an unset posture mode

### LINKED ISSUE #3922: mondoo-gcp-security: gke-security-posture-enabled passes on the UNSPECIFIED enum, so an unset vulnerabilityMode reports as scanning
Found while verifying #3450. That issue's premise (a resource-name collision) no longer holds, but one of the four checks it names has a separate, real vacuous pass underneath it.

## The check

`mondoo-gcp-security-gke-security-posture-enabled-gcp`:

```mql
gcp.project.gkeService.cluster.securityPostureConfig.mode != "DISABLED"
gcp.project.gkeService.cluster.securityPostureConfig.vulnerabilityMode != "VULNERABILITY_DISABLED"
```

Both assertions are written as "not the disabled value". The GKE API has three states per field, not two, and the third one passes.

## Why the unset state passes

The provider builds the sub-resource unconditionally, with nil-safe getters, in `providers/gcp/resources/gke.go`:

```go
secPostureConfig, err := CreateResource(g.MqlRuntime, "gcp.project.gkeService.cluster.securityPostureConfig", map[string]*llx.RawData{
    "id":                llx.StringData(c.Id + "/securityPostureConfig"),
    "mode":              llx.StringData(c.GetSecurityPostureConfig().GetMode().String()),
    "vulnerabilityMode": llx.StringData(c.GetSecurityPostureConfig().GetVulnerabilityMode().String()),
```

`GetSecurityPostureConfig()` returns a nil receiver when the API omits the message, and the generated getters then return the enum zero value. From `cloud.google.com/go/container/apiv1/containerpb/cluster_service.pb.go`:

```go
SecurityPostureConfig_MODE_UNSPECIFIED               SecurityPostureConfig_Mode              = 0  // "Default value not specified."
SecurityPostureConfig_DISABLED                       SecurityPostureConfig_Mode              = 1
SecurityPostureConfig_BASIC                          SecurityPostureConfig_Mode              = 2

SecurityPostureConfig_VULNERABILITY_MODE_UNSPECIFIED SecurityPostureConfig_VulnerabilityMode = 0  // "Default value not specified."
SecurityPostureConfig_VULNERABILITY_DISABLED         SecurityPostureConfig_VulnerabilityMode = 1
SecurityPostureConfig_VULNERABILITY_ENTERPRISE       SecurityPostureConfig_VulnerabilityMode = 3
```

So an omitted message yields `mode == "MODE_UNSPECIFIED"` and `vulnerabilityMode == "VULNERABILITY_MODE_UNSPECIFIED"`, and both `!=` comparisons are true. The check reports the security posture dashboard and vulnerability scanning as in place on a cluster where the API said nothing about either.

## The likelier of the two

The per-field case matters more than the whole-message case. A cluster can carry `securityPostureConfig: {mode: BASIC}` with no `vulnerabilityMode` at all, since the two are independent settings and only the posture dashboard is on by default. `vulnerabilityMode` then reads `VULNERABILITY_MODE_UNSPECIFIED`, and the second line passes on a cluster with vulnerability scanning switched off. That is the exact configuration the check exists to catch.

## Suggested fix

Assert the enabled value rather than the absence of the disabled one:

```mql
gcp.project.gkeService.cluster.securityPostureConfig.mode == "BASIC"
gcp.project.gkeService.cluster.securityPostureConfig.vulnerabilityMode == "VULNERABILITY_ENTERPRISE"
```

`ENTERPRISE` and `VULNERABILITY_BASIC` are both marked deprecated in the proto and are [no longer supported](https://cloud.google.com/kubernetes-engine/docs/deprecations/posture-management-deprecation), so the accepted set is small. Whether `VULNERABILITY_ENTERPRISE` is the right bar for this check, or whether the check should split into a posture-dashboard check and a vulnerability-scanning check, is worth deciding before the rewrite.

## Not confirmed against a live cluster

No GCP project was available. The enum zero values and the nil-safe getter path are read from the pinned sources cited above; the response shape for a cluster that never set `securityPostureConfig` is inferred from them rather than observed. One scan of a cluster with the posture dashboard on and vulnerability scanning off settles it.

Note that the Terraform variants of this check assert against `security_posture_config` blocks and are unaffected.

tas50:
## Summary

`mondoo-gcp-security-gke-security-posture-enabled` asserted "not the disabled value" on both `securityPostureConfig` fields. Each field has three states, not two, and the third one passed. This makes the check assert the enabled values instead, and drops the `vulnerabilityMode` assertion because Google has shut the feature down.

## Why

**The vacuous pass**

The GCP provider builds the sub-resource unconditionally, with nil-safe getters, in `providers/gcp/resources/gke.go`:

```go
secPostureConfig, err := CreateResource(g.MqlRuntime, "gcp.project.gkeService.cluster.securityPostureConfig", map[string]*llx.RawData{
    "id":                llx.StringData(c.Id + "/securityPostureConfig"),
    "mode":              llx.StringData(c.GetSecurityPostureConfig().GetMode().String()),
    "vulnerabilityMode": llx.StringData(c.GetSecurityPostureConfig().GetVulnerabilityMode().String()),
```

`Mode` and `VulnerabilityMode` are both proto3 optional pointers on `SecurityPostureConfig`, so a nil receiver or an unset field yields the enum zero value. From `cloud.google.com/go/container@v1.54.0/apiv1/containerpb/cluster_service.pb.go`:

```go
SecurityPostureConfig_MODE_UNSPECIFIED               SecurityPostureConfig_Mode              = 0 // Default value not specified.
SecurityPostureConfig_DISABLED                       SecurityPostureConfig_Mode              = 1
SecurityPostureConfig_BASIC                          SecurityPostureConfig_Mode              = 2
SecurityPostureConfig_ENTERPRISE                     SecurityPostureConfig_Mode              = 3 // deprecated

SecurityPostureConfig_VULNERABILITY_MODE_UNSPECIFIED SecurityPostureConfig_VulnerabilityMode = 0 // Default value not specified.
SecurityPostureConfig_VULNERABILITY_DISABLED         SecurityPostureConfig_VulnerabilityMode = 1
```

An omitted message therefore reads `MODE_UNSPECIFIED`, both `!=` comparisons were true, and the check reported the security posture dashboard as in place on a cluster where the API said nothing about it.

The Terraform variants had the same shape for a different reason. `mode` and `vulnerability_mode` are both `Optional` in the `security_posture_config` block, so an omitted attribute reads `null`, and `null != "DISABLED"` is true:

```
$ cnquery run -c 'parse.json("t.json").params["b"] != "DISABLED"'   # absent key
[ok] value: _
$ cnquery run -c 'parse.json("t.json").params["b"] == /^(BASIC|ENTERPRISE)$/'
  expected: == /^(BASIC|ENTERPRISE)$/
  actual:   _
```

**Why `vulnerabilityMode` is gone rather than rewritten to a positive assertion**

Google has retired both tiers of GKE workload vulnerability scanning. From [Vulnerability scanning removal from GKE](https://docs.cloud.google.com/kubernetes-engine/docs/deprecations/vulnerability-scanning-gkee):

> **July 31, 2025**: the standard tier of vulnerability scanning is shutdown. Results for these scans no longer display in the Google Cloud console. You no longer see an option to enable or disable vulnerability scanning for GKE in the Google Cloud console.
>
> **June 16, 2026**: Advanced Vulnerability Insights results no longer display in the Google Cloud console.

and

> If the vulnerability scanning tier is deprecated, you can't enable that tier in clusters. [...] Workload vulnerability scanning is disabled in existing clusters that use the feature.

Both shutdown dates have passed. Rewriting the second line to a positive assertion would have failed every cluster in existence with remediation nobody can follow, which is worse than the bug. The Kubernetes security posture tier that the check is named for is unaffected: it stays supported and is [enabled by default on new clusters running GKE 1.27 or later](https://docs.cloud.google.com/kubernetes-engine/docs/concepts/about-security-posture-dashboard).

`ENTERPRISE` is accepted alongside `BASIC` because `gcloud container clusters update --security-posture` still offers `enterprise`, and a cluster on that tier has the posture dashboard on.

## Changes

- All four variants now assert `mode == /^(BASIC|ENTERPRISE)$/`, so `DISABLED` and the unset enum both fail.
- The `vulnerabilityMode` and `vulnerability_mode` assertions are removed from all four variants.
- `desc`, `audit`, and all three remediations describe what the query now asserts, and the desc says plainly that workload vulnerability scanning is not assessed and points at Artifact Analysis and Security Command Center instead. The audit CLI command and the gcloud remediation drop `--workload-vulnerability-scanning`.
- Added the vulnerability scanning removal page to `refs`.
- Fixtures: added `fail/mode-absent` for a block that omits `mode`; moved `fail/vuln-disabled` to `pass/vuln-mode-disabled`, since that configuration now passes deliberately; dropped the deprecated `vulnerability_mode` line from the other two.
- No `version:` line changed.

## Validation

Fixture suite, all five scenarios:

```
$ go test -tags iac_variants ./content/validation/scans \
    -run 'TestTerraformVariants/mondoo-gcp-security/mondoo-gcp-security-gke-security-posture-enabled-terraform-hcl/'
--- PASS: .../fail/absent
--- PASS: .../fail/mode-absent
--- PASS: .../fail/mode-disabled
--- PASS: .../pass/enabled
--- PASS: .../pass/vuln-mode-disabled
ok
```

The new fail fixture passes on `main`'s query, which is the bug:

```
$ # with the HCL mql reverted to main's form
$ go test -tags iac_variants ./content/validation/scans -run '.../fail/mode-absent'
    iac_variants_test.go:452: expected failed but got passed (score=100)
--- FAIL
```

Plan and state variants are not fixture-gated, so I ran them by hand against scratch plan and state JSON, old query against new:

| fixture | old query | new query |
|---|---|---|
| `security_posture_config = [{mode: BASIC, vulnerability_mode: VULNERABILITY_DISABLED}]` | failed | **ok** |
| `security_posture_config = [{vulnerability_mode: VULNERABILITY_BASIC}]`, no mode | **ok** | failed |
| `security_posture_config = [{mode: DISABLED}]` | failed | failed |

Same results for `cnquery run terraform plan` and `cnquery run terraform state`.

Other gates:

- `TestRemediationSatisfiesCheck/^mondoo-gcp-security-gke-security-posture-enabled-terraform-hcl$` PASS
- `TestTerraformVariantCoverage` PASS
- `validate.py gcp`: 802 passed, 0 failed, including both changed gcloud commands
- `typos` clean, YAML parses
- `PROVIDERS_PATH=<released gcp schema> cnspec policy lint content/mondoo-gcp-security.mql.yaml` reports only the two inherited `main` failures, `dnsKeys` and `networkPolicyAddon`, that #3948 fixes. The new query compiles against the released gcp provider on its own.

No live GCP project was available, so the runtime verdict change is argued from the provider source and the pinned containerpb enums rather than observed on a cluster.

Fixes #3922


tas50:
CI: every content gate is green, including all 12 `terraform` fixture shards, `coverage`, the 6 `remediation` shards, all 6 `validate-terraform` shards, and `Spell Check with Typos`.

Four checks are red and all four are inherited from `main`, not caused by this PR, which touches only `content/*.mql.yaml` and `.tf` fixtures:

- `go-mod`, `go-test`, and the `Test Results` aggregation of `go-test`: the `Code Test` workflow is failing on `main` at 28c3a1bc and at the two commits before it. `go-mod` reports `go.mod` tidy drift on `go.mondoo.com/cnspec/v13`, `go.mondoo.com/mql/v13`, `alecthomas/repr` and `onsi/gomega`.
- `lint` (Lint Policies): fails on `main` too. The two compile errors are `bucketGrants` in `mondoo-digitalocean-security-spaces-key-has-scoped-grants-digitalocean` and `networkPolicyAddon` in `mondoo-gcp-security-gke-cluster-network-policy-enabled-gcp`, both unreleased provider fields that #3948 is cleaning up. Nothing in the log mentions `securityPostureConfig`, and the query in this PR compiles on its own against the released gcp schema.

Issue #3904, opened 2026-09-13
mondoo-okta-security: password recovery checks do not score a rule that defers to the Okta account management policy


tas50:
### The checks

`mondoo-okta-security-password-settings-password-email-recovery-api`, `-password-sms-recovery-api`, and `-password-question-recovery-api`, after #3898.

### The gap

#3898 made these read a password policy's rules on Okta Identity Engine, from `actions.selfServicePasswordReset`. The rule-driven half deliberately skips a rule whose `requirement.accessControl` is `AUTH_POLICY`:

```mql
okta.policies.password.all( rules.where( actions['selfServicePasswordReset']['access'] == "ALLOW" && actions['selfServicePasswordReset']['requirement']['primary'] != empty ).all( actions['selfServicePasswordReset']['requirement']['primary']['methods'].contains("email") ) )
```

Such a rule carries no `primary` and no `stepUp` of its own, so the `.where()` drops it and `.all()` or `.none()` then passes vacuously. A password policy whose every self-service reset rule is `AUTH_POLICY` is also excluded from the settings half, because the partition sends any policy with a `requirement` to the rule half. Such a policy therefore passes all three checks without any recovery method being examined.

### Why it matters

The Okta OpenAPI document describes `accessControl` this way:

> Determines which authentication requirements a user needs to perform self-service operations. `AUTH_POLICY` defers conditions and authentication requirements to the Okta account management policy. `LEGACY` refers to the requirements described by this rule.

`AUTH_POLICY` is not an exotic configuration. It is what an org gets when it moves self-service recovery onto the Okta account management policy, which is the direction Okta has been pushing Identity Engine orgs. For those orgs the three checks are silent rather than wrong, but silent is still a clean pass with nothing verified.

### Evidence

- `SsprRequirement` schema, `okta-sdk-golang/v6@v6.1.7/okta/api/openapi.yaml:78701-78717`, for the `accessControl` enum and its description.
- `SsprPrimaryRequirement` and `SsprStepUpRequirement`, same file at `:78673` and `:78719`, for the `methods` enums the checks read.
- Verified by running the post-#3898 query against a policy fixture whose single rule is `{"access": "ALLOW", "requirement": {"accessControl": "AUTH_POLICY"}}` and whose settings say `okta_email: INACTIVE`, `okta_sms: ACTIVE`, `recovery_question: ACTIVE`. All three checks pass.

### Not confirmed

[...]

PR #3959, opened 2026-09-13
📝 okta: document the password policies the recovery checks leave unscored

tas50:
## Summary

Docs-only. The three password recovery checks silently skip a whole class of password policy on Okta Identity Engine, and their docs did not say so. This makes the gap explicit and points the reader at the policy that actually governs those orgs. Closing the gap in the query needs an mql provider addition, so this PR does not change any `mql:`.

## Why

`mondoo-okta-security-password-settings-password-email-recovery-api`, `-password-sms-recovery-api`, and `-password-question-recovery-api` partition password policies in two lines:

```mql
okta.policies.password.where( rules.none( actions['selfServicePasswordReset']['requirement'] != empty ) ).all( settings['recovery']['factors']['okta_email']['status'] == "ACTIVE" )
okta.policies.password.all( rules.where( actions['selfServicePasswordReset']['access'] == "ALLOW" && actions['selfServicePasswordReset']['requirement']['primary'] != empty ).all( actions['selfServicePasswordReset']['requirement']['primary']['methods'].contains("email") ) )
```

A rule whose `requirement.accessControl` is `AUTH_POLICY` defers its recovery conditions to the Okta account management policy, so it carries no `primary` and no `stepUp`:

> Determines which authentication requirements a user needs to perform self-service operations. `AUTH_POLICY` defers conditions and authentication requirements to the Okta account management policy. `LEGACY` refers to the requirements described by this rule.

`SsprRequirement`, `okta-sdk-golang/v6@v6.1.7/okta/api/openapi.yaml:78700-78717`.

That rule fails the `.where()` predicate, so the rule half scores an empty list. The rule still carries a `requirement`, so `rules.none( requirement != empty )` is false and the policy is excluded from the settings half as well. Both halves pass with nothing examined.

Verified on the literal rule shape:

```
$ mql run -c 'p = parse.json("rule.json").params
p["rules"].where( _["actions"]["selfServicePasswordReset"]["access"] == "ALLOW" && _["actions"]["selfServicePasswordReset"]["requirement"]["primary"] != empty ).length
p["rules"].where( ... ).all( ... contains("email") )
p["rules"].none( _["actions"]["selfServicePasswordReset"]["requirement"] != empty )'

  [failed] [].none()          # settings half excludes the policy
  [failed] p.rules.where.length
    actual:   0               # rule half scores nothing
  [ok] value: true            # .all() passes vacuously
```

with `rule.json` holding one rule of `{"access": "ALLOW", "requirement": {"accessControl": "AUTH_POLICY"}}`.

The desc added in #3898 says the deferring rule is not scored. It does not say the policy also falls out of the settings half, so a reader can reasonably think the settings half still covers it. It does not.

**Why the query is not changed here.** Scoring the deferred rule means reading the Okta account management policy, and that policy cannot be picked out of `okta.policies.accessPolicy` from content. Okta identifies it by `_embedded.resourceType` being `END_USER_ACCOUNT_MANAGEMENT` (`AccessPolicy_allOf__embedded`, same OpenAPI file at `:85778-85790`), and the provider drops `_embedded`: `PolicyWrapper` in `providers/okta/resources/sdk/policies.go:20-32` decodes only id, name, description, priority, status, system, type, created, lastUpdated, conditions, and settings, and `newMqlOktaPolicy` in `providers/okta/resources/policies.go` builds `okta.policy` from exactly those. The policy's display name is the only remaining discriminator, and Okta documents the resource type, not the name, as the identifier. Details and the second, semantic half of the problem are on #3904.

## Changes

For each of the three parent checks, `-password-email-recovery`, `-password-sms-recovery`, and `-password-question-recovery`:

- `desc`: add that a password policy whose self-service reset rules all defer is read from neither its rules nor its settings, so the check leaves it unscored.
- `audit`: add a paragraph naming `requirement.accessControl` `AUTH_POLICY` as the cause and directing the operator to **Okta account management** under **Security > Authentication Policies**, whose rules matching the `recover` operation decide which authenticators the reset accepts.
- `refs`: add the Okta guide for that policy.

No `mql:`, no `filters:`, no `impact:`, no tags, and no `version:` changed.

## Validation

- `cnspec policy lint content/mondoo-okta-security.mql.yaml` → `valid policy bundle(s)`.
- `python3 -c "import yaml; yaml.safe_load(...)"` → parses.
- `typos content/mondoo-okta-security.mql.yaml` → clean.
- Admin Console path confirmed against the Okta help topic for the Okta account management policy: "In the Admin Console, go to Security > Authentication Policies and select Okta account management."
- Ref title checked against the page it points at: `Configure an Okta account management policy | Okta Developer`.
- No remediation code or CLI command changed, and there is no okta validator under `content/validation/remediation/commands/`.

Refs #3904


tas50:
CI: every content job passes (terraform, cloudformation, bicep, remediation, coverage, kubernetes, dockerfile, and all the `validate-*` jobs). Four checks are red and all four are inherited from `main` at `28c3a1bcf`, which this branch sits on:

- `go-mod`, `go-test`, and `Test Results` fail identically on `28c3a1bcf` itself. `go-test` is the same signature in both: `FAIL go.mondoo.com/cnspec/apps/cnspec/cmd [setup failed]`, `DONE 1377 tests, 1 skipped, 1 failure, 1 error`.
- `lint` reports exactly two SARIF errors, neither in this file: `cannot find field or resource 'bucketGrants' in block for type 'digitalocean.spacesKey'` and `cannot find field 'networkPolicyAddon' in gcp.project.gkeService.cluster.addonsConfig`. #3948 is clearing the second.

`cnspec policy lint content/mondoo-okta-security.mql.yaml` is clean locally.

PR #3962, opened 2026-09-13
📝 docs: narrow the resource-name collision trap to what still reproduces

### LINKED ISSUE #3901: 📝 docs: the resource-name collision trap in CLAUDE.md no longer reproduces for private sub-resources
Both `CLAUDE.md` and `content/CLAUDE.md` teach the "a dotted path that is also a resource name is not a field read" trap with four worked examples. All four stopped reproducing when mondoohq/mql#10349 landed on 2026-08-24, and the fix is in the mql revision cnspec pins today (`go.mondoo.com/mql v0.0.0-20260912222442-e936bb778bca`).

## The examples that no longer reproduce

`content/CLAUDE.md`, section "A dotted path that is also a resource name is not a field read", and the repo-root `CLAUDE.md`, section "Two real bugs that are easy to miss", both name:

- `azure.subscription.aksService.cluster.autoUpgradeProfile.upgradeChannel`
- `gcp.project.gkeService.cluster.networkPolicy.*`
- `aws.emr.cluster.encryptionConfiguration.*`
- `cloudflare.zone.settings.*`

mql#10349 added `compiler.prefersFieldOverResource` in `mqlc/mqlc.go`, which breaks the greedy dotted-path walk when the longer resource is `private`, the owner declares a field of that name, and the field is not an implicit resource and has exactly that resource's type. Every one of the four sub-resources above is private:

```
azure.subscription.aksService.cluster.autoUpgradeProfile   private=True
gcp.project.gkeService.cluster.networkPolicy               private=True
aws.emr.cluster.encryptionConfiguration                    private=True
cloudflare.zone.settings                                   private=True
```

Compiling all four with cnspec's pinned `mqlc` against the installed provider schemas produces a field read bound to the owner, not a bare sub-resource. A bare sub-resource would appear as its own resource-creation chunk carrying the full dotted name. None of the four does:

```
azure.subscription.aksService.cluster.autoUpgradeProfile.upgradeChannel != "none"
  chunk 1 id="azure.subscription.aksService.cluster"   (resource creation)
  chunk 2 id="autoUpgradeProfile"       binding=chunk1  type=azure.subscription.aksService.cluster.autoUpgradeProfile
  chunk 3 id="upgradeChannel"           binding=chunk2  type=string
  chunk 4 id="!="                       binding=chunk3

gcp.project.gkeService.cluster.networkPolicy.enabled == true
  chunk 1 id="gcp.project.gkeService.cluster"          (resource creation)
  chunk 2 id="networkPolicy"            binding=chunk1  type=gcp.project.gkeService.cluster.networkPolicy
  chunk 3 id="enabled"                  binding=chunk2  type=bool
  chunk 4 id="=="                       binding=chunk3

aws.emr.cluster.encryptionConfiguration.atRestEnabled == true
  chunk 1 id="aws.emr.cluster"                         (resource creation)
  chunk 2 id="encryptionConfiguration" binding=chunk1  type=aws.emr.cluster.encryptionConfiguration
  chunk 3 id="atRestEnabled"           binding=chunk2  type=bool
  chunk 4 id="=="                      binding=chunk3

cloudflare.zone.settings.alwaysUseHttps == true
  chunk 1 id="cloudflare.zone"                         (resource creation)
  chunk 2 id="settings"                binding=chunk1  type=cloudflare.zone.settings
  chunk 3 id="alwaysUseHttps"          binding=chunk2  type=string
  chunk 4 id="=="                      binding=chunk3
```

## Why this matters

The doc's job is to stop a reviewer filing a false positive. Right now it does the opposite: #3450 was filed against four GKE checks purely on the strength of these examples plus a schema read showing the paths are resource names, and none of the four checks is actually broken. Being a resource name is no longer sufficient to win the path walk, so the static schema read the doc implies is enough no longer decides anything.

## The trap is not gone, only narrowed

`prefersFieldOverResource` returns false, and the greedy walk still wins, when:

- the longer resource is **not** private, or
- the owner does not declare a field of that name, or
- the field is an implicit resource (the singular accessor `lr` generates for every `x.y` resource), or
- the field's type is not exactly that resource

So the section should stay, with the precondition stated and the confirmation procedure changed. Compiling the query and inspecting the chunk bindings answers it directly, and it does not need a live cloud account the way the current `provider returned no data and no error for a field ... id=` procedure does.

## Suggested change

1. State the precondition: the collision bites only where the longer resource is not private, or the owner carries no matching non-implicit field of that type.
2. Replace the four worked examples with ones that still reproduce under the current compiler, or mark them as historical and say what fixed them.
3. Update the confirmation procedure to "compile the query and check whether the dotted name appears as its own resource-creation chunk", which is cheap and needs no credentials.
4. Keep both copies in sync: `content/CLAUDE.md` carries the long form, the repo-root `CLAUDE.md` the short one.

tas50:
## Summary

Rewrites the "a dotted path that is also a resource name is not a field read" guidance in `content/AGENTS.md` and `AGENTS.md`. The trap is real but narrower than the docs claim: all four worked examples they carried stopped reproducing when [mondoohq/mql#10349](https://github.com/mondoohq/mql/pull/10349) landed, and the section now states the precondition, lists examples that still reproduce, and swaps the confirmation procedure for two credential-free schema reads.

## Why

mql#10349 (2026-08-24) added `compiler.prefersFieldOverResource` in `mqlc/mqlc.go`. It breaks the greedy dotted-path walk when the longer resource is `private`, the owner declares a field of that name, that field is not an implicit resource, and its type is exactly that resource. The fix is in the mql revision `main` pins (`go.mondoo.com/mql v0.0.0-20260913125602-59b558e6f8ad`); confirmed present in the module cache.

Compiling each of the four documented examples with that pinned `mqlc` against the **released** provider schemas gives a field read bound to the owner, not a bare sub-resource:

```
azure.subscription.aksService.cluster.autoUpgradeProfile.upgradeChannel != "none"
  chunk 1 id="azure.subscription.aksService.cluster"           (resource creation)
  chunk 2 id="autoUpgradeProfile"  binding=1  type=azure.subscription.aksService.cluster.autoUpgradeProfile
  chunk 3 id="upgradeChannel"      binding=2  type=string

gcp.project.gkeService.cluster.networkPolicy.enabled == true
  chunk 1 id="gcp.project.gkeService.cluster"                  (resource creation)
  chunk 2 id="networkPolicy"       binding=1
  chunk 3 id="enabled"             binding=2  type=bool

aws.emr.cluster.encryptionConfiguration.atRestEnabled == true
  chunk 1 id="aws.emr.cluster"                                 (resource creation)
  chunk 2 id="encryptionConfiguration"  binding=1
  chunk 3 id="atRestEnabled"            binding=2  type=bool

cloudflare.zone.settings.alwaysUseHttps == true
  chunk 1 id="cloudflare.zone"                                 (resource creation)
  chunk 2 id="settings"            binding=1
  chunk 3 id="alwaysUseHttps"      binding=2  type=string
```

Left as written, the section kept producing the false positive it exists to prevent. #3450 was filed against four GKE checks on the strength of these examples plus a schema read showing the paths are resource names, and none of the four checks was broken.

The trap is not gone, so the section stays. Scanning every released provider schema and compiling one field read per candidate path turns up around a dozen paths where the greedy walk still wins, all of them the same shape: the owner declares a field of that name with a **different** type, so the redirect is declined. Examples, all verified by compiling:

| Path | Owner's field | Result |
|---|---|---|
| `gitlab.user.email` | `string` | `chunk 1 id="gitlab.user.email"` |
| `azure.subscription.sqlService.server.failoverGroup.readWriteEndpoint` | `dict`, deprecated for `readWriteListener` | `chunk 1 id="azure.subscription.sqlService.server.failoverGroup.readWriteEndpoint"` |
| `ms365.exchangeonline.mailbox` | `[]dict` | `chunk 1 id="ms365.exchangeonline.mailbox"` |
| `arista.eos.portSecurity` | `[]arista.eos.portSecurity` | `failed to compile $whereNot: resource 'arista.eos.portSecurity' is not a list type` |

Two further corrections the old text needed:

- `cnspec providers resources <provider> --json` lists only **public** resources, so the private sub-resources at the centre of this trap never appear in it. Naming the resource explicitly does show it, along with `"private": true`.
- mql#10349 shipped on the v14 line. It is not in mql v13.38.1, which is what a released `cnspec` 13.x binary is built against, so a local `cnquery run <provider> --ast` can disagree with what CI compiles. Verified: the same azure query gives the field read under the pinned compiler and the bare sub-resource under the installed `cnspec 13.38.1`.

## Changes

- `content/AGENTS.md`: section retitled "is not always a field read". States the four preconditions, gives the two-command schema check, marks the old four examples as compiling correctly today with an explicit "don't file against them", adds a table of paths that still collide, and replaces the fix example with one that binds a block to the owner and reads the typed field name.
- `AGENTS.md`: the short form is rewritten to match, and a row is added to "Don't flag these, they're correct MQL" so a reviewer who never opens `content/AGENTS.md` does not refile #3450.

No policy content or `version:` line is touched.

## Validation

- Compiled every example and counter-example with the `mqlc` from the pinned mql revision against the released provider schemas, and inspected the chunk bindings.
- Scanned all released provider schemas for dotted resources whose prefix is also a resource, compiling one field read each, to separate paths that still collide from those that no longer do.
- Verified `readWriteListener` exists on `azure.subscription.sqlService.server.failoverGroup` in the released azure schema, and that `azure.subscription.sqlService.servers.all(failoverGroups.all(readWriteListener.failoverPolicy == "Automatic"))` compiles with `servers` as a field read.
- `typos AGENTS.md content/AGENTS.md` clean.
- Docs only, so no policy lint or fixture suite applies.

Fixes #3901


tas50:
CI status on this branch, for the record. Every content gate is green: all 12 terraform shards, all 4 cloudformation shards, all 3 bicep shards, all 6 remediation shards, dockerfile, and kubernetes. The kubernetes shard failed once on the provider RPC flake (`rpc error: code = Unavailable desc = error reading from server: EOF` inside `runBundleReports`) and passed on rerun.

The four red checks are all inherited and none of them can be caused by a diff of two Markdown files:

- `lint` reports exactly two errors, `networkPolicyAddon` in `mondoo-gcp-security` and `bucketGrants` in `mondoo-digitalocean-security`. Both reproduce locally against the released provider schemas on the base commit, and #3948 is cleaning them up.
- `go-mod` fails on a `go.mod` tidy diff (`github.com/onsi/gomega` indirect), `go-test` on `apps/cnspec/cmd [setup failed]`, and `Test Results` just aggregates `go-test`. The Code Test workflow is failing on `main` at 28c3a1bcf, this branchs base.

PR #3971, opened 2026-09-13
🐛 validation: pair IaC variants with their parent by declaration

### LINKED ISSUE #3874: mondoo-vmware-vsphere-esxi: vlan-4095 port-group check: closed-loop remediation test skips it because the parent UID ends in a hyphen
### The query

`TestRemediationSatisfiesCheck` in `content/validation/scans/remediation_closes_check_test.go` finds a variant's parent by trimming the IaC suffix from the variant UID:

```go
parent := byUid[strings.TrimSuffix(uid, suffix)]
snippet, ok := snippetFor(parent, suffix)
if !ok {
    continue
}
```

### Why it fails

In `content/mondoo-vmware-vsphere-esxi.mql.yaml` the parent UID ends in a trailing hyphen, a truncation artifact:

| Parent | Variant |
|---|---|
| `mondoo-vmware-vsphere-esxi-security-baseline-ensure-port-groups-are-not-configured-to-vlan-4095-and-0-except-for-virtual-` | `mondoo-vmware-vsphere-esxi-security-baseline-ensure-port-groups-are-not-configured-to-vlan-4095-and-0-except-for-virtual-terraform-hcl` |

Trimming `-terraform-hcl` from the variant leaves `...except-for-virtual`, without the trailing hyphen, so no parent matches. The parent does document an `id: terraform` remediation, but the snippet is never scanned against the check and the suite stays green.

### Evidence

```
go test -tags iac_variants ./content/validation/scans -run 'TestRemediationSatisfiesCheck/^mondoo-vmware-vsphere-esxi-security-baseline-ensure-port-groups' -v
```

runs only `...ensure-port-groups-are-not-configured-to-the-value-of-the-native-vlan-terraform-hcl`. The `vlan-4095` variant does not appear.

A scan of every `content/*.mql.yaml` for variants whose UID is not `<parent uid>-<suffix>` found this parent plus the five Okta password-settings checks fixed for #3856. No other policy has the mismatch.

### Suggested fix

Resolve each variant's parent from the parents' `variants:` lists rather than from the UID string, or fail the test when an IaC variant's trimmed UID resolves to no query. Either change keeps a future naming mismatch from silently dropping a check out of the closed-loop suite.

tas50:
## Summary

`TestRemediationSatisfiesCheck` paired an IaC variant with its parent check by trimming the IaC suffix off the variant uid. This reads the pairing out of the parents' `variants:` lists instead, and fails on an IaC variant that no check declares, so the ESXi VLAN 4095 port group check rejoins the closed-loop suite and a future naming mismatch is reported rather than skipped.

## Why

The parent uid of the ESXi port group check ends in a stray hyphen, a truncation artifact:

| Parent | Variant |
|---|---|
| `mondoo-vmware-vsphere-esxi-security-baseline-ensure-port-groups-are-not-configured-to-vlan-4095-and-0-except-for-virtual-` | `...-except-for-virtual-terraform-hcl` |

Trimming `-terraform-hcl` off the variant yields `...-except-for-virtual`, without the trailing hyphen, so `byUid` found no parent, `snippetFor(nil, ...)` returned false, and the loop moved on. The parent documents an `id: terraform` remediation that was never scanned against the check.

On `origin/main`:

```
$ go test -tags iac_variants ./content/validation/scans \
    -run 'TestRemediationSatisfiesCheck/^mondoo-vmware-vsphere-esxi-security-baseline-ensure-port-groups' -v
=== RUN   TestRemediationSatisfiesCheck/...-to-the-value-of-the-native-vlan-terraform-hcl
--- PASS: TestRemediationSatisfiesCheck (0.96s)
ok
```

Only the native VLAN variant runs. The VLAN 4095 variant is absent.

A scan of every registered policy in `tfVariantPolicies` for variants whose uid is not `<parent uid><suffix>` finds this one check and nothing else, and finds no IaC variant that no parent declares. Renaming the parent would fix the trim, but that uid is cited by name in the compliance framework maps for 13 frameworks, so a rename would unmap the check from all of them. The pairing is what should be fixed, and the declaration in `variants:` is where it already lives.

## Changes

- `content/validation/scans/remediation_closes_check_test.go`: add `variantParents`, which maps a variant uid to the check that lists it under `variants:`, and use it in `TestRemediationSatisfiesCheck` in place of `byUid[strings.TrimSuffix(uid, suffix)]`.
- An IaC variant that no check declares is now reported with `t.Errorf` instead of falling through the `snippetFor` miss, so the next mismatch is loud.
- The shard filter moved above the pairing so each orphan is reported by exactly one shard.

No content changes, and no policy version touched.

## Validation

With the fix, the variant runs and passes:

```
--- PASS: TestRemediationSatisfiesCheck/...-vlan-4095-and-0-except-for-virtual-terraform-hcl (0.37s)
--- PASS: TestRemediationSatisfiesCheck/...-to-the-value-of-the-native-vlan-terraform-hcl (0.38s)
```

Mutation test, to prove the new subtest asserts rather than merely running: changing the parent's Terraform remediation from `vlan_id = 100` to `vlan_id = 4095` in the working tree fails it with `the remediation for this check does not satisfy it`. Reverted.

Full suite across all 26 registered policies:

```
$ go test -tags iac_variants -timeout 45m ./content/validation/scans -run 'TestRemediationSatisfiesCheck'
ok  	go.mondoo.com/cnspec/content/validation/scans	714.112s
```

No orphan-variant errors, so the new assertion holds across the corpus today.

Fixes #3874


tas50:
`go-test` and `go-mod` are red here for a reason that predates this branch, filed as #3975.

`apps/cnspec/cmd/serve_test.go` imports `go.mondoo.com/cnspec/v13/apps/cnspec/cmd/config` and two `go.mondoo.com/mql/v13/...` packages, and 1d8ce4225 (#3920) removed those `v13` requirements from `go.mod`. `go-test` reports it as `apps/cnspec/cmd [setup failed]`, and `go-mod` fails because `go mod tidy` wants the requirements back.

Reproduced with this branch's only change reverted to `origin/main`:

```
$ go vet ./apps/cnspec/cmd/
apps/cnspec/cmd/serve_test.go:14:2: no required module provides package go.mondoo.com/cnspec/v13/apps/cnspec/cmd/config
```

and `go mod tidy` then produces byte-for-byte the diff CI prints (`index ece84066..1d59e00e`). This PR touches one file, `content/validation/scans/remediation_closes_check_test.go`, behind the `iac_variants` build tag, which neither job compiles.

The suite this PR changes passes in full locally:

```
$ go test -tags iac_variants -timeout 45m ./content/validation/scans -run 'TestRemediationSatisfiesCheck'
ok  	go.mondoo.com/cnspec/content/validation/scans	714.112s
```

Issue #4018, opened 2026-09-18
IaC variant fixtures: terraform-plan and terraform-state scenarios cannot be loaded


4ndr3-mondoo:
**Describe the bug**

`tfAssetForVariant` in `content/validation/scans/iac_variants_test.go` recognises `-terraform-plan` and `-terraform-state` uid suffixes and sets the connection type accordingly, but passes the scenario directory through unchanged as `path`:

```go
case strings.HasSuffix(uid, "-terraform-plan"):
    connType = "terraform-plan"
case strings.HasSuffix(uid, "-terraform-state"):
    connType = "terraform-state"
case strings.HasSuffix(uid, "-cloudformation"):
    connType = "cloudformation"
    path = cfnTemplatePath(path)
case strings.HasSuffix(uid, "-bicep"):
    connType = "bicep"
    path = bicepTemplatePath(path)
```

CloudFormation and Bicep each resolve the directory down to the template file, because those connections take a file path. The Terraform plan and state connections take a file path too: `NewPlanConnection` calls `os.ReadFile(cc.Options["path"])` (`mql/providers/terraform/connection/tfplan.go`). Handed a directory it fails, so a plan or state fixture placed under `fixtures/iac-variants/` cannot run. There are currently none, which is consistent with that.

**Expected behavior**

A scenario directory holding a single `plan.json` (or state JSON) is resolved to that file, the same way `cfnTemplatePath` and `bicepTemplatePath` do for their formats.

**Suggested fix**

A resolver for both, next to the two that already exist. Written out here rather than opened as a PR, so take it as a starting point rather than something to merge as is:

```go
case strings.HasSuffix(uid, "-terraform-plan"):
    connType = "terraform-plan"
    path = tfDocumentPath(path)
case strings.HasSuffix(uid, "-terraform-state"):
    connType = "terraform-state"
    path = tfDocumentPath(path)
```

```go
// tfDocumentPath returns the single JSON document inside a scenario dir. The
// terraform plan and state connections take a file PATH, not a dir:
// NewPlanConnection reads it with os.ReadFile.
func tfDocumentPath(dir string) string {
	entries, err := os.ReadDir(dir)
	if err != nil {
		return dir
	}
	for _, e := range entries {
		if e.IsDir() {
			continue
		}
		if strings.HasSuffix(e.Name(), ".json") {
			return filepath.Join(dir, e.Name())
		}
	}
	return dir
}
```

[...]

PR #4019, opened 2026-09-18
🐛 aws: fix IAM one-active-access-key check on Terraform plan and state

4ndr3-mondoo:
`mondoo-aws-security-iam-users-only-one-access-key` has four variants, and the two Terraform ones do not test what the check's title says.

The `aws` and `terraform-hcl` variants count active access keys, which is the control the title describes. The `terraform-plan` and `terraform-state` variants instead look for a preventive IAM policy that denies `iam:CreateAccessKey` above a key count, and they filter on `aws_iam_policy` rather than on `aws_iam_access_key`:

```yaml
- uid: mondoo-aws-security-iam-users-only-one-access-key-terraform-plan
  filters: asset.platform == "terraform-plan" && terraform.plan.resourceChanges.contains(type == "aws_iam_policy")
  mql: |
    terraform.plan.resourceChanges.where(type == "aws_iam_policy").any(
      change.after.policy.contains('iam:CreateAccessKey') &&
      change.after.policy.contains('Deny') &&
      change.after.policy.contains('NumericGreaterThan') &&
      change.after.policy.contains('iam:AccessKeysCount')
    )
```

Both the filter and the query point at the wrong resource type, so the check runs on a set of assets that is nearly disjoint from the set it is meant to cover.

Any plan containing an ordinary `aws_iam_policy` fails the check, because ordinary policies do not carry that guardrail. A Lambda function with its execution role and a CloudWatch Logs policy is enough, and that is one of the most common shapes a Terraform plan takes. The report then states that a user holds more than one active access key, about a plan that manages no IAM users at all.

Meanwhile a plan that really does create two active access keys for the same user passes, as long as it carries no `aws_iam_policy`, because the filter excludes it from the check entirely.

### Fix

Port the `terraform-hcl` variant to the plan and state shapes. Filter and query both key off `aws_iam_access_key`:

```yaml
- uid: mondoo-aws-security-iam-users-only-one-access-key-terraform-plan
  filters: asset.platform == "terraform-plan" && terraform.plan.resourceChanges.contains(type == "aws_iam_access_key")
  mql: |
    terraform.plan.resourceChanges.where(type == "aws_iam_access_key").where(change.after["status"] != "Inactive").map(change.after["user"]).join("|").split("|").duplicates == empty

- uid: mondoo-aws-security-iam-users-only-one-access-key-terraform-state
  filters: asset.platform == "terraform-state" && terraform.state.resources.contains(type == "aws_iam_access_key")
  mql: |
    terraform.state.resources.where(type == "aws_iam_access_key").where(values["status"] != "Inactive").map(values["user"]).join("|").split("|").duplicates == empty
```

The `join("|").split("|")` detour is not cosmetic. Values read out of `change.after` carry no element type, so `.duplicates` refuses the mapped list outright with `cannot extract duplicates from array, must be a basic type`. The string round-trip gives it the basic type it needs. The `terraform-hcl` variant already carries the same detour for the same reason, and its comment in the bundle explains it.

### Verification

Run against the full `mondoo-aws-security.mql.yaml`, not an isolated bundle:

| Plan contains | Before | After |
|---|---|---|
| two `aws_iam_access_key` for one user, no `status` attribute (Terraform's default is Active, so this is how most real plans look) | check does not run | fails, correct |
| two `aws_iam_access_key` for one user, both `status: Active` | check does not run | fails, correct |
| one key each for two users, plus a third key that is `Inactive` | check does not run | passes, correct |
| `aws_iam_policy` for a Lambda, no access keys | fails, wrong | check does not run, correct |

The `terraform-state` variant compiles and mirrors the plan variant line for line, but has no fixture yet because the harness cannot load state fixtures either (see #4018). Treat it as reviewed rather than verified.

### Fixtures in this PR

`content/validation/scans/fixtures/iac-variants/mondoo-aws-security/mondoo-aws-security-iam-users-only-one-access-key-terraform-plan/` with `pass/one-key-per-user/plan.json` and `fail/two-active-keys-one-user/plan.json`. The fail case uses no `status` attribute on purpose.

These are the first terraform-plan fixtures in the repo. They cannot run until `tfAssetForVariant` resolves a file path for plan scenarios. Related to #4018. They are included anyway so that whoever lands the harness change has something to run against immediately.


4ndr3-mondoo:
Thanks, both points checked.

**Pipe delimiter:** AWS restricts IAM user names to `[A-Za-z0-9+=,.@_-]`, so `|` cannot appear in a valid `aws_iam_access_key.user`. The delimiter matches the `terraform-hcl` variant on purpose so the four variants read the same; changing it here alone would make them diverge for a case AWS rules out.

**Missing `status`:** verified, and it is the first row of the verification table above. The fail fixture in this PR (`fail/two-active-keys-one-user/plan.json`) deliberately has no `status` attribute on either key, and the check fails it as expected against the full policy. MQL evaluates `null != "Inactive"` as true inside `.where()`, which is also what the HCL variant relies on.

arlimus:
Added the loaders that close #4018 
Closes #4018

